PR #2434, opened 2015-06-28
trying to merge firemodels into rmcdermo

rmcdermo:
See if this works.


rmcdermo:
Nope.  This is not the direction I wanted to merge.

PR #2473, opened 2015-07-08
FDS Source: make EXTINCT(1:N_REACTIONS) to handle extinction with multiple reactions

rmcdermo:
Glenn,
Please double check that the makefile changes here are ok.  Thanks.


rmcdermo:
Need to change mpiifort to ifort for mpi windows makefire entries.

PR #2543, opened 2015-07-29
FDS Verification: rework Shunn manufactured solution

rmcdermo:
This is a fairly big commit to fix the convergence study for the Shunn manufactured solution.


rmcdermo:
I will also comment that in this pull request I have added a new solution "Saad" that is under construction.

PR #2559, opened 2015-08-04
Manuals: change SVN to Git and add \gitrevision to FDS docs

rmcdermo:
Glenn,
Please have a look at these changes.
Thx


rmcdermo:
Well, out of habit I merged the request.  Just note that I touched a couple of SMV docs to make the Git string consistent.

PR #2589, opened 2015-08-08
Improve support for xlf compiler on power architecture both on linux and AIX

kiwifb:
Provide generic folders for use to compile the targets `xlf_openmp` and `aix_mpi`. The hard bits was to create the proper preprocessing option. The tricky bits are fully documented in `makefile`.
Made the architecture optimization more generic so it can be applied not only on power7 but power8 or power6.


mcgratta:
We do not have IBM computers or compilers, so I will leave it to you to test these changes.

kiwifb:
Thanks, I am doing some testing this week. That stuff was just the necessary bits to compile (on three different variations of OS/Hardware aix-pwr7, aix-pwr6 and linux-pwr7).

PR #2594, opened 2015-08-10
Variable omp thread number

lu-kas:
Support for variable OMP thread number per MPI process. 

Usage:
use N_THREAD keyword in the MESH namelist. N_THREAD is checked to be consistent within a MPI process.

Note:
The "MPI and OMP" checking routines have been rearranged, as the OMP part may only be executed after the MESH information have been read.


mcgratta:
Hi Lukas

There appears to be a conflict with your pull request. I would like Randy or Glenn to take a look at it because I am just learning git. Or you could withdraw the request, do an update, and resubmit.

lu-kas:
Hey Kevin,

I see, that I have an rather old fork base. I think it will be easier that I move the changes to a recent base.

Best,
Lukas

rmcdermo:
Hi Lukas,
We have a wiki that shows how to track upstream remote repos.  This is how most of us work.  We have our own forked repos for development.  But every morning we fetch firemodels/development and merge any changes that are there.  Then all our pull requests are up-to-date.  (probably you already know how to do all this).
https://github.com/firemodels/fds-smv/wiki/Git-User-Workflow
Best,
Randy

PR #2606, opened 2015-08-11
added capability for individual number of OMP threads per MPI process

lu-kas:
Second try.

Support for variable OMP thread number per MPI process.

Usage:
use N_THREAD keyword in the MESH namelist. N_THREAD is checked to be consistent within a MPI process.

Note:
The "MPI and OMP" checking routines have been rearranged, as the OMP part may only be executed after the MESH information have been read.


mcgratta:
Lukas, thanks. Do we need to add anything to the FDS User's Guide?

lu-kas:
see next pull request

PR #2636, opened 2015-08-14
hidden OpenMP library calls

lu-kas:
hidden OpenMP library calls, in case that FDS is compiled without OpenMP support


mcgratta:
Thanks!

PR #2638, opened 2015-08-14
FDS Source: Fix bug in SWELL_RATIO calculation.

mcgratta:


shostikk:
So SWELL_RATIO was not capturing some cases? Looks OK to me.

mcgratta:
We got an access violation from valgrind. The MAX_DENSITY algorithm allowed access to an uninitialized variable. Should be fixed now.

PR #2639, opened 2015-08-16
FDS Source: Ensure that OpenMP picks up the environment variable OMP_…

mcgratta:
…NUM_THREADS if there is no user specified value.


lu-kas:
this is not needed, but does not hurt

however, strictly speaking you do not ensure with this, that the environment variable is picked up correctly. you just set the value that was just returned by the OMP library get-call.

btw, if i remember correctly, the OMP specifications says that you just set the maximal number of threads, the implementation / runtime may choose to pick a smaller number.

mcgratta:
Without the line that I added, my jobs were getting only 1 thread. Only by setting N_THREADS on the MESH line could I change the number. Our verification case that tests the OpenMP has failed the past few nights. I am hoping that my change fixes it.

PR #2644, opened 2015-08-24
FDS Tech Guide: spell out volumetric source terms and Lagrangian dive…

rmcdermo:
…rgence expression


rmcdermo:
Jason, Kevin,
Have a look at these changes to the guide and make sure you agree.

mcgratta:
What is meant by the summation over p and n in Eq 8.29? I understand the p, but are we summing over time steps, n?

rmcdermo:
I will clarify.  We use sub-time steps in the integration of the heat and
mass transfer equations for the droplets.

PR #2673, opened 2015-09-01
FDS Configuration Management Guide: Scrub references to GoogleCode an…

mcgratta:
…d SVN.


mcgratta:
I got rid of references to GoogleCode in the FDS Config Plan. The description of GitHub and the new procedures is light. Add if you want, but I would avoid being too specific because I don't think that this guide is the appropriate venue for really detailed discussion of version control.

PR #2674, opened 2015-09-01
FDS Users Guide: Remove references to GoogleCode.

mcgratta:


mcgratta:
There was only a few small references to GoogleCode in the User's Guide.

PR #2686, opened 2015-09-03
FDS Validation: Add BRE_Spray results and git.txt files.

mcgratta:


mcgratta:
Glenn -- there are now two "new" validation data sets committed, BRE_Sprays and CAROLFIRE. We'll see how firebot handles them.

PR #2697, opened 2015-09-07
FDS Source: Set various real output variables to be single precision …

mcgratta:
…to avoid underflow problems.


drjfloyd:
I think this will wind up causing us problems. For example, if the DV%X is  single precision, then we will wind up having cases where due to precision the device will not wind up in the intended grid cell as the DV%X will differ from the grid X due to rounding.  Also, this means we will never be able to have the device output actually be the computed value in the code.  I had changed the dump routine for device to EB from FB for that reason as FB was causing problems in trying to compare FDS combustion predictions to cantera.

mcgratta:
OK, revert my changes (I haven't figured out how to do that with git yet). I will just set the exponent in the failed cases to have 3 decimal places. I usually truncate the size of the devc output to reduce the overall size of the repo.

PR #2700, opened 2015-09-07
FDS Source: Change device output back to double precision.

mcgratta:


mcgratta:
I changed the device variable output back to double precision.

PR #2719, opened 2015-09-10
pull merges (don't know why git wants me to do this - since I didn't make any of these changes)

gforney:


rmcdermo:
Glenn,
I think this happens because you work in smv_dev and all merges are themselves commits, which change the git hash.

PR #2722, opened 2015-09-10
smokebot: add file at root level to identify repo

gforney:


gforney:
the file I need is in the bot directory now

PR #2743, opened 2015-09-14
FDS Validation:  Misc results for FDS 6.3.0 release.

mcgratta:


mcgratta:
The validation runs for FDS 6.3.0 are almost done. The remaining sets are

Backward_Facing_Step
FM_FPRF
Purdue_Flame (which I had to restart today)
UL_NFPRF (the big box burning cases)
UMD_Gas_Burner (not started -- are these ready to go? There are no Exp or Output folders)
USN_Hangars

rmcdermo:
I have not been given the all clear to publish the UMD data.  But I think
we are close.  I will send a note to Andre et al. and see if they are OK
with this going into the 6.3.0 manual.

PR #2807, opened 2015-09-22
Update Firebot Figure Script and Update User Guide to add chemistry subiterations

cweinschenk:


rmcdermo:
Craig,
This is better, but you need to correct the Copy_Firebot_Figures.sh script.  Or, if this does not work for you we may need to not keep this under version control.  But I found I needed to make the changes I made in order to make it work with current Firebot.

cweinschenk:
The version as it existed in the repository did not work which is why I made changes.

rmcdermo:
Hmmm... works on my end.  We should get this sorted out.  It is a useful
script.  Are you on burn?  Not sure why it would be different.  Maybe Glenn
has an idea.

On Tue, Sep 22, 2015 at 10:05 AM, Craig Weinschenk <notifications@github.com

cweinschenk:
I am on Blaze. That could be why. I converted the script back and submitted a new pull request with just the user guide changes.

rmcdermo:
Glenn,
Does this behavior make sense to you?

If so, I suggest Craig commits his version as Copy_Firebot_Figures_Blaze.sh
and I'll do likewise for Burn.

On Tue, Sep 22, 2015 at 10:45 AM, Craig Weinschenk <notifications@github.com

gforney:
what doesn't work?

rmcdermo:
See the Copy_Firebot_Figures.sh changes from pull requiest 2807.

https://github.com/firemodels/fds-smv/pull/2807/files#diff-0

On blaze, Craig needs to make these changes.  On burn, the original works
fine for me.  Does it make sense that we might need two scripts?

cweinschenk:
I logged out of blaze and burn and relogged in and now both scripts seems to be functioning now. I guess it was something to do with Smokevis or something

PR #2854, opened 2015-09-28
FDS Validation: Final results for NRCC_Smoke_Tower.

mcgratta:


mcgratta:
I've finished with the NRCC Smoke Tower validation case. I'm OK with a release of 6.3.0.

PR #2962, opened 2015-10-16
tkorhon1: main.f90 common DT for all meshes: EVACUATION update

tkorhon1:
Now main.f90 is working also for evacuation cases after the common
DT for all meshes changes. The verification suite is run with
the Git base version "Git-r1-120-g2e8b2d2". (This source + Timo'2 modifications
needed for evacuation meshes due to a common DT.) The verification suite was
run using the debug option. Everything seems to be fine.

The changes were made to latest Git version "Git-r1-134-ge2451e6" and this
was checked by Timo's own test cases (fire, fire+evac, mpi, OpenMP, etc
checks that results were the same, both binary and text output was checked).


mcgratta:
Thanks, Timo. None of the recent errors that we have been getting from firebot are related to Evac. They are related to how we queue the MPI jobs.

PR #2989, opened 2015-10-23
FDS Source: Modify fire.f90 to cleanup series rxns loops and Q calcs

cweinschenk:


rmcdermo:
Kevin,
Craig is going to have us do the merge after a local firebot run.

rmcdermo:
Went ahead a merged this because this definitely fixes a problem with series reactions: with the loop in COMBUSTION_MODEL we were doing mixing in between series fast reaction steps, and this is not correct.

PR #2992, opened 2015-10-23
FDS Source: Create separate cpu file.

mcgratta:
I ran these changes through firebot.


rmcdermo:
Kevin,
I think we need to make the old .out file format the default and have a special flag on DUMP for those who want to do large calcs or scaling tests.  I just launched a 12 mesh case and let it run for half an hour and then wanted to go see what the velocity errors were at the mesh boundaries and was surprised to get no information in the .out file.  Until we have solved the velocity error problem at interpolated and immersed boundaries we still need to be monitoring this stuff closely AND reporting the errors to users.  This default makes it look like we are hiding the errors.

rmcdermo:
OK, I found it.  You have the default set to 10 meshes.  Guess I am just unlucky.  But IMO this should just be a user input on DUMP.

PR #3045, opened 2015-11-03
FDS Verification: add hrrpuv_reac_soot.fds

rmcdermo:


drjfloyd:
For a test case should we up the soot yield (right now it is ~1.5 %) so we
get more Q in the second reaction ?

rmcdermo:
Sure.  Go ahead and add.

drjfloyd:
Upped the yield to 30 %.  With N_FIXED_CHEMISTRY_SUBSTEPS=1, Q and Q_R1 + Q_R2 are very close. More error without N_FIXED_CHEMISTRY_SUBSTEPS, but not very large.

PR #3053, opened 2015-11-04
FDS Source: IMPORTANT, fix bug in mixing step of combustion model

rmcdermo:


rmcdermo:
This bug will not affect SIMPLE_CHEMISTRY because there we have UNMIXED_FRACTION=1 and a single chemistry substep.  Also, anything that is purely premixed (UMIXED_FRACTION=0) will also not be affected.  This is why this bug was not obvious---most all of our verification cases are one of these two cases.

However, anything that needed multiple substeps with mixing and reaction up to this point were in error.  We should think about a maintenance release soon to get this fix out there.

drjfloyd:
Looks like we got a clean firebot. Do we want to push out the  maintenance?

rmcdermo:
Let's get release notes in shape and see where we are.  I'm working on improving the time integrator to be a linear combination of FE steps.  This should make it possible to get the same results with H_F and HoC.

rmcdermo:
Here is a text file with one line commit messages for all the commits between 6.3.0 and now.  This may help in preparing release notes.  Please scan this file and make sure anything you've added that is important is in the release notes for 6.3.1.
[fds-smv_6.3.1_gitlog.txt](https://github.com/firemodels/fds-smv/files/28305/fds-smv_6.3.1_gitlog.txt)

mcgratta:
I added mine.

rmcdermo:
Here is a version of the file without merge commits... a little cleaner.
[fds-smv_6.3.1_gitlog.txt](https://github.com/firemodels/fds-smv/files/28323/fds-smv_6.3.1_gitlog.txt)

drjfloyd:
Added mine

PR #3070, opened 2015-11-06
FDS Source: Add HVAC_DT to MISC to override use of FDS timestep in HV…

drjfloyd:
…AC solver. Issue 3002


rmcdermo:
Jason,
Sorry to be picky... but all our DTs in the code are DT_something.  Can I ask you to flip your DT?  Thanks!

drjfloyd:
Good point. No problem. I will fix and push up new.

PR #3074, opened 2015-11-06
FDS Source: fix bugs in ODE_Solver for finite-rate chemistry

rmcdermo:


rmcdermo:
Jason,
Checkout latest.  Basic problem was one of mass conservation between mixing substeps because the "total mixed mass" changes with time.  So we have to work in terms of extensive quantities for species mass going into and out of the integrators.  Before we were just passing ZZ.
I also added hooks for RK2 and RK3 integrators just for completeness.  They helped me along the way with debugging.

drjfloyd:
Things are greatly improved.  If I write debug for ZZ and Q_REAC in RK2, I am seeing very small errors in Q_REAC in the first call to the routine.  The relative error in Q1+Q2 to Q for hrrpuv_reac_soot has dropped an order of magnitude. Before your change we were near 15 % in that early transient and now we are near 1 %.

image

PR #3094, opened 2015-11-09
modified near wall gas density

mpachera:
I tried to add a new &devc which as done for the gas temperature evaluate the gas density near the wall


rmcdermo:
mpachera, thanks for the pull request!  You will notice that I had to modify your source very slightly to avoid tabs.  We hope you will continue to help with code improvements.  Please study the [Developer Commit Guidelines](https://github.com/firemodels/fds-smv/wiki/Developer-Commit-Guidelines), as this will make merging easier in the future.  Cheers!

rmcdermo:
We also like to add features like this to the list of output quantities in the FDS user guide:

https://github.com/firemodels/fds-smv/pull/3096

mpachera:
Thanks should I also write a brief description about the use of this quantity or it can be as it is?

rmcdermo:
For now things are fine.  If we get questions about it, we may want to add
more details.  Thanks.

PR #3159, opened 2015-11-24
FDS Validation: Fix bug in plotting routine.

mcgratta:


mcgratta:
Simo -- you left "o" character in the first column of the dataplot input file. I fixed it.

PR #3204, opened 2015-12-09
Fix PRISME Plot Labels

khamburger:
This corrects several validation plot labels that were inadvertently marked as "DOOR" instead of "SOURCE."


mcgratta:
Looks good, thanks.

PR #3227, opened 2015-12-11
FDS Source: Combine D_REACTION and D_LAGRANGIAN and apply M_DOT_PPP i…

mcgratta:
Don't accept this pull request until Randy has looked it over. This code still breaks the propane_flame_deposition case, but all the rest are OK. The reactionrate_series_reaction case just needed to be run for two time steps rather than one. 

Maybe you can create a test branch and merge this with your current changes.


rmcdermo:
I checked this out with my simple test case.  Let's talk about it Monday.  I think it is not quite ready for prime time.  We need to look at real fire cases.  It is likely we would have to retune the mixing time model for flame height.

rmcdermo:
Kevin is still working on enhancements and testing.  Will try an updated version.

PR #3239, opened 2015-12-15
FDS Source: Change handling of mass source terms.

mcgratta:
Do not accept this pull request until it has been tested.


rmcdermo:
Kevin,

This input file seems to behave the same with current base code and your pull request. I think I had some other modifications to my repo when we saw the low temps before.

[tmp_simple.txt](https://github.com/firemodels/fds-smv/files/62955/tmp_simple.txt)

Good news is that the time steps seem to be roughly equivalent for the same output.

Bad news is that just removing the TAU_MF=0 sends both schemes into lows temperatures.  So, I think we have other problems still.  Try adding back the ramp and confirm you see the same.

rmcdermo:
Also, checkout the simple fire in Verification/Flowfields/realizable_mass_fractions.fds.  I get the same min temps with or without the new scheme, about 17 C.

mcgratta:
For realizable_mass_fractions, the new scheme drops to 17 C initially, then returns to 20 C. This happens regardless of the ramp.

PR #3246, opened 2015-12-16
FDS Source: Fix cell-centered particle init for plane at cell boundar…

drjfloyd:
…y Issue 3243.


drjfloyd:
Had an error in placing cell centered particles for plane at the cell edge. I fixed that.

PR #3253, opened 2015-12-17
smokeview: add position menu labels for geometry slice files, add slices to geometry slice test case

gforney:


rmcdermo:
Obi Wan... put down your weapons.

PR #3266, opened 2015-12-21
tkorhon1: FDS Source, evacuation time change in EVAC_MAIN_LOOP.

tkorhon1:
Added line in EVAC_MAIN_LOOP so that at ICYC=1 T_EVAC is a little bit
less than T_FIRE. This way the EVAC_EFF file is written at ICYC=1.
This used to be the case before (when each fire mesh nad differend DT).
This was broken in the process, where same T and DT was made for all
fire meshes. A small EVSS related bug is also fixed (PERS line given fac_v0_hori).
Also dia_mean value is correct for distributions, where it is not input value (not
needed to specify the random distributrion). Dia_mean is used to calculate
the moment of inerria, so this bug did not matter much. And it did not matter
for the pre-defined person classes (dia_mean was set in the program source).

### DIFF
--- a/FDS_Source/evac.f90
+++ b/FDS_Source/evac.f90
@@ -1950,7 +1950,6 @@ CONTAINS
             CASE Default
                CONTINUE
             END SELECT
-            DIA_MEAN = 0.51_EB
          END IF
 
          !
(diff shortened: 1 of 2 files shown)


tkorhon1:
This affects just fire+evacuation and evacuation calculations, because 
evac_main_loop exits right away, when just fire meshes are present.

Now the evacuation EFF file is once again written to the disk correctly.
This was broken, when Kevin unified the DT for fire meshes.

PR #3306, opened 2015-12-30
FDS Source: make REAC D_SOURCE consistent for CONSTANT_SPECIFIC_HEAT_…

rmcdermo:
…RATIO


drjfloyd:
Why did we go back to using a loop which can't be optimized as well as the array notation plus makes multiple costly calls to the GET_SENSIBLE function?

Also in the loop you are calling GET_SENSIBLE_ENTHALPY+Z with ZZ_GET and then multiplying by DZZ.  If you want to have the loop you should be calling GET_SENSIBLE_ENTHALPY_Z.

rmcdermo:
Because I understand the code.  When we get all this nailed down and have solid verification cases, then we can profile and optimize.  Right now, I still don't trust everything.

drjfloyd:
If we are keeping it this way, you should call GET_SENSIBLE_ENTHALPY_Z and not GET_SENSIBLE_ENTHALPY

rmcdermo:
Perfect example of me not understanding the code.  I will fix.

drjfloyd:
GET_SENSIBLE_ENTHALPY_Z returns the enthalpy for a single tracked species

GET_SENSIBLE_ENTHALPY returns the sum of the single tracked species
enthalpies times a vector of mass fractions.  Typically we call this with
ZZ, but if you call it with DZZ then:

DO N=1,N_TRACKED SPECIES
    CALL G_S_E_N(N,T,H_S_N)
    D_SOURCE=D_SOURCE+H_S*DZZ(N)/RCPT/DT
ENDDO

is equivalent to

CALL G_S_E(DZZ,T,H_S_N)
D_SOURCE = D_SOURCE + H_S_N/RCPT/DT

rmcdermo:
I see.  For now, I want to be able to look at this term and quickly compare it to the notes.  Once the tmp issue is sorted out we can revisit.

PR #3355, opened 2016-01-12
FDS Source: Fix a few issues in CHECK_MASS_DENSITY and clarify variables.

mcgratta:


mcgratta:
Note that I changed variable names to reflect when they represent RHO*ZZ rather than ZZ.

PR #3390, opened 2016-01-21
FDS Source: Add write statement to prevent optimization of an initial…

mcgratta:
This is a bit of a hack. In cylindrical coordinates, the calculation of the radiation solid angles is slightly different due to optimization (-O2) of the routine. The angles DLY(1), DLY(11), DLY(21), etc., are very close to zero, but positive with -O1 or less optimization. They are slightly negative with -O2. The calculation is sensitive to positive and negative angles. The verification case plate_view_factor_cyl_100.fds is different by about 10 kW/m2 depending on the optimization level. The write statement added in this commit suppresses the optimization, and the same result is obtained at all levels of optimization. I could not figure out how exactly the optimization changed the calculation, so I'm just going to suppress it for now.

### DIFF
--- a/FDS_Source/radi.f90
+++ b/FDS_Source/radi.f90
@@ -80,6 +80,7 @@ DO I=1,NRT
          DLY(N) =  (-SIN(DPHI0/2.)*(SIN(PHIUP)-SIN(PHILOW))  +COS(DPHI0/2.)*(COS(PHILOW)-COS(PHIUP)))*F_THETA
          DLB(N) =  (-SIN(DPHI0/2.)*(SIN(PHIUP)-SIN(PHILOW))  -COS(DPHI0/2.)*(COS(PHILOW)-COS(PHIUP)))*F_THETA
          DLZ(N)    = 0.5_EB*(PHIUP-PHILOW)   * ((SIN(THETAUP))**2-(SIN(THETALOW))**2)
+         IF (N==1000000) WRITE(LU_ERR,'(A)') 'This line should never get executed. It is here only to prevent optimization.'
       ELSEIF (TWO_D) THEN
          DLX(N) = (SIN(PHIUP)-SIN(PHILOW))*F_THETA
          DLY(N) = 0._EB


drjfloyd:
Looking at the change I notice that section of code did not explicitly type 2. as 2._EB.  If that is fixed, does O1 and O2 match?

mcgratta:
I did all sorts of things to make -O2 and -O1 match. I pulled the IF (CYLINDRICAL) construct to the outside and created 3 separate loops, one for each geometry. DId not work. I added EB to the REAL statements, 0.5_EB instead of /2., I created new variables like `COSPU=COS(PHIUP)`. Nothing changed things.  The angle values are between 1e-17 and 1e-19. I changed

```
IF (DLY(N)<0._EB) ...
```

to 

```
IF (DLY(N)<-EPSILON_EB) ...
```

which helped, but still I had two different answers between -O1 and -O2. I suspect there are more places where I would have to fix the if statements, but I could not see anything obvious.

shostikk:
Hard to undestand this is.  But suppressing the optimization in the initialization routine does not make much harm.

shostikk:
1. Commented out the write statement.
2. Then compiled on Windows (using Intel(R) Visual Fortran Version 14.0.3.202 Build 20140422).
3. Run plate_view_factor_cyl_100.fds with intel_win_64 and intel_win_64_db 
4. Got identical results (74.0438)

gforney:
Kevin , were you doing this on blaze?

mcgratta:
I did this on burn, but I assume that blaze would exhibit the same problem. At some point, the Intel Fortran compiler changed behavior, or introduced a bug, with the -O2 option. At first, I thought the problem was that the IF statements are inside the DO loops. I found a problem like this last year, reported it to Intel, and they acknowledged that it was probably an optimization problem (and to just add the dummy write statement). I don't like adding dummy statements, so I rearranged the loops. But with the current problem, this did not work. 

Simo, can you confirm the DLY(1), DLY(11), etc, are supposed to be zero in this case. I assume that values of 1e-19 are below precision when adding and multiplying these various trig functions. Maybe what we should do is set values to zero that are below EPSILON_EB.

shostikk:
Yes, I can confirm that in CYLINDRICAL calculation,
IF (J==1) DLY(N) = 0 
IF (J==NRP(I)) DLB(N) = 0 

This was actually written in the original article of Murty and Marthur (1998), but I have never coded it explicitly.

mcgratta:
Should we add a few IF statements? Or modify the STEP logic when we compute IL for CYLINDRICAL coords. I think that the calculation of IL for CYLINDRICAL is done with unique code -- maybe we should add safeguards to it to avoid having the sign of JSTEP change due to numerical round off.

PR #3392, opened 2016-01-23
add a git hash to fds2ascii, misc repo cleanup

gforney:


gforney:
we need to sign any executables we distribute.  we were ok with all apps except for fds2ascii.  I added a git hash to fds2ascii output (when you type fds2ascii -v )

PR #3426, opened 2016-02-02
FDS Source: Associate MPI process with range of meshes.

mcgratta:


mcgratta:
In this commit, I define two integers for each MPI process -- `LOWER_MESH_INDEX` and `UPPER_MESH_INDEX`. This indicates the range of meshes controlled by that particular MPI process. The inverse of this is `PROCESS(NM)`, which indicates the MPI process associated with mesh `NM`. In `main.f90`, we now see

```
DO NM=LOWER_MESH_INDEX,UPPER_MESH_INDEX
```

This prevents having to loop over all meshes and cycling out of those meshes not under control of the given MPI process. In running my huge cases on the Oak Ridge cluster, I have noticed the CPU time spent in `MAIN` is significant for cases where the number of meshes is in the thousands. I am trying to cut as much waste as I can from the main loop.

rmcdermo:
Looks like better code to me!

PR #3432, opened 2016-02-03
FDS Validation: Fix minor bug in script that extracts Git date.

mcgratta:


mcgratta:
This change enables `Utilities/Scripts/makegitentry.sh` to properly extract the date of a given git hash.

PR #3487, opened 2016-02-15
FDS Validation: latest results

rmcdermo:
Pushing these for Kevin because of Pink Unicorn problems.


gforney:
Would erasing repo and re- cloning fix this?

rmcdermo:
What I suggested to Kevin is erasing fork and re-forking.  I don't think this is a local problem.   I think you can actually do this without re-cloning, which would save any untracked files in his project.

Issue #3484, opened 2016-02-12
Ramp up for mass flux not working properly


mcgratta:
In this case, if you set TAU_MF to anything but zero, things work properly.

```
&HEAD CHID='scalar' /
&MESH IJK=48,48,48, XB=-2,2,-2,2,0,4 /
&TIME T_END=10. /
&RADI RADIATION=.FALSE. /
&VENT XB=-1,1,-1,1,0,0, SURF_ID='INLET' /
&VENT MB='ZMAX', SURF_ID='OPEN' /
&SPEC ID='METHANE' /
&SURF ID='INLET', SPEC_ID(1)='METHANE', MASS_FLUX(1)=10.0, COLOR='RED', TAU_MF(1)=0.0 /
&SLCF PBY=0, QUANTITY='MASS FRACTION', VECTOR=.TRUE., SPEC_ID='METHANE' /
&DEVC XB=-2,2,-2,2,4,4, QUANTITY='MASS FLOW', SPEC_ID='METHANE', ID='MASSFLOWS2' /
&TAIL /
```

CakeDelivery:
[Online Cake Delivery](http://www.cakengift.in/)
[Cake Delivery in Delhi](http://www.cakengift.in/by-city/cake-delivery-in-delhi-333.html)
[Cake Delivery in Noida](http://www.cakengift.in/by-city/cake-delivery-in-noida-335.html)

CakeDelivery:
[Cake Delivery in Gurgaon](http://www.cakengift.in/by-city/cake-delivery-in-gurgaon-334.html)
[Cake Delivery in Ghaziabad](http://www.cakengift.in/by-city/cake-delivery-in-ghaziabad-336.html)
[Cake Delivery in Faridabad](http://www.cakengift.in/by-city/cake-delivery-in-faridabad-337.html)
[Online Birthday Cake](http://www.cakengift.in/by-occasion/birthday-cakes-339.html)

CakeDelivery:
[Online Birthday Cake Delivery](http://www.cakengift.in/by-occasion/birthday-cakes-339.html)
[Order Cake in Delhi](http://www.cakengift.in/by-city/cake-delivery-in-delhi-333.html)
[Order Cake in Noida](http://www.cakengift.in/by-city/cake-delivery-in-noida-335.html)
[Order Cake in Gurgaon](http://www.cakengift.in/by-city/cake-delivery-in-gurgaon-334.html)
[...]

PR #3499, opened 2016-02-17
FDS Source: Issue #3484. Fix bug when TAU_MF=0.

mcgratta:


mcgratta:
This fix is for cases when `TAU_MF=0`. The problem was that we call `WALL_BC` during the `INITIALIZATION_PHASE` of the simulation so that certain initial conditions will be properly implemented. During this call, `CORRECTOR=.TRUE.` and `ONE_D%UW` gets set to something other than 0 because when `TAU_MF=0`, `EVALUATE_RAMP` returns 1. When the calculation actually starts, we need `UW` to be zero so that we can ramp the velocity up to whatever its value needs to be. We don't specify the normal velocity, but rather the change in the normal velocity with time, so we have to start at zero. The fix is to simply not do the specified mass flux calculation during the `INITIALIZATION_PHASE`.

PR #3525, opened 2016-02-19
FDS Source: Ensure that droplet evaporation routine does not start wi…

mcgratta:
…th a gas temp lower than TMPMIN.

### DIFF
--- a/FDS_Source/part.f90
+++ b/FDS_Source/part.f90
@@ -2134,7 +2134,7 @@ SPECIES_LOOP: DO Z_INDEX = 1,N_TRACKED_SPECIES
 
             ! Gas conditions
 
-            TMP_G  = TMP_INTERIM(II,JJ,KK)
+            TMP_G  = MAX(TMPMIN,TMP_INTERIM(II,JJ,KK))
             RHO_G  = RHO_INTERIM(II,JJ,KK)
             D_AIR = D_Z(NINT(TMP_G),Z_INDEX)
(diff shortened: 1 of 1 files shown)


mcgratta:
This change is a band-aid to fix a problem that originated with
[this commit](https://github.com/firemodels/fds-smv/commit/1304c6a7cbb1ec22920ca3fc76a9dd3c29ca85f2). I cannot figure out exactly what went wrong, but I do know that for -O2 optimization (and not lower), the verification case `activate_sprinklers.fds` generated a numerical instability in 

```
SUBROUTINE PARTICLE_MASS_ENERGY_TRANSFER(T,DT,NM)
...
IF (DT_SUBSTEP <= 0.00001_EB*DT) THEN
   CALL SHUTDOWN('Numerical instability in particle energy transport, TMP_G')
   RETURN  <--- Note that I added this. SHUTDOWN no longer shuts down the run, it only generates a stop code
ENDIF
```

The problem was that `TMP_G` for a certain droplet was something like -45 C whereas `TMP_G_NEW` was -10 C (`TMPMIN`). The temperatures locked in, so to speak, until the time step dropped below the tolerance and the run was halted. My band-aid was to just not allow `TMP_G` to get assigned a value lower than `TMPMIN`. This case is a bit extreme -- it involves a high flow rate of very small droplets that cause the gas temperature field to checkerboard. Setting `PARTICLE_CFL=.TRUE.` remedies the problem, but this is not a default setting.

Issue #3521, opened 2016-02-19
Thermocouple statistics max


mjpart:
Hi!

I have some problems with statistics measurement in FDS (version 6.3.2).
I have examined small scale experiments regarding to smoke and heat control systems (smoke duct sections) and how these components are exposed to heat in fire scenarios. In my test case (1K-K_3,0m) I have set thermocouple devices under the smoke duct and measured thermocouple statistics max inside the certain volume (just under the duct). 

Results and input file are as attachments:
-Point device (green line) measures higher temperature than statistics MAX although measuring properties are same and point device locates inside the statistics measurement volume. I have tested this in different other situations and results are similar.  How is that possible?
Statistics max should give the highest temperature possible… I haven’t noticed this phenomenom in earlier FDS versions. 

Thank you for your comments!

[1K-K_3,0m.zip](https://github.com/firemodels/fds-smv/files/137791/1K-K_3.0m.zip)

Regards
Mikko

mcgratta:
Please simplify your case. It is difficult to sort through all these lines of input to diagnose a possible problem. Create a simple test case that demonstrates the problem by removing unnecessary details from your case. For example, RAMPs, multiple MESHes, etc.

mjpart:
Hi,
Simplified input file below.

[test_case.fds.txt](https://github.com/firemodels/fds-smv/files/138668/test_case.fds.txt)

mcgratta:
The problem is with the `THERMOCOUPLE` output. When you ask for `THERMOCOUPLE`, FDS solves an ordinary differential equation that governs the heating of a small spherical bead. This calculation is only done at a single point, not multiple points when `STATISTICS='MAX'`. I have added an error flag to stop FDS whenever `THERMOCOUPLE` is used with `STATISTICS`. For the moment, do not use the two flags together.

Edward101:
I have a similar problem for MIN statistics on VISIBILITY quantity. But I did't use THERMOCOUPLE. I run it using FDS6.2.0.

Besides, I find it took a longer time to dump slices for FDS6.2.0 than FDS6.0.1. The time could be increased from 5min to 30min if the slices are dumped. Is this problem fixed in later versions?

Attached is one modification version based on the hallway example provided by FDS.
[hallways3.1.fds.txt](https://github.com/firemodels/fds-smv/files/197993/hallways3.1.fds.txt)
[...]

PR #3531, opened 2016-02-22
FDS Source: Do not allow THERMOCOUPLEs to be used with volume STATIST…

mcgratta:
…ICS.


mcgratta:
This change is in response to Issue #3521 where THERMOCOUPLE output was used with a volume STATISTIC. TCs are only appropriate for a single point value.

Issue #3522, opened 2016-02-19
Pyrosim: Urgent Help, couldn't find a solution severe (157) program exception access violation


hasanadel88:
Dear Support team and members,
Greetings
Wish you can help me getting this simulation to run. I don't know anything related to programming, I use PYROSIM GUI for that purpose. I did spend more than 40 hours trying to solve what causing severe (157) program exception access violation but without any luck. Really, i'm upset of those endless error messages we face when we are about to deliver the project to client. Don't make my life any harder. I'm eagerly waiting your reply on this.

Attachments
1. The pyrosim PSM file (Uploaded on dropbox)
https://www.dropbox.com/s/d0iu6iz8kjgmzmn/Rev2%20Basements%20B2%20On%20Fire%20Fire%20Scenario%202.psm?dl=0
2. FDS (Uploaded on dropbox)
https://www.dropbox.com/s/m8sp6yi9ksi6p5p/Rev2%20Basements%20B2%20On%20Fire%20Fire%20Scenario%202.fds?dl=0
3. Error Log (Uploaded on dropbox)
https://www.dropbox.com/s/58ojaxj3qqffkiw/Error%20Hassan.log?dl=0

Cluster configurations
2 Identical PCs joined a domain with same exact hardware
Core i7-4790 @3.60 GHZ
Ram: 16 GB
OS: Windows 7 Enterprise 64 bit
Graphic: GeForce GT 740

mcgratta:
This is not a good way to ask for help. You pay nothing for this software and you pay nothing for the support. You should act accordingly.

I cannot access Dropbox. Post a simple version of your input file that clearly demonstrates the problem. Tell me what version of FDS you are using. Do not post the PyroSim file. I will not need it.

hasanadel88:
What is meant by a simple version of the input file? Can I spotlight which part of code has an issue? 

 The error log
Starting FDS: fds.exe...

 Mesh      1 is assigned to MPI Process      0
 Mesh      2 is assigned to MPI Process      1
 Mesh      3 is assigned to MPI Process      2
 Mesh      4 is assigned to MPI Process      3
 Mesh      5 is assigned to MPI Process      4
 Mesh      6 is assigned to MPI Process      5
 Mesh      7 is assigned to MPI Process      6
 Mesh      8 is assigned to MPI Process      7
 OpenMP thread   1 of   3 assigned to MPI process      0 of      7
 OpenMP thread   2 of   3 assigned to MPI process      0 of      7
[...]

PR #3538, opened 2016-02-22
FDS Source: Do not allow an HVAC VENT on a thin OBSTruction.

mcgratta:


mcgratta:
This addresses Issue #3522.

PR #3549, opened 2016-02-24
FDS Source: Ensure reasonable CPU output for stopped jobs.

mcgratta:


mcgratta:
This change ensures that even if the job is stopped at T=T_BEGIN (set up only), the CPU times are not wrong.

gforney:
thanks.  I added a FDS_Timing_Cases.sh script in Verification where you can add a "Dow Jones" list of cases (need to be from FDS_Cases.sh)  .  I put in a placeholder list to test my timing archiving script changes.

PR #3557, opened 2016-02-25
FDS Source: fix BCOR for multiple fast reactions

rmcdermo:
Do not accept this pull request until I test with Firebot.


rmcdermo:
This passed all tests related to reactions.  There is one warning from HVAC door_crack, very slightly out of tolerance.  Jason, please check.

PR #3602, opened 2016-03-03
FDS Verification: Set initial time step for soot_grav_settling case.

mcgratta:


mcgratta:
This is to bring a the soot_gravitational_settling case into tolerance. I just lowered the initial time step to 0.05 s.

Issue #3483, opened 2016-02-12
Multi-mesh head flux gauge


rmcdermo:
Kevin,

Here is an email I got from James White at UMD.  Can you please have a look?

---

Hi Randy,

I’m having trouble getting heat flux diagnostics to work in the latest FDS builds. If I use the surrogate target input lines (pg 215 of the user guide), the devc.csv output only prints a single row at the initial time step and then doesn’t print any more results (including for all other devices that may be in the input file). The RADIATIVE HEAT FLUX GAS shortcut seems to work fine. Maybe I’m missing something, but can you check this on your end, let me know if it works for you?

Thanks,

James White

[HEAT_FLUX_GAUGE.fds.txt](https://github.com/firemodels/fds-smv/files/128626/HEAT_FLUX_GAUGE.fds.txt)

mcgratta:
The particle for the surrogate target is sitting on the boundary between two meshes. There's a bug related to this. For the moment, just move the particle a tiny bit one way or the other.

jameswhite255:
Thanks Kevin, great catch. I can confirm that adjusting the Y coordinate on the INIT line resolves this issue.

CakeDelivery:
[Online Cake Delivery](http://www.cakengift.in/)
[Cake Delivery in Delhi](http://www.cakengift.in/by-city/cake-delivery-in-delhi-333.html)
[Cake Delivery in Noida](http://www.cakengift.in/by-city/cake-delivery-in-noida-335.html)

CakeDelivery:
[Cake Delivery in Gurgaon](http://www.cakengift.in/by-city/cake-delivery-in-gurgaon-334.html)
[Cake Delivery in Ghaziabad](http://www.cakengift.in/by-city/cake-delivery-in-ghaziabad-336.html)
[Cake Delivery in Faridabad](http://www.cakengift.in/by-city/cake-delivery-in-faridabad-337.html)
[Online Birthday Cake](http://www.cakengift.in/by-occasion/birthday-cakes-339.html)

CakeDelivery:
[Online Birthday Cake Delivery](http://www.cakengift.in/by-occasion/birthday-cakes-339.html)
[Order Cake in Delhi](http://www.cakengift.in/by-city/cake-delivery-in-delhi-333.html)
[Order Cake in Noida](http://www.cakengift.in/by-city/cake-delivery-in-noida-335.html)
[...]

PR #3606, opened 2016-03-04
FDS Source: Ensure that INIT XYZ does not fall on mesh boundary.

mcgratta:


mcgratta:
This commit addresses a bug reported in Issue #3483. If a particle that is being used as a measurement device falls on a mesh boundary, we move it a small distance from the boundary so that only one MPI process will work with it.

PR #3660, opened 2016-03-12
FDS Verification: fix bound_test CHID; hrrpuv_reac_arrhenius tolerance

rmcdermo:


rmcdermo:
I hope the CHID problem with the bound_test cases is what caused the subsequent errors.  I did not see anything else wrong with the dataplot input lines.  I looked at them in vi and Sublime... no line ending problems or anything.  Somehow we are not gracefully skipping cases like we think.

PR #3746, opened 2016-03-31
FDS Verification Guide: Add more write up on water_evap_6 case.

mcgratta:


mcgratta:
Randy, Jason -- take a look at the write-up for `water_evaporation_6`. It's a more detailed discussion of how we get the exact answer. Not sure if we can use this template directly for the other cases since this one has simple, constant properties.

rmcdermo:
Kevin,
Looks good.  I made one minor edit; some time back we scrubbed the manuals
of {\cal R} for gas constant and {\cal H} for pressure.

mcgratta:
Yeah, I thought R looked a bit over the top. Better for a wedding invitation. Thanks.

drjfloyd:
In FDS R=8.314472, Should the R=8.314 use the approximate symbol so we don't imply FDS has R=8.314?

Aspects of the template could be used for the other cases. Probably worth expanding the discussion a little on how the solutions were obtained.

mcgratta:
OK. I did not change the analytical solution. I assume that someone (Jason or Randy?) has done what I did. I checked the solution with a calculator, cutting off digits as I went along. I could do it out more formally to check -- or does a script already exist for this case?

drjfloyd:
I did a calc for this case back in 2012 when it is was first pushed up. I assume I just used Excel for that.

rmcdermo:
I think it is fine as is.  That is, we do not need \approx just because we
are not using full precision.

PR #3779, opened 2016-04-11
FDS Source: Use sub time steps for fast particles. Add accuracy metri…

mcgratta:
…cs for flat_fire test case.


mcgratta:
This change to the particle transport routine ensures that if a particle's CFL number exceeds 1, the time step to move it will be reduced accordingly. This is somewhat like `PARTICLE_CFL` except that it does not force the gas phase time step to be decreased -- only the time step to move the given particle. I also moved the `flat_fire` processing to `dataplot` to make sure we catch errors. My change broke `flat_fire` but I did not notice until I looked at the plot.

rmcdermo:
Nice work.  I think this is a big improvement.

tstopi:
This broke VTT_Sprays (and any fats spray case where PARTICLE_CFL=.FALSE.).  I submitted a pull request #3792 to fix the issue. Maybe one of you could take a look at it.

Issue #3807, opened 2016-04-25
Baroclinic torque in tunnel geometries


mcgratta:
This is a simple case that fails unless `BAROCLINIC` is set to `.FALSE.`

```
&HEAD CHID='tunnel_2d' /
&TIME T_END=10. /
 MISC BAROCLINIC=.FALSE. /  <-- This line needed to prevent numerical instability
&MESH IJK=768,1,16, XB=0.0,12.0,-0.01,0.01,0.0,0.25 /
&REAC FUEL='PROPANE' /
&SURF ID='fire.source', COLOR='RED', HRRPUA=2126.0 /
&SURF ID='entrance', VEL=-0.3 /
&VENT SURF_ID='fire.source', XB=5.955,6.044,-0.01,0.01,0.0,0.0, COLOR='RED'/
&VENT SURF_ID='entrance', PBX=0.0, COLOR='BLUE' /
&VENT SURF_ID='OPEN', PBX=12.0 /
&SLCF QUANTITY='TEMPERATURE', VECTOR=.TRUE., PBY=0.0/
&SLCF QUANTITY='DIVERGENCE', CELL_CENTERED=.TRUE., PBY=0. /
&SLCF QUANTITY='H',          CELL_CENTERED=.TRUE., PBY=0. /
&SLCF QUANTITY='HRRPUV',     CELL_CENTERED=.TRUE., PBY=0. /
&TAIL /
```

mcgratta:
I can run the above case successfully if I call `BAROCLINIC_CORRECTION` from inside the `PRESSURE_ITERATION` loop in `main.f90`, just before I call `NO_FLUX`. The purpose is to more tightly couple `H` and `KRES` when computing the pressure in the baroclinic torque term. 

There are still things to work out, like what to do in cases where we do not have to iterate the pressure solver.

mrvlty:
Mr. Mc Grattan ,
I can't understand your last advice exactly. Could you please clarify it? 
Now I am installing the older version of fds as Daniel Swenson advised. However I want to implement your advice too to compare the results of them. This study will be a part of my master thesis so finding a best solution method is really important for me. 
Thank you for your support.

mcgratta:
[...]

PR #3833, opened 2016-05-10
FDS Source: Major change related to Issue #3807. Iterate baroclinic term in pressure solver.

mcgratta:


mcgratta:
The baroclinic torque term, p grad(1/rho), is now updated as we iterate on the pressure.

PR #3895, opened 2016-05-27
FDS Source: HVAC init averaging for tmp and rho down duct

bmralph:


rmcdermo:
Ben,
You can do this averaging more efficiently.  We'll go over it and you can submit a new pull request.

bmralph:
Cool!:) I had a sneaky feeling one could as just
I was staring at it on Friday...

PR #3908, opened 2016-06-02
CC_IBM combustion overhaul

rmcdermo:
I have revamped the COMBUSTION_MODEL to be independent of Cartesian cell (I,J,K).  So, all quantities are passed to the routine.  The point of this is to enable use of exactly the same combustion routine for the cutcell immersed boundary method (CC_IBM).


rmcdermo:
Do not merge.  I am testing first.  Thanks.

rmcdermo:
Found some unused variables.  Plus need to rerun on blaze.

PR #3937, opened 2016-06-10
FDS vv: add C2H2 experimental radiative distribution data downwards and sideways

ahaminsgithub:


ahaminsgithub:
first pull request ! files will not affect firebot - glenn promises

rmcdermo:
Anthony,
This is great!  Thanks!  However, we really don't want to commit docx and rtf files to the repo.  Please convert to md (Markdown) or latex.  I can help.  In the mean time, you could just submit your .csv experimental files.  This is a good lesson in committing small chunks instead of too many things at once.  The reason I don't just accept what you have and then delete your other stuff is that it unnecessarily increases the size of the repo.

PR #3972, opened 2016-06-14
Lua scripting: added functionality.

JakeOShannessy:
Exposes more commands/functions/variables etc. to the Lua interpreter.

This PR looks quite large but is simply the addition of many more options (that are usually exposed via the existing scripting system, the .ini file, or via the menus). It is written in a more verbose manner in order to favour simplicity.

These changes are predominantly to the isolated files affecting the Lua interpreter (c_api.\* and lua_api.*), there is only one significant addition to the core of Smokeview which is another version of the rendering subroutine to render to a variable instead of to a file (in renderfile.c), allowing the Lua interpreter to use the image data directly. This duplicates some code in order to avoid modifying the existing, and puts it under the `pp_LUA` directive. All changes are made under this directive or isolated in other files.


mcgratta:
Glenn is on vacation this week, so he may not be able to look at this until he returns next week.

gforney:
passed smokebot and code changes are isolated with pp_LUA directives

PR #4003, opened 2016-06-17
add modified exp. data

ahaminsgithub:
1. please scrutinize this pull request before accepting it - it seems like some extraneous files are staged for the pull
2. please note that the plotting routine appears to not properly plot cases when the independent variable (radial location) is slightly negative. 

thanks


mcgratta:
I have closed the pull request. I will wait for all the changes to the experimental data files.

PR #4014, opened 2016-06-20
FDS Source: Improve interpolation scheme for particle momentum.

mcgratta:


mcgratta:
I re-ran the spray cases VTT, BRE, LEMTA, and I re-ran the WUI particle cases NIST_Douglas_Firs and CSIRO_Grassland. No significant changes in the results.

PR #4045, opened 2016-06-29
FDS Source: makes DUCT_INTERP_TYPE inputable in HVAC namelist

bmralph:
Users can now select inter-node interpolation scheme (1, 2, default). Tested on verification cases with mass transport on and off.


drjfloyd:
For being able to more easily understand the input file, I think it would be better for the input variable to be a text string that is then converted to the appropriate integer type during input processing.  Maybe 'NODE1' 'NODE2' 'LINEAR INTERPOLATION'   So you would read in the text string and then due a SELECT CASE on the string to set the DU%DUCT_INTERP_TYPE

bmralph:
Agreed. This was a quick n dirty interim just to help me set up
verification cases. I'll get on it.

PR #4047, opened 2016-06-30
FDS source: duct interp type string based user input added

bmralph:
Tested with `HVAC_MASS_TRANSPORT` as `TRUE`and `FALSE`.

Requires user to input `DUCT_INTERP_TYPE` if `HVAC_MASS_TRANSPORT` is `TRUE` - otherwise returns an error. i.e. there is no default behaviour for interpolation - @drjfloyd: let me know if you think there should be a default.


bmralph:
Forgot to run through debug compiler - closing this pull request and will re-pull request shortly.

PR #4079, opened 2016-07-12
FDS Source: handle multi-mesh cases with EXTERNAL_BOUNDARY_CORRECTION

rmcdermo:


rmcdermo:
With the current code we still have the problem of mesh boundaries that are a combination of INTERPOLATED (Dirichlet) and NEUMANN (UW=0).  For now, the user must reconfigure their mesh to eliminate completely this error.  Thus, the current version "does no harm" to this case, but it does solve the problem of velocity errors on mesh boundaries that are OPEN (Dirichlet) + NEUMANN (UW specified), which at least is an improvement.

PR #4092, opened 2016-07-14
FDS Source: Make SECOND_ORDER_INTERPOLATED_BOUNDARY the default.

mcgratta:
This variable is no longer an input. It is done automatically.


rmcdermo:
Huge!

Issue #4122, opened 2016-07-27
Issue/Solution for VTT Validation using PARTICLES_PER_SECOND = 5000 


ChangminCao:
**Issue Description:**
The FDS default PARTICLES_PER_SECOND = 5000 does not generate a droplet size distribution that is specified in the input file. 
Figure 1 shows plot that were created from running the VTT Water Spray Validation input file LN02_1.fds with the PARTICLES_PER_SECOND = 5000 (see bottom plot). 
![default model results](https://cloud.githubusercontent.com/assets/20332448/17167010/16c5355a-540d-11e6-9a96-5f353d9184bd.png)

The reason for this issue is that the particle insert number per time step is too small to generate the desired droplet size distributions.

**Solution:**
One proposed method to solve the issue is to increase the number of particles inserted per time step by setting a minimum N_INSERT limit (by defining a new variable MIN_N_INSERT) in the particle injection process. For example when MIN_N_INSERT =7, particles will not be injected into the domain until the number of particles inserted per time (N_INSERT) reaches 7, which occurs after two or three time steps.  
Below plot is droplet size distributions using modified spray injection model.
 ![modified model](https://cloud.githubusercontent.com/assets/20332448/17167026/31342f72-540d-11e6-8602-056c68d96c66.png)

You may find the detailed information about the issue description, results discussion, solution and improved results in [FDS spray injection model issues.pdf](https://github.com/firemodels/fds-smv/files/385465/FDS.spray.injection.model.issues.pdf)

In order to use the new approach, the user should modify the following three files, read.f90, part.f90, devc.f90. See detailed information in [FDS source code modification.pdf](https://github.com/firemodels/fds-smv/files/385468/FDS.source.code.modification.pdf)

rmcdermo:
Changmin,

Thank you for submitting this.  This looks like a reasonable solution to me.

@mcgratta @drjfloyd Please confirm and I will implement.  Thanks.

drjfloyd:
Whether we insert 1 particle per step or 1,000,000 particles per step, over a sufficiently long period of time the mass distribution of the particles at the moment they are inserted should converge to the correct size distribution.  If that is not happening, doesn't that suggest we are doing something wrong in picking from the distribution?

mcgratta:
Could you post the input file that you are using to generate the distributions above.

rmcdermo:
@ChangminCao,
[...]

PR #4128, opened 2016-07-28
Thanks

ChangminCao:
Modified FDS Spray Injection Model for using Less PARTICLES_PER_SECOND=5000 with fine mesh (1cm)


rmcdermo:
@ChangminCao 
I think you meant to reply to Issue #4122

PR #4223, opened 2016-08-19
FDS vg: when building VG, copy aux file from UG dir before build and …

gforney:
…copy VG pdf file to UG dir after build


gforney:
kevin,
I put both guides in the user guide directory (when you build the VG guide).  I also need to modify the bundle generating scripts (to put manuals in the right directory) and the windows batch files (to do the same things I did in the .sh scripts)

check out my changes and let me know if you see any problems.

PR #4320, opened 2016-09-13
FDS Source: choose max local spec based on face average composition

rmcdermo:
I ran this through FDS_Cases.sh manually and all was fine.

@mcgratta I'd like to add OpenMP to these loops.  Could you take a look at this?


mcgratta:
Why not this?

```
!$OMP PARALLEL DO PRIVATE(N) SCHEDULE(STATIC)
```

rmcdermo:
I pushed up OMP code.  Hopefully it works.  Actually, _db compiler catches some mistakes with OMP syntax, which is nice.

Issue #4326, opened 2016-09-15
FDS crashes for MPI cases with more than two meshes and periodic boundary conditions


rmcdermo:
I have a large case I am trying to run with obstacles in the center of the domain.  To optimize the run,  I have sized the mesh to surround these obstacles. I was having issues with periodic boundary conditions.  To explore the problem I whittled the problem down to this simple case.

&HEAD CHID='SIMPLE_CASE', TITLE='Test Simulation'/

// Level 1 (Bottom)
&MESH IJK=  256, 64, 8,        XB=-236,   20, -96, -32,   0,   4.5/
&MESH IJK=  256, 64, 8,        XB=-236,   20, -32,  32,   0,   4.5/
&MESH IJK=  256, 64, 8,        XB=-236,   20,  32,  96,   0,   4.5/
!MESH IJK=  256, 64, 8,        XB=-236,   20,  96, 160,   0,   4.5/

&TIME T_END=40. /

&MISC CFL_MAX=0.5, TMPA=24.11, P_INF=94200, RADIATION=.FALSE./      

&VENT MB='XMIN', SURF_ID='WIND'/
&VENT MB='XMAX', SURF_ID='OPEN'/

&SURF ID ='WIND'
    PROFILE='ATMOSPHERIC', Z0=3, VEL=-6.72, PLE=0.05, 
    TMP_FRONT= 24.11/ 

&VENT PBY= -96, SURF_ID='PERIODIC'/
&VENT PBY=  96, SURF_ID='PERIODIC'/

If I run the above case with more than 2 meshes in the y-direction, the MPI process fails with a segmentation fault.  If I run the case on a single processor (with just the 'fds' command), there are no issues no matter how many meshes are used.  Additionally, if I force the meshes on either side of the periodic boundary to be on the same MPI Process (by setting MPI_PROCESS), I can run the case with periodic boundary conditions regardless of the number of meshes; however, I have noted abnormalities in the flow at mesh boundaries when using this work around.  FDS works normally if I don't apply periodic boundary conditions on the y-faces.

I updated from 6.5.1 to 6.5.2, but it made no difference.  I am using Open MPI v1.8.4.   Does anyone else have this issue? Or is this a problem localized to my install?   I have run many successful FDS simulations with MPI in the past and the MPI install I am using is on a heavily used server, so it seems relatively unlikely (but still possible) this is due to a poor install of MPI.

Thanks in advance!

rmcdermo:
[...]

PR #4327, opened 2016-09-15
FDS Source: Issue #4326. Add PERIODIC_MESH_IDS to MESH line for PERIO…

mcgratta:
…DIC multi-mesh cases.


rmcdermo:
Kevin,
Any reason you used
`PERIODIC_MESH_IDS = 'pnull'`
or is this a typo?  If you have anything where you cycle on `mesh name /='null'` this could be a problem.

mcgratta:
The name must be distinct from the MESH ID or else they will be matched.

rmcdermo:
Ah, got it.

PR #4393, opened 2016-10-06
FDS User Guide: add section on HT3D (Beta)

rmcdermo:
Kevin and Jason,
Please have a read of this section (it is short) and make sure what I have written is accurate and clear.  Thanks!


mcgratta:
I think we should always use `HT3D=.TRUE.` rather than `HT3D=T`, just to be consistent with the rest of the guide.

We should spell out what the VN restriction is, i.e. dt<...

We should limit the `SURF` line to a single material. In fact, should we use `MATL_ID` instead? I think that the `SURF_ID` could lead to confusion and inconsistency, for example, `THICKNESS`, layers, etc. We can also restrict the material of the `MATL_ID` to be non-reacting.

PR #4423, opened 2016-10-12
FDS Source: Fix asymmetrical behavior of thin obstructions.

mcgratta:


rmcdermo:
Kevin,
I'm not sure I understand this line:
IF (.NOT.THIN_OBSTRUCTION .OR. UU(IIG-1,JJG,KKG)>=0._EB) FX(IIG-1,JJG,KKG,N) = WC%RHO_F*WC%ZZ_F(N)

I think UU can be + or -, so we need something like ABS(UU)<TWO_EPSILON_EB.

I think what you are trying to do is "if we are on a thin obstruction, only set FX if UU == 0", right?

mcgratta:
UU will never be exactly zero on a thin obstruction. So we have to decide which side of the obstruction is to be used to set FX.

rmcdermo:
But why is there a preference for one direction depending on IOR?

PR #4460, opened 2016-10-27
FDS Source: safer way to ensure T_END at output, remove possibility o…

rmcdermo:
…f inf loop in dump


gforney:
ran smokebot on blaze - all cases finish now

PR #4474, opened 2016-11-01
FDS Tech Guide: More edits of vegetation particles.

mcgratta:


rmcdermo:
Suggestions:
1. I don't think m''' is used, so we can avoid adding it to the notation.
2. I think it is best to avoid rewriting the force law.  I would not write eq. (8.7).  Rather, take the parameters in this section and write down two equations, one for C_d and one for A_p that go into eq. (8.1).

mcgratta:
m''' is the actual input from WFDS -- `VEG_BULK_DENSITY`, so I want to define this here so we see how it relates to beta, which is not an input.

I like the drag law because it's what you see in these WUI papers. If we just add conversion formulae, I think it will be harder to decipher.

PR #4491, opened 2016-11-09
FDS Verification: Adjust OpenMP test tolerance.

mcgratta:
I've created a new success criteria for the OpenMP test case.


rmcdermo:
Excellent.  I like tightening the tolerance!

mcgratta:
Yes, but at the cost of 4 additional cores. I still think there's some fat to be cut.

Issue #4511, opened 2016-11-28
.out file properties


drjfloyd:
In the .out file we write properties at TMPA but label it 293 K.   This should be fixed.  Should we write at 293 K and not write TMPA, write both, or just write TMPA?  If we do write TMPA should we insure that it is writing in the proper numerical order with respect to the other values?

mcgratta:
I think we should just write it out at TMPA. There's nothing particularly meaningful about 293 K.  I wouldn't worry about the order -- that would just add a lot of code for something very minor.

drjfloyd:
OK. I'll get this done. Thanks for the input.

drjfloyd:
done

PR #4512, opened 2016-11-28
FDS Source: Issue #4511 make 293 K => TMPA in .out file properties

drjfloyd:


rmcdermo:
Jason,
I think we should actually say what TMPA is.

rmcdermo:
Also, you have not merged this yet.  But you don't have to.  You can just make the changes and this pull request will stay intact.

drjfloyd:
Will do.

PR #4526, opened 2016-12-04
fixing scripts

RIO4FDS:


mcgratta:
If you are in the fds repo, the Build directory is spelled with a capital B. Things can get very tricky with git when you confuse lower and upper case letters. Check the repo and fix if necessary, then we will merge.

RIO4FDS:
I am sorry! My mistake. Fixed!

Issue #4549, opened 2016-12-12
Asymmetries occur with thin obstructions


mcgratta:
Run this case with `CONSTANT_SPECIFIC_HEAT_RATIO=.TRUE.` and it remains symmetric in the x=constant plane. Set to `.FALSE.` and asymmetries arise quickly. I suspect that the transport terms in the divergence are not symmetric when there is a thin obstruction.
```
&HEAD CHID='RSE_15', TITLE='NIST Reduced Scale Enclosure Experiments, 2007' /

&MESH XB= 0.00,1.42,0.00,0.95, 0.00,0.98, IJK = 48,32,32 /

&TIME T_END=4. /

&REAC FUEL='N-HEPTANE' /

&MISC NOISE=.FALSE., STRATIFICATION=.FALSE., CONSTANT_SPECIFIC_HEAT_RATIO=.TRUE. /

&RADI RADIATION=.FALSE. /

&SURF ID='WALL', MATL_ID='GYPSUM', THICKNESS=0.025, COLOR='ANTIQUE WHITE', DEFAULT=.TRUE., FREE_SLIP=.TRUE. /
&SURF ID='STEEL SHEET', MATL_ID='STEEL', THICKNESS=0.00635, COLOR='BLACK', FREE_SLIP=.TRUE. /
&SURF ID='fire', SPEC_ID(1)='N-HEPTANE', MASS_FLUX(1)=0.0082, COLOR='ORANGE', TMP_FRONT=100. /

&OBST XB=0.54,0.89,0.31,0.66,0.00,0.002, SURF_IDS='fire','STEEL SHEET','STEEL SHEET' /
&OBST XB=0.54,0.54,0.31,0.66,0.00,0.12, SURF_ID='STEEL SHEET' /
&OBST XB=0.89,0.89,0.31,0.66,0.00,0.12, SURF_ID='STEEL SHEET' /
&OBST XB=0.54,0.89,0.31,0.31,0.00,0.12, SURF_ID='STEEL SHEET' /
&OBST XB=0.54,0.89,0.66,0.66,0.00,0.12, SURF_ID='STEEL SHEET' /

&VENT XB=0.0, 0.0, 0.24, 0.72, 0.00,0.81, SURF_ID='OPEN' / 

&MATL ID            = 'STEEL'
      SPECIFIC_HEAT = 0.77
      EMISSIVITY    = 0.8
      DENSITY       = 7800.
      CONDUCTIVITY  = 45. /

[...]

PR #4550, opened 2016-12-12
FDS Source: fix asymmetry in divg corrections for thin obstructions

rmcdermo:


rmcdermo:
This solution is to address Issue #4549.  There is hopefully a better way to do this that avoids IF statements inside the 3D loop.

PR #4557, opened 2016-12-13
FDS+Evac: Completion of the column for the initial positions of the agents

gjaeger:
I missing the last column of the table "Evac: Initial positions of the agents" in the evac out-file. I have added the code accordingly.

I hope my addition to a pull request is the right way. Or should I create an issue?


rmcdermo:
Hi Gregor,
Thanks.  I see no problem.  Firebot will hopefully catch anything tonight.

PR #4614, opened 2017-01-07
Remove Verification/Scripting directory.

JakeOShannessy:
This directory is only relevant to smokeview, which is no longer
part of this repository.


gforney:
Thanks!
Glenn

PR #4638, opened 2017-01-17
Matlab: set Figure_Visibility default to off

rmcdermo:


rmcdermo:
Glenn,

Please check that this does not have any impact on Firebot.  I thought at one point Kris was having Firebot change this setting to off to avoid pushing figures to the screen.  Maybe there is a more graceful way you are handling that now.

gforney:
I haven't come across this yet.  I know that figures go to the screen in the PC bots.  If something goes wrong when the next firebot runs I'll take a look.

PR #4646, opened 2017-01-18
FDS User Guide: input file exerpt correction

bmralph:
Updated to reflect FDS 6's device species quantity method.


rmcdermo:
Good catch.  Thanks!

PR #4651, opened 2017-01-19
merge requests

gforney:
don't know why I'm getting all these merge requests that need to be incorporated via pull requests


rmcdermo:
I wanted to talk to you about this, but did not want to ruffle your
feathers.  You should not need to do this.  Let's talk about it when you
get time.

bmralph:
From my experience this happens if someone has made and pushed commits in
between your last merge and your push of new commits.

PR #4720, opened 2017-02-10
Directory structure post-reorg (for review?)

bmralph:
I needed to alter this portion of hard coded directory as clean.sh was looking in <myGitLocation>/fds/fds/Build/

Changes directory as $GITROOT includes /fds/ hence there is an extra /fds/ when cleaning.


bmralph:
Did this affect anyone else? I rarely use this script so hadn't found the problem until now.

mcgratta:
Where is $GITROOT defined?

bmralph:
In this script.

PR #4740, opened 2017-02-18
FDS Source: rework 1D-3D heat transfer coupling

rmcdermo:


rmcdermo:
This commit has a RANDOM_UNUSED_VARIABLE declared in wall.f90.  We'll see if firebot catches it.

PR #4747, opened 2017-02-21
FDS Source: Remove uninitialized variables etc.

mcgratta:


mcgratta:
FYI, these uninitialized variables were found by the Gnu Fortran compiler.

PR #4765, opened 2017-02-28
FDS guide: parameter typo, FDS 6 quantity correct

bmralph:
Typo correction (DT LIMITING RATIO vs LIMITING DT RATIO) and FDS 5->6 output correction.


rmcdermo:
Thanks, Ben!

PR #4798, opened 2017-03-07
Add vegetation chapter to User Guide

tmilac:
This pull request is only for Manuals/FDS_User_Guide/FDS_User_Guide.tex.  Are you able to pull files individually from my fork?


rmcdermo:
Tom,

You're asking us to ADD all the SCRIPT_FIGURES that you used to compile the guide.  This is not what we want to do.  The answer to your question above is no.  You can only merge branches in git, not individual files within branches.

You will need to do a git reset to recover from this.  If you are not familiar with that, what I suggest you do is to save your FDS_User_Guide.tex (if that is what you want to submit) and do a fresh clone of the latest fds repo.  Then from now on, never do a git add --all.  Always add individual files or a git add -u, which will only add tracked files.

Sorry for the learning curve.  Let me know if you have other questions.

Randy

PR #4807, opened 2017-03-08
Introduce new verification case in Vegetation subdirectory

tmilac:
This verification case is derived from WUI/pine_needles.fds, modified to reflect conventions used by the wild land fire community.  I have introduced subdirectory Verification/Vegetation to hold the case, and a `Vegetation' chapter in the Verification Guide to display the data generated by the case.

vege_mass_conservation *does not* run correctly in that total mass is not conserved throughout the run.  The total mass is zero at time zero, and asymptotes to the correct value 1.0 as time progresses.  Am I using the correct quantities in the DEVC?


mcgratta:
Tom

Thanks. A few things. First, I don't think we need a separate matlab script for this case. We can plot the columns in the _devc.csv file directly. Use pine_needles as an example of how to add a case to `FDS_verification_dataplot_input.csv`. The advantage of using dataplot.m is that we automatically check all the cases to make sure that they are within tolerance. We have no way (at the moment) to check the accuracy of the individual matlab scripts. We end up having to do this manually. Second, the total mass appears to be calculated as a running average. Add `TIME_AVERAGE=.FALSE.` to the `&DEVC` lines in the input file. Also, no need to create a new folder for each verification case. This case could have gone in the WUI folder, but because I like the name Vegetation better, I'll migrate my cases over.

rmcdermo:
Adding TIME_AVERAGED=F does not fix the problem.  I have a feeling I have a bug related to applying this to DEVC with VOLUME INTEGRAL.

mcgratta:
OK, we'll fix the time averaging. Just deal with the dataplot issue. Thanks.

rmcdermo:
Actually, not my fault.  The culprit is this thing called RUN_AVG_FAC=0.5.  I'll leave this up to @mcgratta 

You can trace this by following MPUV in GAS_PHASE_OUTPUT, which points you to AVG_DROP_DEN, which gets updated with this RUN_AVG_FAC.

tmilac:
Hello Kevin,
Are the meaning of the data columns of the 
FDS_verification_dataplot_input.csv file
documented somewhere?  The info in the `Verification Case Setup Example' 
document
isn't sufficient, and I can't figure out what's going on from the 
pine_needles.fds
entry.  In particular, I need to be able to do calculations (a sum) 
across output
columns of the DEVC file.

Thanks,
Tom

rmcdermo:
Tom,
Unfortunately, dataplot is not well-documented.  If summation is necessary, there are two options, either you can do this with a DEVC inside FDS (but I never do this, I'd have to ask Jason) or you can create a pre-processing script in Matlab.  This may seem like what you are already doing, but it is still advantageous to have the plot made by dataplot, as this automatically gives error metrics and adds to other stats we keep in the guide.

We will have to answer specific questions about the dataplot columns until we get better documentation of it.

drjfloyd:
You can do this by having a SUM CTRL function that sums the DEVCs and then add another DEVC to report out the value of that CTRL function.

&CTRL ID='Sum',TYPE_ID='SUM',INPUT_ID='DEVC 1',...'DEVC N'/
&DEVC ID='Sum Output',CTRL_ID='Sum',QUANTITY='CONTROL VALUE'/

rmcdermo:
Thanks, Jason!

drjfloyd:
Note there is a limit of 40 inputs to single control function, so if you need to sum more than 40 DEVC, you will need to have multiple SUM functions: a set of SUM functions to sum groups of DEVC each having less than 40 inputs and then a SUM function that sums the previous SUM functions.

mcgratta:
Documentation on dataplot is in

https://github.com/firemodels/fds/wiki/Validation-Case-Setup-Example

Also in the Verification-Case_Setup

tmilac:
I'm working with the Validation-Case-Setup-Example documentation.
It explains adding a Matlab script clearly, which is one reason I
introduced such a script. The documentation for dataplot is incomplete.
I'm basically just guessing right now.

rmcdermo:
Tom,
I understand it is not easy to decipher.  At the moment, the best I can do is point you to the define_drow_variables.m script and then to look at how those variables are used in dataplot.m itself.  Believe it or not, this is how I still deal with all the columns when I need to sort something out.  In short, the d1 data is for the experiment or exact values and the d2 data is for the FDS results.

PR #4812, opened 2017-03-10
Syntax error in vege_mass_conservation entry in FDS_verification_dataplot_inputs.csv

tmilac:
FDS is complaining of a syntax error in the d1 columns for the vege_mass_conservation case.  There are other problems as well.  Would you be able to view what I've done via this pull request and help me past this? What I'm trying to do is to check if the total_mass column of the vege_mass_conservation_devc.csv file is equal to 1.0 at every time point.  Thank you!


mcgratta:
OK, I'll take a look at it.

mcgratta:
I'm locked out of Matlab right now because of limited site licenses, but I noticed that you are using \_ for the names of the data columns. We only use \_ for the labels on the plots, which are done with Tex or LaTeX. See if that helps.

PR #4813, opened 2017-03-10
FDS Verificaiton: Minor tweaks to vege entry.

mcgratta:


mcgratta:
Tom, try it now and let me know if it works. You can see what I did by looking at the diff or old and new.

PR #4820, opened 2017-03-13
vege_mass_conservation test complete but with errors

tmilac:
Test adapted to `dataplot' scheme.  Test fails.  Total mass is not constant across time.  Included TIME_AVERAGED=.FALSE. in DEVC lines.


rmcdermo:
Tom,
Thanks for fixing the errors.  Is the mass error still the one I mentioned related to RUN_AVG_FAC*AVG_DROP_DEN ?  Or is there another problem?

tmilac:
Hello Randy,
Yes.  You mentioned this in response to Kevin when he
suggested TIME_AVERAGED=.FALSE. would fix the problem.
As you predicted it does not.

Tom

rmcdermo:
OK, this should not be tough to fix.  Kevin will look at it once he is done with experiments.

mcgratta:
I just committed a revision to `vege_mass_conservation.fds` in which `RUN_AVG_FAC=0.` on the `MISC` line. This parameter is old and obscure -- it was originally intended to smooth the mass distribution of particles and droplets for radiation absorption calculations. I think we should make it a function of the particle type, so that solid, stationary particles do not undergo smoothing. I'll do that, but for now, use this as a place holder.

mcgratta:
This issue should be resolved now.

PR #4847, opened 2017-03-20
FDS user guide: typo

bmralph:
Fixed naming to be in line with rest of document.


mcgratta:
Thanks.

rmcdermo:
Hey guys, I can't tell what was changed---could you point me to the typo.  Also, it seems like there is still a "decrease reduce" typo.  Maybe that was the problem?

sbenkorichi:
FDS timestep  >>> time step

PR #4870, opened 2017-03-25
FDS UG, subsubsection, example case: spray_burner

gjaeger:
I found a minor error in the subsubsection: _example case: spray_burner_ in the UG. The parameter *CLASS* no longer exists in the *PROP*-namelist. The example file has already been corrected.


mcgratta:
Great, thanks,

Kevin

PR #4941, opened 2017-04-10
FDS Source : Add triangle edge tagging for special polygons. 

marcosvanella:
Glenn, I added the tagging for FLAG=.TRUE. in TRIANGULATE. 

You can test geom_test for more challenging polygons now.


rmcdermo:
@marcosvanella Please also get rid of the PAUSE statement in geom.  Thanks.

PR #4951, opened 2017-04-12
FDS Source: Update to solid cut-face calculation, boundary slices TRIANGULATE.

marcosvanella:
This update is to make boundary slices definition for Smokeview more robust. It contains:
 
- Zero area and repeated segments discard test cases for solid side cut-faces/edges. 
- Drop vertices in TRIANGULATE when they lay in the line connecting neighbors.


rmcdermo:
@marcosvanella I invited you as FDS developer so that you should now be able to merge these requests yourself.  As you know, let's be careful when commits overlap other source files, we may need to run firebot before merging.

PR #4954, opened 2017-04-13
	FDS Users Guide: Add brief two-step extinction discussion

drjfloyd:


rmcdermo:
@drjfloyd Thanks for the writeup.  You note that when EXTINCTION=T after evaluating the set of series reactions, then the CO reaction is evaluated using the CO CFT.  In looking at the code, it seems that CFT is taken from the value in the input file, correct?  In the cases, I have run to date, I was just using the same value as the first reaction.  Is there a better set of values for CO CFT and AIT I should use.  I'll check the SFPE handbook.  But wanted to double check that this is what you had in mind---that the CFT and AIT should be different for the two reactions.  Please confirm.

PR #4959, opened 2017-04-14
FDS User Guide: add text re use of SPEC_ID with QUANTITY in DEVC

bmralph:
I get a lot of students asking how to output e.g. CO concentration or "model a CO detector". They were unclear how to output species-specific gas phase quantities using a DEVC. I've added an extra line of description to the device output section.


rmcdermo:
Thanks, Ben!

PR #4977, opened 2017-04-18
Update FDS+EVAC_Guide

sbenkorichi:
Parameters* 
I've changed some format of namelists written in plain text to \Timtt{ } as used in most of the guide.


tkorhon1:
I did update the FDS+Evac LaTeX source some time ago. The LaTeX source at Github was very old, it was for FDS version 5. And it has "5" in the file name also. The present file name is FDS+Evac_Guide.txt. I did add this to my file system (Git managed). Then I removed (using git commands) the old file (having "5") and the used git command "add" to add the new file to the git. When I did the "push" or something like that, Git said that one file was renamed. 

Before the push, I did by myself some of the svenkorichi changes (I used search in Emacs to find those words) to the LaTeX source. But I surely have missed some of the changes.

Now I have copied the latest 19.4.2017 git version to my own directory, where I update the manual.
So, now on, I will always get the latest git version (in my computer after git remote update), if I'm going to edit the manual. Previously, I was not doing this. Why? Well, nobody else was ever editing the FDS+Evac manual...

Wbr,
Timo

PR #5002, opened 2017-04-24
FDS Verification: start conversion to NEW_MOMENTUM_NUDGING

rmcdermo:


mcgratta:
Randy, don't burn the old nudging scheme just yet. Still need to play with it.

rmcdermo:
Actually, we can't.  It is still needed in the case where you want the forcing to act like a mean pressure drop.  I discovered this by, guess how, a verification case---the ribbed_channel_case has a variable duct height.  So, when you use local forcing you can't get the same effect as a mean pressure drop.

If you discover better properties with the old nudging scheme, we might want to think about applying the SPONGE_LAYER_DISTANCE to that method.  Currently SLD is applied only to the new scheme.  But it is independent of the nudging method and I think gives us better control over the boundaries.

mcgratta:
OK. I like the idea of the old scheme; that is, just superimposing a wind on top of the existing motion. The sponge layer might help.

PR #5023, opened 2017-04-30
Update FDS_User_Guide

sbenkorichi:
Removing room_fire.fds from the text. See \subsubsection{Modeling Upholstered Furnishings}
I wonder why this file was taken away from the files. I still remember it was in version 6.
If it was by mistake, then reject this request and the file should be added to the Fire directory.


mcgratta:
Thanks. The case is still in the repository, but we removed it from the verification suite because it really did not verify anything. It was just an example, but it takes a few hours to run.

sbenkorichi:
Ah OK thanks for the info.

PR #5042, opened 2017-05-04
FDS Source: Major restructuring of wall/boundary arrays.

mcgratta:


mcgratta:
These changes were done so that we can process boundary conditions for immersed (i.e. non-rectangular) solids the same way we do for "wall" cells and particles. `CFACE` is a derived type that is similar to `WALL`. I also moved some variables like `RHO_F` and `ZZ_F` into the `ONE_D` derived type. I also created new children of `ONE_D` like `TMP_G`, `RHO_G`, and `ZZ_G`. The idea is that we store the necessary gas cell info in `ONE_D` so that in the boundary routines we do not have to invoke, for example, `TMP(IIG,JJG,KKG)`. This was fine for wall cells and particles, but `CFACE's do not work like that.

PR #5069, opened 2017-05-10
Fix  the overview page

sbenkorichi:
I somehow was going through the Overview page coming with the bundle, I found an issue with the discussion group's link, when I looked into the script, I found the link of the manuals wasn't set up correctly which caused a distortion of the rest of the script.

### DIFF
--- a/Utilities/Scripts/for_bundle/Overview.html
+++ b/Utilities/Scripts/for_bundle/Overview.html
@@ -16,7 +16,7 @@ FDS generated data.
 <h3>FDS Documentation</h3>
 
 <ul>
-<li><a href="Guides_and_Release_Notes/FDS_User_Guide.pdf">User’s Guide for FDS</a>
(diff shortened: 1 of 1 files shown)


mcgratta:
Glenn, could you check this?

sbenkorichi:
Thanks Glenn, 
While verifying it, can you also correct the word "issue" in the last section "User Support"
"When reporting a problem to the **ISsue** Tracker"

gforney:
I changed ISsue to Issue.  I also checked all the links.  They look ok.  For future reference the Overview.html file is located in the installation at
C:\Program Files\firemodels\FDS6\Documentation

so if we see a problem we can edit the file there to get it right then copy it to the repo location at fds/Utilities/Scripts/for_bundle/Overview.html

sbenkorichi:
Brilliant Glenn. 
Yes, I was in the documentation directory under windows when I found out about it.

PR #5081, opened 2017-05-13
Update build_bundle webpage

sbenkorichi:
@gforney 
Glenn, can you take a look at this?

It looks to me that Update_repos_gitlatest.bat  has changed to UPDATErepos.bat. I've rechecked it on the smv repo and I couldn't find Update_repos_gitlatest.bat .

### DIFF
--- a/Utilities/build_bundle.html
+++ b/Utilities/build_bundle.html
@@ -30,7 +30,7 @@ function install_fdssmv(platform){
 <h3>Bundle Steps</h3>
 
 <ol>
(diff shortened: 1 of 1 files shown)


gforney:
thanks.  your are right, I changed the script name a while back but missed this link.  I'm going to fix the problem in a different way though by pointing to the smokeview build page where the same link/script exists.  That is where I usually update the repo anyway.  I'm not quite ready to put in my own pull request as I made some other changes which are not quite working yet.

sbenkorichi:
That's fine Glenn, you can fix it at a later time once you fix the other changes, this is a trivial thing, it can wait.

Cheers,

Salah

PR #5085, opened 2017-05-14
Remove blank pages

sbenkorichi:


sbenkorichi:
Oh snap, don't accept it plz. I forgot to remove the previous modification of the bundle webpage.

sbenkorichi:
I've fixed it back as it was.
Randy, How can I remove an old commit?

rmcdermo:
Salah,
The only way to remove an old commit is via reset, which is dangerous.  It is better to just use revert, or overwrite the old commit with a correction.  If you use reset, you lose everything after it; reset should only be used if you do something locally that you want to correct immediately and really understand what you are about to reset.

sbenkorichi:
Thanks for the info Randy, I will keep it in mind.

PR #5106, opened 2017-05-19
isojart1: Added comments and an under-construction IBM radiation rout…

ghost:
…ine. Some defined constants and variables have been described with more detail, and there's a subroutine "RADIATION_FVM2" that will contain a radiation solver for immersed boundaries (non-cartesian surfaces).


shostikk (approved):
I recommend combine the changes to prec.f90, and other files that are common prerequisites to all, to few cases, so that complete recompile may not always needed.

PR #5128, opened 2017-05-30
Matlab: clear path before adding scripts to path

BradHend:


rmcdermo:
If you are running Matlab on your local machine and have several different repositories (say local, burn, etc.) then Matlab can get confused as to which scripts directory to use and errors will occur.  This solves that problem.

PR #5153, opened 2017-06-20
Updated the makefile so that the intel_linux version compiles properly

ghost:
There was an error in the makefile and the intel_linux version didn't compile because the dependence of radi.f90 on geom.f90 wasn't mentioned. The makefile has been corrected in this version and this seemed to compile correctly in CSC Taito.


rmcdermo:
Thanks, Teemu!

PR #5188, opened 2017-07-01
Add CSVF entries to smokeview file for ctrl and mass.

JakeOShannessy:
Smokeview files contain entries for the location of hrr.csv and devc.csv files but not ctrl.csv or mass.csv. This adds these entries.

Smokeview ignores CSVF entries it doesn't recognise so this doesn't affect smokeview reading the .smv file.


rmcdermo:
Thanks, Jake.  @gforney Please review and merge if it looks ok.

gforney:
thanks, I'll take a look

PR #5191, opened 2017-07-03
Updated natconv cases to be evenly spaced in logspace by Rayleigh num…

BradHend:
…ber, changed naming convention to include Ra number in the name, replaced devices with line devices


rmcdermo:
Bradley,
This is looking good.  But a few things to modify.
1. I think the width needs to be doubled so that it is 4 times the channel width H.
2. Make line DEVC start at H and end at L-H with 30 points.
3. Add point DEVC to test steady state.  Put these just above the pressure device.  As an example:
```
&DEVC XYZ=0.0,0,1.6, IOR= 1, QUANTITY='CONVECTIVE HEAT FLUX', ID='Q1-0'/
&DEVC XYZ=0.2,0,1.6, IOR=-1, QUANTITY='CONVECTIVE HEAT FLUX', ID='Q2-0'/
&DEVC XYZ=0.1,0,1.6, QUANTITY='BACKGROUND PRESSURE', ID='p0' /
&DEVC XYZ=0.1,0,1.6, QUANTITY='DENSITY', ID='rho' /
```
4. Rename files so that Rayleigh number is exponential notation and has the same sig figs (say, 3) for each case so that the _8.fds line up.  This makes it easier to read when we have a list.  For example, 
```
natconv_Ra_2p28e00_8.fds
natconv_Ra_2p28e10_8.fds
```

rmcdermo:
Bradley,
Good job.  Note that I removed a couple of the files that looked like duplicates.  Update your repo and let me know if you see a problem.

PR #5202, opened 2017-07-05
FDS Verification: Created new natconv input file set, grid res. of H/16 and created new natconh input files with same Ra numbers as natconv

BradHend:


Verification/Heat_Transfer/natconv_10_16.fds
```diff
@@ -0,0 +1,83 @@
+&HEAD CHID='natconv_10_16', TITLE='LES of free convection in enclosed space, Ra = 2.28e09' /
+
+&MULT DY=4,DZ=4, J_UPPER=1,K_UPPER=7, ID='m1'/ one mesh in X, two in Y, eight in Z
+&MESH IJK=8,16,16, XB=0,2,-4,0,0,4, MULT_ID='m1'/
```
rmcdermo, line 4: Bradley,
It looks to me like these are still set at N=8 cells across the channel.  I think IJK needs to be 16,32,32

rmcdermo (commented):
I accepted this for now, but Bradley is going to fix the _16 cases.

PR #5210, opened 2017-07-07
FDS Verification: updated T_END values for natconh, added NOISE_VEL. to laminar cases

BradHend:


rmcdermo:
Bradley,
I'm surprised the T_END values are lower; also the noise velocity is lower at higher Ra, which I don't understand.  Try computing the time scales and velocities using the heat transfer velocity from

v_ht = h/(rho*cp)

where h = Nu*k/H and H is the cavity height.  Compute rho from the ideal gas law and the initial temperature of the cavity (average of the two wall temps).

T_END = H/v_ht

NOISE_VELOCITY = 0.05 * v_ht.

PR #5240, opened 2017-07-17
Clean up Cardboards

sbenkorichi:
Remove ABSORPTION_COEFFICIENT.


rmcdermo:
Why is this change being made?

@mcgratta or @drjfloyd please review.

sbenkorichi:
Randy,
Kevin requested it. He will review it and merge it.

Cheers
Salah

mcgratta:
Stas Stoliarov provided material properties for the double wall cardboard used at FM. His properties included very large value of ABSORPTION_COEFFICIENT, which means that all radiation is absorbed at the surface. We do not need these values in the input file, as FDS defaults to this behavior.

PR #5250, opened 2017-07-19
Update character limit text for TITLE and CHID.

bwklein:
TITLE is defined with a 256 character limit.
https://github.com/firemodels/fds/blob/master/Source/cons.f90#L100

In cons.f90 CHID is defined with a 40 character limit. 
https://github.com/firemodels/fds/blob/master/Source/cons.f90#L101

This update changes the text in the User Manual to match the source.


rmcdermo:
Thanks, Bryan!

PR #5270, opened 2017-07-21
FDS Validation: Waterloo Methanol Updates (not .125)

drjfloyd:
Added 2 cm case
Added pan with 1 cm lip
Moved burner up off floor (in test it was in mid air)
Extend domain (prior inputs had the plume close to the OPEN sides)
For predicted case increased conductivity from quiescent value of 0.2 to 9 (estimate based on predicted heat feedback and thickness).


rmcdermo:
Jason,
Don't forget to merge, unless you want one of us to.

drjfloyd:
Thanks Randy.

PR #5313, opened 2017-08-05
New verification case for Radiation chapter

tmilac:
Compares the heat flux received along two axes by a small target at various distances from a gas-fired panel, as computed by FDS and configuration factors.  The results are also compared with data from an old paper by Simms.


mcgratta:
Tom, I'm off to Germany for a week, so if there are any minor snafus with this commit, work with Jason Floyd and Glenn to sort them out. Sometimes it's just a matter of committing a missing file or fixing a link within the guides. Thanks.

PR #5330, opened 2017-08-17
FDS Source: Move wind related to parameters to WIND namelist group

mcgratta:


mcgratta:
This commit only moves parameters from the MISC to WIND lines. It does not change functionality. V&V input files will be committed in a separate commit shortly.

PR #5335, opened 2017-08-17
Use MPI_Gather instead of MPI_Send/MPI_Recv

sayantansur:
Hello FDS team!

I have been looking at FDS and its network related performance at Intel. I noticed that due to a pattern of communication in EXCHANGE_DIAGNOSTICS, there can be a large queue of messages piled up on rank 0. In this patch, I have tried to solve the problem by using a MPI_Gather operation instead. As you might be aware, the collective operation usage allows the MPI library to optimize the message passing pattern depending on the platform.

While I have been part of the MPI and HPC community for some time now, I am *not* an expert in Fortran :) Therefore, please review the patch for correctness. I ran some of the workloads on a 4 node setup with Intel compiler, OpenMP and Intel MPI on Knights Landing w/ Intel Omni-Path fabric. It seemed to work for my test case.

Please let me know any feedback you might have on this patch!

Thanks,
Sayantan.

===
In the current scheme every rank sends 23 small messages
to rank 0. Rank 0 specifically picks out messages from rank
1, and then rank 2 and so on. When there are sufficient ranks,
there can be a lot of messages that have arrived and piled up
in internal MPI queues.

The solution is to call MPI_Gather instead of individual p2p
messages, such that the MPI library can internally use scalable
communication patterns to reduce the message pileup.

Pack the data into a structure type and manually unpack on rank 0.

Signed-off-by: Sayantan Sur <sayantan.sur@intel.com>


mcgratta:
Wow, thanks. Let me just test it and if all goes well I'll merge.

mcgratta:
I'm running a 4 process case.  Two issues

```
../../Source/main.f90(2891): warning #7416: Fortran 2008 does not allow this intrinsic procedure.   [SIZEOF]
STRUCT_DISP(2) = SIZEOF(1.d0) * 7
```
I think you need `STORAGE_SIZE` which returns the size in bits, not bytes.

Second issue, when I run the case using Open MPI, I get:
```
forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable MESH_RECV when it is not allocated

Image              PC                Routine            Line        Source
fds_mpi_intel_lin  0000000008F7ACD6  Unknown               Unknown  Unknown
fds_mpi_intel_lin  0000000003678A8E  fds_IP_exchange_d        2932  main.f90
fds_mpi_intel_lin  0000000003567C6D  MAIN__                    815  main.f90
fds_mpi_intel_lin  000000000054D71E  Unknown               Unknown  Unknown
libc-2.12.so       000000333501ED5D  __libc_start_main     Unknown  Unknown
fds_mpi_intel_lin  000000000054D629  Unknown               Unknown  Unknown
```
Seems odd that all ranks would need `MESH_RECV` to be allocated. 

I'll try the case wtih Intel MPI and see if that is different.

mcgratta:
Same problem with
```
 MPI version: 3.1
 MPI library version: Intel(R) MPI Library 2017 Update 3 for Linux* OS
```

sayantansur:
Thanks for the comments, I'll check. I thought I had run with Intel MPI 2017.1.132. In any case, I'll also run with OpenMPI to verify.

sayantansur:
I updated for the storage_size usage. I'm not sure why MESH_RECV is being accessed by a non-root rank, the MPI library should really be ignoring the recvbuf argument at the non-root ranks. I double checked with Intel MPI 2017 Update 3, and the wrf_prof_ramp.fds workload. 4 ranks with 16 OpenMP threads each.

Can you tell me the workload and configuration that failed for you? Maybe there's something specific in that?

mcgratta:
I can get the job to run by allocating MESH_RECV on all ranks. But the results are not right, as there are several bugs in the way the numbers are stored and unpacked. The thing to do is to run the case with the original code and then with the changed lines, and diff the .out files. They should be identical. The values being passed in this routine are diagnostic values for each mesh.

Also, we allow one to run with multiple meshes assigned to one MPI rank. This would be a problem here because MESH_SEND would get overwritten with the values from multiple meshes.

sayantansur:
I see, I'll check the packing/unpacking aspects and get back to you.

sayantansur:
question - can the number of meshes at each rank be unequal? i.e. one rank as 2 meshes whereas all the others have 1?

If that is true, we cannot use Gather, and will have to use either:
a) Gatherv w/ exchange of counts before, complicating the logic further
OR
b) pack/unpack the 23 sends in the struct but use a single MPI_**S**send() instead. The ssend keeps the senders in check and doesn't overflow the single receiver.

Any thoughts? Otherwise, I might implement b) since it is easier.

sayantansur:
I submitted the ssend patch in #5342

PR #5342, opened 2017-08-22
Use MPI_Ssend+pack data in EXCHANGE_DIAGNOSTICS

sayantansur:
In the current scheme every rank sends 23 small messages to rank 0.

In this patch, they are combined into one message that is then
unpacked at rank 0. Additionally, the current patch uses MPI_Ssend
that blocks the senders until the receiver has arrived. This prevents
many senders from overwhelming the receiver across multiple iterations.

Signed-off-by: Sayantan Sur <sayantan.sur@intel.com>


sayantansur:
I ran this with Intel MPI 2017 Update 3 and the WRF input files. I diffed the *.out files and verified that the contents were similar.

mcgratta:
Thanks. Yes, this must work when more than one mesh is assigned to an MPI process. I think the way you are doing things now is fairly safe. One additional thing to do is to make the sends and receives persistent. I'll add that.

sayantansur:
Thanks for merging!

PR #5357, opened 2017-08-28
FDS Verification: Fix thread check scripts

mcgratta:


mcgratta:
Glenn -- note that I fixed a minor typo in the inspect_fds script, and switched to using Intel MPI instead of Open MPI until I can figure out why the Open MPI is giving us an error.

PR #5366, opened 2017-08-30
New validation cases

ghost:
Deepak Paudel had modified some validation cases and we were not able to do the pull request from his account (it gave a mysterious 403 error and didn't accept password). So I put the modified files to my own copy of fds and now I'm submitting them.


rmcdermo:
Teemu,
There is a conflict that must be resolved first.  Do you see this?

dpkPrasadPaudel:
Hi Randy,

I have modified this file: Added matlab plot lines for the experiment NFSC2, (Added output quantities are HGL Temepratures, Plume Temperatures, Adiabatic Surface Temperature and Wall Temperatures" 

I do not have write access to this file,  what can we do about it ? should I use the original file ?

-Deepak

sbenkorichi:
Did you update your repo before you make this modification? I think this is due to that. You might think of updating it and then making another pull request.

ghost:
Ok I'll do that immediately.

Teemu
Lähettäjä: Salah Benkorichi <notifications@github.com>
Lähetetty: 30. elokuuta 2017 15:48:32
Vastaanottaja: firemodels/fds
Kopio: Isojärvi Teemu; Author
Aihe: Re: [firemodels/fds] New validation cases (#5366)


Did you update your repo before you make this modification? I think this is due to that. You might think of updating it and then making another pull request.

-

ghost:
Removed conflicts from the previous one.

sbenkorichi:
There you go.

ghost:
Thanks for noting the cause of problem, guys.


-Teemu

Lähettäjä: Salah Benkorichi <notifications@github.com>
Lähetetty: 30. elokuuta 2017 16:19:40
Vastaanottaja: firemodels/fds
Kopio: Isojärvi Teemu; Author
Aihe: Re: [firemodels/fds] New validation cases (#5366)


There you go.

-

rmcdermo:
Teemo,

There is a lot that needs to be addressed in this change before it can be accepted.

For example, now I see a change to type.f90.  Did you intend this?  It is better for source code to be sent in an independent pull request, at least until we are confident that there are not problems.

I am going to close this so that no on inadvertently merges it.  In the mean time, please work with Kevin and Simo to get things clean before sending the PR.  Thanks.

ghost:
Sure, I'll check what's wrong with the type.f90 and do a new pull request tomorrow.


-Teemu

Lähettäjä: Randy McDermott <notifications@github.com>
Lähetetty: 30. elokuuta 2017 16:22:55
Vastaanottaja: firemodels/fds
Kopio: Isojärvi Teemu; Author
Aihe: Re: [firemodels/fds] New validation cases (#5366)


Closed #5366<https://github.com/firemodels/fds/pull/5366>.

-

mcgratta:
Yes, in addition to making the source code change a separate commit (which makes it easier for us to identify when and where bugs might be introduced into the code), here are some other things you need to do:

1. These cases should be put in the `VTT` folder of the fds repo. The 3 current VTT cases were part of our original validation set, and that is why we did not choose a particularly good name for the folder. We like to identify the lab or university or individual that did the experiments in the name, plus something else to identify what exactly was done. For example, `VTT_NFSC2` might be what we call all of these experiments, assuming all were done under this name. `NFSC2` is OK for naming the input files, but put them in the VTT folder for now, and at some later date I will go and change the name from `VTT` to something more useful. 

2. Change the `FDS_validation_dataplot_inputs.csv` back into unix format. To do this, open the file on a linux machine with vi, and :set ff=unix. When you do a pull request, you should check to see that the GitHub browser shows us only the changes you have made. Currently, the entire file shows up as changed because of the line endings. Currently, we cannot see the changes in the browser -- the file is too big.

3. Try not to change the `FDS_validation_scatterplot_inputs.csv` file. You added `Wall Temperature` even though we use `Surface Temperature` instead. If you add `Wall Temperature`, then we will have separate accuracy statistics for your experiments. We do not want this. 

4. Input files -- `TITLE='Test'` is not useful. It might be better to spell out what NFSC2 means, or something more illustrative of the experiment. This just helps someone who is browsing the cases.

5. Input files -- Ramps -- they do not need to be so long. Long ramps make it difficult to work with the file. Other contributors might see this and think that such long ramps are necessary. Try to use 10 ramp points or less, unless the complexity of the ramp function is truly needed.

6. Wherever possible, make the FDS output, experimental files, and FDS input files as short as possible to limit the size of the repos, but also to make the processing as fast as possible.

Finally, before a new pull request, send us copies of the sections of the FDS Validation Guide that you have completely compiled in your cloned repo. There might be minor style comments that we would have.

Sorry for all this, but it is becoming increasingly difficult to manage the validation cases and guide. Over the years since we first started, we have developed rules to maintain consistency. Some, but not all, of this guidance is found [here](https://github.com/firemodels/fds/wiki/Validation-Case-Setup-Example).

shostikk:
Thanks Kevin for taking time to explain all this (that I should have done).

One comment, the Wall temperatures in these tests were measured inside a wall, at different depths. That's why Deepak decided to create a new quantity. I realize we should have discussed this before submitting a request. Anyway, we think 'Surface temperature' is not really what was measured. The data can of course be included in the same set, though.

mcgratta:
Yes, but I have other inside wall temperature data under `Surface Temperature`. This is a bit confusing because we also have `Target Temperature`, which includes other solids that heat up. We keep this because of our work with the US NRC, who distinguish walls and targets. In any case, just add the wall temperature data to `Surface Temperature` so that the statistics can be compiled.

PR #5385, opened 2017-09-03
FDS Build: Change names of various folders for consistency

mcgratta:


mcgratta:
Note that now we refer to the Windows FDS build folders as `impi_intel_win_64`

PR #5388, opened 2017-09-04
FDS Build : Remove instructions for TAU instrumented build on makefile.

marcosvanella:
Hi Glenn, please check that the win targets compile properly now.
Thank you.


gforney:
teststec on both burn and my pc.  fds builds fine now. thanks.

PR #5389, opened 2017-09-04
Update box_burn_away_2D_residue.fds

sbenkorichi:
Correct the mass of the box.  Even though the box is set to 1 m, however the domain is only 0.4 m, where .6 m will be ignored. Therefore, the actual mass that will be compared is within the domain, i.e. 0.4. Hence the total mass is 1.28 kg and 0.64kg for residue as found in the Devc file, or by integrating the MLR area.


sbenkorichi:
Please review first. Thanks !

Issue #5411, opened 2017-09-06
2 step extinction model error


rmcdermo:
Jason,

I'm happy to work this, but wanted to give you a heads up in case you think of something.  Something has happened to the 2 step CO extinction scheme---it no longer matches the 1 step reaction results for UMD line burner.  See below.  I actually think this may have happened before MaCFP, but I did not show the coarse cases there.  In the plots, focus on the 1.25 cm results (the finer cases have not been rerun).

My first step will be to bisect the code and see if I can figure out where things when off.  If you have any thoughts, let me know.

[methane_eta_new.pdf](https://github.com/firemodels/fds/files/1282704/methane_eta_new.pdf)
[methane_eta_old.pdf](https://github.com/firemodels/fds/files/1282703/methane_eta_old.pdf)
[propane_eta_new.pdf](https://github.com/firemodels/fds/files/1282702/propane_eta_new.pdf)
[propane_eta_old.pdf](https://github.com/firemodels/fds/files/1282705/propane_eta_old.pdf)

drjfloyd:
The old cases were 4/25 source.  If it was something in fire.f90 my guess is https://github.com/firemodels/fds/commit/704a025e2ba223e5103ae5cff2d8641ee9784242#diff-f44fc1cac74f2b00107607131ecbb726 or https://github.com/firemodels/fds/commit/5662001e4c237fc8b6d5cd42d086ccaa625c7c20#diff-f44fc1cac74f2b00107607131ecbb726

rmcdermo:
Your guess was correct:

```
704a025e2ba223e5103ae5cff2d8641ee9784242 is the first bad commit
commit 704a025e2ba223e5103ae5cff2d8641ee9784242
Author: dr_jfloyd <jfloyd@jensenhughes.com>
Date:   Fri May 5 13:09:41 2017 -0400

    FDS Source: Fix AIT check for Extinction 5
```

drjfloyd:
Looking at those changes again, I now think they were not correct.  Try the source I just pushed up and see if it is working again.

drjfloyd:
Why is the 2nd (CO) step AIT set to 20 to in the UMD cases? In the testing I had done back in April I was using 610.  Any CO made at the burner surface in the low AIT INIT zone would not have any AIT restriction further up.

rmcdermo:
Ugh!  Good catch.  I don't know.  I must have overwritten the input files.

rmcdermo:
[...]

PR #5442, opened 2017-09-12
FDS Source:  Undo 704a025, fix issue #5411

drjfloyd:


rmcdermo:
Unfortunately, this did not fix things.  I am running the attached test case.  It is a shorted version of the coarse UMD line burner, 2 step.  It only goes from 40 s to 45 s.  Of course, depending on which version of the code, you have to be careful about which extinction model you apply (3 or 5).

[methane_2step_XO2_ramp_dx_1p25cm.fds.txt](https://github.com/firemodels/fds/files/1297170/methane_2step_XO2_ramp_dx_1p25cm.fds.txt)

Here is the old working version result:

methane_2step_xo2_ramp_dx_1p25cm_bd87dc6

And here is the result of the code from this Pull Request:

methane_2step_xo2_ramp_dx_1p25cm_9be34e5

rmcdermo:
I'm going to close this for now.

PR #5445, opened 2017-09-13
FDS Source: remove CPU Time from .err

rmcdermo:


rmcdermo:
@mcgratta If you want to add this under an input flag on DUMP, that's fine.  But I cannot stare that this by default.  I suggest you simply add it to the _cpu.csv file or the .out file.

PR #5455, opened 2017-09-14
New verification case

sbenkorichi:
This is a clean PR with the latest update.


sbenkorichi:
@rmcdermo 
Randy, take a look.

rmcdermo:
Salah,
This new PR looks nice!  Last thing.  On your units fix for q''', I think the factor should be to divide through by (rho cp) or, in terms of the properties you have defined, multiply by alpha/k.

rmcdermo:
Also, are all citations working? If you sent me the pdf I have not checked it because I'm still at home.

sbenkorichi:
Yes, the citations are fine. Check the file I sent you. I'm not sure about factor though,

rmcdermo:
I am :).  First write down the conservation of energy equation, which has units of kW/m3, and then assume constant rho and cp and divide through by it.  Or, look in the FDS source code to see what the heat source terms looks like.

sbenkorichi:
Ok

sbenkorichi:
I think you're right, I forgot to multiply it by \alpha. Now, I moved it into the brackets. 
\frac{\partial T}{\partial t} = \alpha \left[ \frac{1}{r^2} \frac{\partial}{\partial r^2} \left( r^2 \frac{\partial T}{\partial r} \right) + \frac{\dot{q}^{\prime\prime\prime}}{k} \right]

As you said before, the Q is in W and then gets multiplied by 1000 before it goes to the init.

rmcdermo:
OK, you've suffered enough.  I need to add the capability to carve a sphere out of a block of OBST, but this is not your problem.  Thanks for all the effort.

sbenkorichi:
Haha it's part of learning. No pain no gain lol. Thanks anyway. I'm also thinking to follow Kevin's suggestion to write a separate script to generate these type of from.

rmcdermo:
Salah,
Once you have caught up on sleep, please have a look at the latest User Guide and double check that all the inputs on OBST, INIT lines, etc., are properly described for HT3D.  Thanks!

Issue #5408, opened 2017-09-06
Problem with PRISME HVAC


mcgratta:
I just ran the PRISME validation cases and noticed strange behavior with the HVAC that affects all the results. This image sums it up. Has there been a major change to the HVAC stuff in the last few months?
[PRS_D3_Room_1_Pressure.pdf](https://github.com/firemodels/fds/files/1282563/PRS_D3_Room_1_Pressure.pdf)

mcgratta:
The NIST_NRC closed door cases are also affected by this problem. Something related to leakage?

drjfloyd:
I haven't touched HVAC since April.  The last changes to the HVAC routines were made by you in May when you added CFACE. How long has it been since the PRISME cases were run?

mcgratta:
Dec 8, 2016

drjfloyd:
The April change kept FDS from trying to solve for flow in a particular leakage path if there were no wall cells for leaking (e.g. someone defined LEAK_AREA on ZONE but no active surfaces with a corresponding LEAK_PATH).  I can check to see if change introduced some other issue.

mcgratta:
OK, I vaguely remember that the original leakage functionality would assume the entire zone leaked to ambient, and these input files might have been designed based on that assumption.

drjfloyd:
The original leakage model always required that some SURF have a LEAK_PATH for FDS to define what walls cells to apply a leakage velocity to.  Prior to using HVAC for leakage, if no LEAK_PATH was defined, then the velocity was not applied to anything.  With leakage computed by HVAC the ZONE leak area is used to define a duct and two duct node; however, if there are no wall cells to assign to the node then the HVAC model needs to be told to not solve for that duct.  NIST_NRC_01 (closed door) has a LEAK_PATH defined for the marinite surfaces, so the leak area change may not be the problem.

mcgratta:
This short input file appears to give the correct peak pressure of about 900 Pa. Not sure why this works and the actual cases do not:
```
&HEAD CHID='NIST_NRC_test' /

&MESH IJK=50,32,16, XB=0.0,21.7,0.0,7.04,0.0,3.82 /
&TRNX IDERIV=0,CC=10.8,PC=10.8 /
&TRNX IDERIV=1,CC=10.8,PC=0.5  /

&TIME T_END=300. /

&ZONE LEAK_AREA(0)=0.058 /

&REAC FUEL='METHANE' /
[...]

PR #5456, opened 2017-09-14
FDS Source: Issue #5408. Fix bug related to leakage through default b…

mcgratta:
…oundary


mcgratta:
Randy, if you have a second, could you double check this. I ran the verification cases and nothing fell out of tolerance, but then again, nothing had before. I'm going to rerun NIST_NRC and PRISME.

rmcdermo:
Looks pretty clean to me.  Only thing that could be wrong would the sign of something, which I expect verification would easily catch.

PR #5462, opened 2017-09-14
FDS Verification: cleanup ht3d_sphere

rmcdermo:


sbenkorichi:
Two things here:
1. ssf script inst able to load sf3 and sf5. Thus rendering a blank image.
2. Even after uploading them manually, there is sth strange with this multi-mesh. As shown in the attached image.

ht3d_sphere_75_0859

rmcdermo:
Right.  I'm working on it.  But I had to do something, because the single mesh case was taking way too long.  Hopefully have this working before firebot tonight.

sbenkorichi:
Sure, I can fix the 1st issue maybe by re-uploading a new ini and ssf files.

rmcdermo:
The last commit should have fixed everything.

sbenkorichi:
Yea, that's better since it had multiple slices.
Thanks.
Let's see by tomorrow if the bot likes it.

gforney:
if you draw the slices as opaque they should look ok. the z-buffer takes care of obscuring portions of one slice that are behind another.  The correct way to draw semi-transparent objects is to sort all triangles in both slices and then draw them back to front.  this would be hard to do with slice files.  you would have to draw some of one slice then some of another then some more of the first slice etc.

PR #5471, opened 2017-09-15
pull request to test updates to support modules

gforney:
DO NOT ACCEPT THIS PULL REQUEST - this pull request is for testing only

this pull request makes changes to various linux fds make_fds.sh build scripts to support modules.  The changes are also designed to work if you do not have modules ie make_fds.sh scripts should work as before if you do not have modules.  Typical work flow

0.  put in appropriate module load commands in startup file so that the environment is set up to either run the installed FDS or to build and run fds from the github repo

1. switch from installed fds to repo fds
module unload FDS6
module load openmpi/64ib

2.  switch from repo fds to installed fds
module unload openmpi/64ib
module load FDS6


gforney:
I added a wiki on  modules at
https://github.com/firemodels/fds/wiki/Modules

PR #5510, opened 2017-09-26
mult_mesh generator for windows

sbenkorichi:
Add new app for windows users.


sbenkorichi:
@rmcdermo 
Randy, can you check this?
I've added an app for windows users, they can simply open it by double click and use it directly.
Let me know if you have comments on this.

rmcdermo:
Do not send PR for exe.

rmcdermo:
@sbenkorichi we do not put exe under version control.  If you want to develop your own mesh generator, then work on it offline and then we can point to it as a third party tool or something.

sbenkorichi:
Hmm  I see. Ok sounds good. 
I will let you know when I get a link for it.

sbenkorichi:
Take a look 
https://github.com/sbenkorichi/mult_mesh_generator

rmcdermo:
That's cool.  Let's see where it goes.

But I think you misunderstand my point about committing the exe to the repo.  It is not a good idea in general to put binary files under version control.  That is what you want to use your "releases" page for.  So, tag your repo (v1.0 or whatever) and remove the .exe and point to it from the README.md file.

rmcdermo:
You should, by the way, do a git reset to purge your repo of the exe file.  But at this early stage, you may even want to delete the repo and start over.  The exe history will be there on GitHub and you would have to use BFG to get rid of it.  It's fine, of course, if you don't care, but I'm just trying to point you toward good repo practices.

sbenkorichi:
Thanks man. 
Yes I just went and removed, reform it again and it's even with the firemodel master branch.

sbenkorichi:
I will try your above suggestion.

sbenkorichi:
I think I have a better idea, I will just create an online webpage GUI for it, so one will only use a webpage instead of downloading and stuff.

rmcdermo:
That would be great.

sbenkorichi:
Here http://www.sbenkorichi.com/handy-tools/mult-mesh-generator/
Please try it on your side to double check.

rmcdermo:
Very nice!  There is still some work to do in the script.  My original had some flaws that need to be fixed.  Check out the following example.  Notice that the Y domain is off.  It should start at 0 and go to 0.33.  Also, I think you want to output a MULT line.

screen shot 2017-09-27 at 3 31 36 pm

sbenkorichi:
Thanks ! I didn't know that.
The MULT line can be added no problem, but let's fix that flaw first.

rmcdermo:
The logic needs to be a little different depending on whether you want M0 to be in the center or if it is on the edge.  In the case I was working with it was in the center.

sbenkorichi:
I see your point, 
I will try to see if I can work on it tomorrow.

bwklein:
What is the use case for this?

rmcdermo:
@bwklein I developed the original script to make the MULT lines for the core meshes surrounding the burner in the most recent McCaffrey Plume input files.  [Here](https://github.com/firemodels/fds/blob/master/Validation/McCaffrey_Plume/FDS_Input_Files/McCaffrey_14_kW_45.fds) is an example.

In this case, I wanted to make certain that the mesh over the top of the burner had an odd number of cells.  I typically design this by first asking what I want a single mesh case to look like.  But it is not a trivial thing to then get a sensible MULT configuration.  What I found is that I have to start with the mesh block I care most about (M0 in the script) and extrude from there.  This is why you end up with nasty looking MESH lines in the example.  But it all started from a very reasonable single mesh.  It is now quite typical, however, that when we do validation we use a single mesh for our coarsest case, about 16 meshes for our mid-level grid, and 100 to 200 for our highest resolution.

rmcdermo:
Oh, why an odd number of cells?  Because I had a line of DEVC for centerline TMP and I did not want to be off by DX/2.

sbenkorichi:
This should fix the mult line,
Still need to fix non central cases. 
multline

gforney:
salah, what development system did you use to develop your app, ie C, Perl etc

sbenkorichi:
Python

rmcdermo:
OK, looking good.  But the MBLKS and the I_UPPER, etc. are over-specifying the problem.  Notice that the MULT line is not consistent with your MBLKS.  Once you have M0 and MBLKS, you do not need to enter I_LOWER, I_UPPER, etc.  Assuming M0 is the MESH line you are going to write out, I_LOWER = M0[1] - MBLKS[1], I_UPPER = I_LOWER + MBLKS[1] - 1 = M0[1] - 1.

Plus, to avoid clutter, if, for example, MBLKS[1] = 1, no need to put DX, I_LOWER, I_UPPER on the MULT line.

sbenkorichi:
This should fix the first flaw, I'm starting form a simple approach, later we can add more complex stuff.
See if you can break it now,

test

I've tested it on the sphere_75 and it's giving same results. See the comparison:

sphere
```
&MULT ID='m1', DX=0.08333333333333333,DY=0.08333333333333333,DZ=0.08333333333333333, I_UPPER=2,J_UPPER=2,K_UPPER=2/
&MESH IJK=25,25,25, XB=-0.125,-0.04166666666666667,-0.125,-0.04166666666666667,-0.125,-0.04166666666666667, MULT_ID='m1'/ 27 mesh
```

PR #5530, opened 2017-09-29
Source and V&V : Added SOLVER variable in &PRES line.

marcosvanella:
This pull request contains:
- makefile : Glenns fix for Testing cluster tools for MKL lib in Windows OS. 
- makefile : Add link line for Intel MKL library on Windows targets + /no-wrap-margin compilation flag on these.
- Source and V&V : Added SOLVER variable in &PRES line.


rmcdermo:
@marcosvanella looks good.  How about going back to PRES_ON_WHOLE_DOMAIN=F by default?

marcosvanella:
All right, should we name this flag something else? How about OBST_VEL_ERROR?

rmcdermo:
Since we are going to make it F by default, generally users won't need to set it, so let's leave it alone for now.

PR #5539, opened 2017-10-02
FDS Source : Set default of PRES_ON_WHOLE_DOMAIN=.TRUE. when using FF…

marcosvanella:
…T os SCARC pressure solvers.

### DIFF
--- a/Source/read.f90
+++ b/Source/read.f90
@@ -7352,11 +7352,16 @@ SELECT CASE(TRIM(SOLVER))
       ITERATE_PRESSURE = .FALSE.
       IF (SCARC_METHOD == 'null') SCARC_METHOD = 'KRYLOV' ! Taken as default for SCARC when SOLVER is SCARC and
                                                           ! SCARC_METHOD is not defined.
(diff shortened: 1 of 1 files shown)


rmcdermo:
Not to beat a dead horse here... but I did not realize that PRES_ON_WHOLE_DOMAIN could touch the default algorithm.  I suggest setting it back to T by default and then setting it to F within your SELECT CASE for GLMAT.  Thanks.

mcgratta:
I suggest making sure that it doesn't affect the "FFT" solver, regardless of its value.
```
IF (SOLVER='GLMAT' .AND. PRES_ON_WHOLE_DOMAIN) ...
```

marcosvanella:
Yes, the culprit is in the routine NO_FLUX in velo.f90. I'm taking the factor that makes the gradients zero in SOLID_BOUNDARY wall cells out, and changing the conditional to what Kevin suggests. 
I'll also do a quick grep to make sure it's not interfering with the default settings anywhere else.

PR #5543, opened 2017-10-03
 FDS Source and Validation : Replace PRES_ON_WHOLE_DOMAIN &PRES line variable for EXACT_VELOCITY_BC, only used when SOLVER='GLMAT'.

marcosvanella:


marcosvanella:
Randy don't merge it yet. I figured there's still another change to do.

rmcdermo:
My only comment is, why not just let heated channel flow cases use default GLMAT?

marcosvanella:
We have the "trip wire" obstacles disappearing in these cases before we start taking stats. At this point that can be only done having the solver work on the whole domain. Using the default EXACT_VELOCITY_BC=.TRUE. would be feasible if we rebuild the new matrix and reinitialize the solver for the no obstacles problem. This is not done now and explosions ensue. Maybe something to do for the next release?
I think we need to take out the SHUTDOWN command on the default solver CASE statement on READ_PRES.
It gives the shutdown when the EXACT_VELOCITY_BC variable is not being defined in FFT solver cases. That is practically all V&V cases.. ouch!
When SOLVER=FFT, it doesn't matter what you set this variable to on the pres line.

PR #5552, opened 2017-10-05
FDS Utilities: Do not specify mapping option for OpenMPI

mcgratta:


mcgratta:
This should enable faster run times for Open MPI.

PR #5607, opened 2017-10-23
FDS Verification Guide: Replace \clearpage with \FloatBarrier

mcgratta:


mcgratta:
I discovered a useful LaTeX command. `\FloatBarrier` instead of `\clearpage` will force all floats up to that point to be printed, but will not force the text to the next page. This prevents having a single plot abandoned on a single page.

PR #5608, opened 2017-10-23
FDS Validation Guide: Replace \clearpage with \FloatBarrier

mcgratta:


bmralph:
If you wanted to omit all the `\FloatBarrier` lines and you want this functionality only prior to new `\section{}`s you can just include the `[section]` option in the `placeins` package. i.e.

`\usepackage[section]{placeins}`

Which will change the definition of `\section` to include a `\FloatBarrier`.

rmcdermo:
Thanks, Ben.  That sounds like a good idea.  I know my additions of \newpage, etc., have gotten pretty random in editing the documents.  It would be nice to have one line upfront to control this.

bmralph:
I can send a pull request later today if it helps.

rmcdermo:
Let's get a thumbs up from @mcgratta before you do the work.

mcgratta:
Randy and I are checking to docs. We'll do as we check. There might be places where we still want to use \clearpage or \newpage, or we want clear floats at subsections.

PR #5617, opened 2017-10-24
remove tmi fabric from INTEL mpi portion of qfds.sh

gforney:
also update scripts used by bots to use Intel MPI version of fds


mcgratta:
Are you waiting to merge this?

PR #5630, opened 2017-10-26
 makefile : Add Intel MKL link lines for GNU targets. 

marcosvanella:
I'm currently writing a wiki on how to install gfortran, OpenMPI and Intel MKL (downloadable for free from the Intel site) on a Linux machine, and compile FDS with these.


sbenkorichi:
I've seen the wiki page, it looks cool. Here is some notes with how I could get gnu 7 working.

**Gnu installation:**

```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt update
sudo apt install gcc-7 g++7 gfortran-7
```
Check gcc, g++ and gfortran version:
There is a big chance they will be set tot he default, i.e. how they were installed first, maybe 4 or 5 or 6, if you're installing ubuntu 17, then you would get by default 7, then no need to install them. 
Ok, to fix this, do:
`sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 60 --slave /usr/bin/g++ g++ /usr/bin/g++-7 --slave /usr/bin/gfortran gfortran /usr/bin/gfortran-7` 

All should be good. If you get complains about update-alternatives and you see you don't get them updated to 7, then do this fix: 
`$ yes '' | sudo update-alternatives --force --all`
Now, this should fix it, once and for all. Check : gfortran --version 
You should get it set to 7.2

**2. OpenMPI installation:**

Download and extract the latest version of Open MPI.
cd to the directory that you extracted Open MPI to (supposedly it’s in the /Downloads directory).
4.Install Open MPI using the gnu compiler; configure and install using:
But, before that, you should created this directory first:
```
$ sudo mkdir /shared/
$ sudo mkdir /shared/openmpi_64/
$ sudo chmod -R 777 /shared/openmpi_64/
```

Then configuration:

`$ ./configure CC=gcc --enable-mpi-fortran --enable-static --disable-shared  --prefix=/shared/openmpi_64`
then type
```
$ make
$ sudo make install
```

export some paths to your .bashrc file in your home directory:
```
#FDS -----------------------------------
export MPIDIST_ETH=/shared/openmpi_64
export PATH=$MPIDIST_ETH/bin:$PATH
export LD_LIBRARY_PATH=$MPIDIST_ETH/lib:$LD_LIBRARY_PATH
export MPIDIST_IB=/shared/openmpi_64ib
#FDS -----------------------------------
```

log off and log in again.
Check if mpifort is in the right directory, it should be in:
```
$ which mpifort
/shared/openmpi_64/bin/mpifort
```

Next, Install MKL libraries...

marcosvanella:
Thank you Salah, I'm still working on this wiki. I practically did what you did for gcc-7 and openmpi, except I did not update the links for gcc. We can leave that as an option.
Btw I'm using Ubuntu 14, so to configure openmpi it was sufficient to define FC=gfortran-7 CC=gcc-7.

sbenkorichi:
See there you go, that's another way how to do it.

PR #5672, opened 2017-11-03
Possible change to DEVC STATISTICS

shostikk:
Can anyone figure out possible adverse consequences of this?

This would include all HT3D obstructions in volume statistics. We might need a switch for this to toggle.


rmcdermo:
@shostikk I do not see a problem with this.  But I do not have time this second to review.  Do you have a test case we can check?

PR #5678, opened 2017-11-06
FDS_User_Guide: Update Makefile directory

sbenkorichi:


sbenkorichi:
I would like also this statement to be reviewed :  

"To compile FDS, you need a Fortran compiler (2008 or greater) and MPI libraries. More details on MPI
can be found at the FDS repository https://github.com/firemodels/fds."

To be found where? Lots of wiki pages, can we direct the user to one specific direction? Thanks !

rmcdermo:
@sbenkorichi I agree with you that the statement is vague.  But the problem with wikis is that they change frequently.  I'd rather the user guide remain general so that we can have the flexibility to change the wikis without breaking links.

sbenkorichi:
Ok

Issue #5683, opened 2017-11-08
Color DIMGRAY isn't defined in the read file


sbenkorichi:
In table 7.1, we have these range of colors, I was trying to use DIMGRAY, but looks like it's not defined in the read file as the other colors. Although, its RGB can be used.
Can someone take a look. Thanks !

Er9y714:
https://github.com/firemodels/fds/blob/48df14d7d3df0eb3c8eb496a0acd1d21a23358b9/Source/data.f90#L1513

I guess it should be with a blank ; DIM GRAY

sbenkorichi:
Nice spot.
I will make the correction. Thanks !

sbenkorichi:
Done !

PR #5684, opened 2017-11-08
User_Guide: fix typo in table 7.1

sbenkorichi:
Correct DIM GRAY color name, reported in #5683


sbenkorichi:
Note: I've checked with color table in wiki page, it's set correctly.

rmcdermo:
Thanks!

PR #5697, opened 2017-11-10
FDS Source: fix TMP_F for IOR=+/-3 INTERPOLATED bcs

rmcdermo:


rmcdermo:
Meant to mention that this applies to STRATIFICATION=T (default) case.

PR #5702, opened 2017-11-14
Fix qfds for SLURM at CSC(Finland)

shostikk:
This now works at taito.csc.fi
- nodes cannot be higher than 1 in the 'serial' queue. Now it is not set at all, and SLURM will decide.
- walltime must be set
- MPIRUN command is srun

Could not test for not breaking PBS.

### DIFF
--- a/Utilities/Scripts/qfds.sh
+++ b/Utilities/Scripts/qfds.sh
@@ -247,10 +247,12 @@ shift $(($OPTIND-1))
 
 # ^^^^^^^^^^^^^^^^^^^^^^^^parse options^^^^^^^^^^^^^^^^^^^^^^^^^
 
-if [ "$RESOURCE_MANAGER" == "SLURM" ]; then
-  walltime=99-99:99:99
-else
-  walltime=999:0:0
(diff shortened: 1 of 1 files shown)


mcgratta:
Thanks, Simo. 

Randy, Marcos -- do you see any problems with these changes if running on AWS?

shostikk:
Are you using Slurm as well?

L?hetetty iPhonesta

Kevin McGrattan <notifications@github.com<mailto:notifications@github.com>> kirjoitti 14.11.2017 kello 15.42:


Thanks, Simo.

Randy, Marcos -- do you see any problems with these changes if running on AWS?

-

rmcdermo:
We'll see when we get there.  Every time I log into scorch there is something not working.  So, I am yet to have a successful run.  That doesn't mean that the next time I try it will be ok.  But it generally is not a 5 minute exercise.

mcgratta:
On the NIST clusters, we use OpenPBS/Torque. If you just execute qfds.sh like this:
```
qfds.sh -p 12 -v simo.fds
```
you can see what the default settings are.

PR #5755, opened 2017-12-01
MPI_Scaling_Tests for ScaRC and some GMG-cases

SusanKilian:
Hi, all,

this is a new pull request with:

- some new weak and strong scaling tests for ScaRC in MPI_Scaling_Tests corresponding to the existing tests for FFT and GLMAT
- two further experimental 2D cases pressure_iteration2d_scarc_gmgc.fds and pressure_iteration2d_scarc_gmgb.fds with settings for geometric multigrid with different interpolation techniques between the grid levels
- corresponding changes in the scrc.f90-code 
- revision of the output for ScaRC in dump.f90

Hopefully, everything works. I checked the compilation with the Intel- and the GNU-compiler  (without any warnings). 

Best, Susan


rmcdermo:
@SusanKilian Compiles fine with Intel on our end.  Please note that we will be shutting down our cluster for TWO WEEKS for power upgrades.  So, we will not have nightly complete firebot runs for that time.  If you send PRs during that time, please run firebot on your end first.  Thanks.

SusanKilian:
Hi, Randy, alright, I'll keep that in mind. S.

sbenkorichi:
Nice @SusanKilian 
Wouldn't be possible to add some explanation in the Manuals as well on these cases. Thanks !

SusanKilian:
Okay, I'll have a look at this. Currently, I don't have any idea how ScaRC behaves on that many meshes, because I don't have the possibility to check it here (I only tested up to 64 meshes). Please let me know if there are some troubles with scaling.

PR #5766, opened 2017-12-07
FDS Source: Add OBST_ID to VENT line to uniquely assign VENT.

mcgratta:


mcgratta:
This input makes it easier to tie a `VENT` to a specific `OBST`. I have a case with an obstruction with a vent on top that moves a little bit every so often. The `VENT`s were being applied to multiple instances of the `OBST`. This input prevents that.

PR #5799, opened 2017-12-19
FDS Source: restructure HT3D input to accept MATL_IDs from SURF

rmcdermo:
In this PR, I am setting the input logic to accept MATL_IDs from a SURF for HT3D.  This is in preparation for 3D pyrolysis.  The idea is that one SURF will serve to specify the MATL_IDs.  This SURF can have only 1 layer, but a mixture of MATL_IDs, with the bulk material properties taken as the volume weighted average as processed in READ_SURF.


sbenkorichi:
Good job man :+1:

PR #5805, opened 2017-12-21
FDS Source: Major Change, create a stand-alone pyrolysis subroutine

mcgratta:


mcgratta:
Simo, could you please have a look at these changes. In particular, the `PYROLYSIS_LIQUID` routine. The new `PYROLYSIS` subroutine can only work on a single solid phase node (i.e. the `I=1,NWP`). In the old `PYROLYSIS` routine, we assigned mass flux `MFLUX` to multiple nodes, starting from the surface. Now, the liquid pyrolysis model extracts mass ONLY from the surface layer. The verification cases work, but there are not too many of them.

PR #5809, opened 2017-12-21
FDS Verification: Shorten symmetry test case

mcgratta:


mcgratta:
This change just shortens this test case from 5 s to 4 s.

PR #5820, opened 2017-12-23
FDS Source: Read OBST lines for neighboring meshes.

mcgratta:


mcgratta:
This commit fixes `ht3d_2mesh_x.fds` by forcing each MPI process to read and allocate `OBST`s from its own and neighboring meshes.

PR #5823, opened 2017-12-26
FDS Source: Do not process NOM=0 in HT3D loop

mcgratta:


mcgratta:
Randy, this is a simple fix to the problem, but check it. I suspect this has do with the external boundary of a single mesh where the `OBST` abuts it.

PR #5867, opened 2018-01-09
FDS V&V: Switch dataplot and scatplot to use LaTeX again

mcgratta:


rpeacoc:
I assume we should lift these for CFAST as well, correct?

mcgratta:
No, CFAST is OK. I converted CFAST back to LaTeX last year. FDS has the added problem of many scripts that use TeX. I just added a directive to use LaTeX into dataplot.m and scatplot.m, but CFAST is already doing this.

PR #5868, opened 2018-01-09
cleaning up the Verification/scripts directory

gforney:
wanted to add a script to do some openmp testing and saw that many scripts were not being used. many of these scripts were holdovers from before we split the fds and smv repos .


mcgratta:
https://www.nbcnews.com/better/health/what-swedish-death-cleaning-should-you-be-doing-it-ncna816511

sbenkorichi:
😆 😆

PR #5888, opened 2018-01-14
FDS Source: allow OBST MATL_SURF_INDEX from SURF line

rmcdermo:
In theory, now HT3D will take the material properties specified on a SURF line.  See ht3d_A_B.fds as an example.


sbenkorichi:
You did it !!!

PR #5928, opened 2018-01-20
Update for ht3d_sphere case

sbenkorichi:
@rmcdermo 
Please review. I have made the necessary changes. It should compile fine.


rmcdermo:
@sbenkorichi 

On the 101 case, first IJK are integers, please change "51.0" to 51.  Second, if IJK=51 and I_UPPER=1, doesn't this make the case 102 points across?  I do not understand why this gives a better result than the 100 point case.

sbenkorichi:
@rmcdermo 
I have corrected the integer. I will need to fix the web script to avoid this.
As for the cells nbr, I set it to 102. The reason for doing this is because I need to get the devices set in the cell center. Using 100 isn't helpful. Unless maybe changing the position of the OBST.

rmcdermo:
Next you need to change 51.0 in the 51 case.  Then you need to remove the 101 case.

sbenkorichi:
Ah sorry for that. I missed it.

sbenkorichi:
@rmcdermo 
Now?

Utilities/Matlab/scripts/ht3d_sphere.m
```diff
@@ -82,9 +82,9 @@
 %% gather FDS results
 
 ddir = '../../Verification/Heat_Transfer/';
-fnt = {'ht3d_sphere_50'};
-fileName = {'ht3d_sphere_25','ht3d_sphere_50','ht3d_sphere_100'};
-nc_array = [25,50,100];
+fnt = {'ht3d_sphere_51'};
+fileName = {'ht3d_sphere_25','ht3d_sphere_51','ht3d_sphere_102'};
+nc_array = [25,51,101];
```
rmcdermo: Should nc_array(3) be 102?

sbenkorichi:
Thanks for the catch.

rmcdermo:
OK, thanks!

sbenkorichi:
Welcome, hopefully firebot wont complain about it.

PR #5960, opened 2018-01-27
FDS Source: Clean up RESIDUE variables

mcgratta:


mcgratta:
I ran the verification cases and there was nothing out of tolerance.

PR #5961, opened 2018-01-28
FDS_User_Guide: Add an example for heat detector to the hrr_freeze.

sbenkorichi:
Add some clarification and an example to the DEVC/CTRL functionality. 
@drjfloyd  Review please.


drjfloyd:
Why is another example needed? The text at the start of the section indicates that this works with any device.  The input example in the Guide is not even a sprinkler as suggested in your edits, it is just a gas temperature device. I think an example for one device is sufficient.

sbenkorichi:
The input example for a heat detector not a sprinkler. 
I just thought it would be better to give more explanation. 
I will close it. Thanks !

PR #5994, opened 2018-02-07
FDS Source: Fix pyrolysis output quantity

mcgratta:


mcgratta:
Randy, this change is what we talked about. Now `M_DOT_S_PPP` better represents the mass gain or loss per unit time per unit volume of solid species alpha.

PR #6008, opened 2018-02-12
FDS Source 

ruddymell:
Changes to FDS to implement the current WFDS boundary fuel model for vegetation and the level set model for fire front propagation. The level set model works only on one mesh and needs the  two dimensional array LSET_PHI to be MPI "enabled". Most changes are to vege.f90, init.f90, func.90, type.f90, read.f90. Some changes to wall.f90 were made for the vegetation pyrolysis model. An additional convective heat transfer model for cylinders was added in wall.f90. Additional changes were made to other .f90 files to control the flow when the level set model is run in a manner uncoupled from the CFD solver (similar to SOLID_PHASE_ONLY=.TRUE.).


rmcdermo:
Kevin,
Please test this as a topic branch asap, since we will have cluster down on Wed.  Thanks.

rmcdermo:
Ruddy,
You have a ton of things here that cannot be merged.  For example, all the .slog files, .gsf files (binary, not sure what this is).  Please check the "changed files" before hitting submit on these PRs.  Thanks

rmcdermo:
Given that you have these binaries, we need you to do a clean clone of the repo, copy your changes over to it, then submit a PR with only those commits that contain the changes to source.  If this is too confusing, then just email the source code to us.  Thanks.

Please understand how hard it is to undo commits that create problems, one binary could bloat the repo irreversibly.   Sorry for the hassle.

mcgratta:
It appears that you did a global `git add` such that all unversioned files in your repo got added. You should just add individual files.

rmcdermo:
Yes, but the key now is that you need to use "reset" and not "revert". If you just revert, then this doubles the size of the repo (it adds and removes the binaries).

PR #6018, opened 2018-02-13
FDS Source: Add global RADIATIVE_FRACTION to COMB

mcgratta:


drjfloyd:
Radaitive fraction is not global. It is tied to a specific reaction.

mcgratta:
That option still exists. If you set `RADIATIVE_FRACTION` on the `COMB`, then this value is applied to all reactions. If you want to specify each reaction, you can do that too.

PR #6022, opened 2018-02-14
Doc: Add 'PARTICLE WEIGHTING FACTOR' to particle output options.

marcfehling:
We stumbled over the PARTICLE WEIGHTING FACTOR in our studies, and thought it should be mentioned in the User Guide.
Feel free to change anything you like.

Best, Marc


mcgratta:
I will add to the code and then merge your change. Thanks.

mcgratta:
Sorry, I just noticed that this quantity is in the code but we forgot to add to the Guide. Thanks for catching that.

PR #6028, opened 2018-02-14
FDS Tech Guide: Define tau_chem

mcgratta:


mcgratta:
Randy, could you check my wording, and add if you see fit. I just want to provide a bit more on tau_chem and what it means.

PR #6040, opened 2018-02-16
FDS Source: Change some variable bounds to prevent instabilities.

mcgratta:


mcgratta:
Randy, have a look at this. For typical compartment scenarios, `RHOMIN` is about 0.04 kg/m3, based on the MW of soot (12 g/mol) and `TMPMAX` of 3000 K. I believe that for some of these ISO compartment experiments (NIST RSE/FSE) that this can sometimes lead to numerical instabilities. Not often, and not consistently. Increasing `RHOMIN` to about 0.08 kg/m3 solved the problem, at least for now.

drjfloyd:
Do we want RHOMAX that restrictive?  That is saying at our default TMPA of 20 C we couldn't have temperatures anywhere that are slightly below zero.

PR #6042, opened 2018-02-16
FDS Source: Relax restriction of RHOMAX

mcgratta:


mcgratta:
Just remembered that in many pressure rise cases, we need to allow for a significant increase in density. We cannot determine at the start what that increase is going to be, so a factor a 2 ought to be sufficient. FDS should not be used to push beyond 1 atm over-pressure.

PR #6072, opened 2018-02-26
FDS Verification: Add new test case for synthetic eddy

mcgratta:


mcgratta:
Randy, just to check things first. Go to `Utilities/Matlab` and start matlab. 
```
addpath 'scripts'
synthetic_eddy_method_2
```
There should be 5 plots in `FDS_User_Guide/SCRIPT_FIGURES`. The all start with `BL_Stats`

The rough cut input file is called `Verification/Turbulence/sem_test.fds`

PR #6088, opened 2018-02-28
 emanuelegissi: add all the _geom.fds Vettori cases, update Run_All.sh

emanuelegissi:


emanuelegissi:
As requested, I will not update the Run_All.sh

PR #6108, opened 2018-03-02
FDS Verification: Start of wake drag reduction verification

btrettel:
This pull request includes two verification cases and partially documentation for obscure FDS configuration.

FDS can have periodic particles, specified with PERIODIC_X=.TRUE., etc., on the PART line. This feature seems to have existed as of 2012, but was undocumented. I use it here to avoid having a very large domain for the terminal velocity test. The test case periodic_particle.fds works as expected.

wake_drag_reduction_periodic.fds is the start of a wake drag reduction test. The test has randomly generated particles fall under the influence of gravity, eventually reaching a terminal velocity. A fairly high particle density is necessary to see an effect here. For the moment I have not seen any noticeable difference with respect to having this feature on or off. E.g.:

10000 particles in a periodic 2 m x 2m box (40x1x40 grid):

W_term = -22.2862923263995 m/s (on)
W_term = -22.2862923263995 m/s (off)

Higher number of particles would take considerably longer to run than the 5-10 mins this case does.


rmcdermo:
@btrettel Let's practice getting this done in one clean PR.  Here is what we need to do:

1. Add write up to UG of PERIODIC_X.  It looks to me like you just added this to Table at the end.
2. Verification cases need to be added to FDS_Cases.sh
3. Cases need to be added to dataplot to generate plots.
4. Plots need to be added to verification guide with short writeup.
5. PR should not include merge commits.  Stash your local changes and apply them to an up-to-date repo then commit and send PR.

rmcdermo:
I'm closing this so no one accidentally accepts it.  Send a new PR once the above is done.  Let me know if you have questions.

mcgratta:
There is a [wiki](https://github.com/firemodels/fds/wiki/Verification-Case-Setup-Example) that describes the process of doing a verification case.

PR #6111, opened 2018-03-02
FDS Verification: add viewpoint to ini files

rmcdermo:


gforney:
I'll change smokeview so that it won't generate images when running a
script if it can't find a viewpoint. So it will be more obvious something
is wrong

PR #6117, opened 2018-03-05
FDS Verification: Adjust HVAC mass transport cases to remove bad ZONEs.

drjfloyd:


bmralph:
Hi Jason. I see you've taken out `DT_HVAC` which was keeping these cases stable in the past; is this related to the recent changes to HVAC (e.g. pressure BC) or something in the code? Cheers!

drjfloyd:
The issue was having a ZONE with an OPEN boundary.  I need to get the HVAC model to look at CONNECTED_ZONES.  In this case the ZONEs don't do anything and removing them means DT_HVAC isn't needed.  Also long term I am probably getting rid of DT_HVAC.  I have a potentially better solution.

PR #6133, opened 2018-03-08
FDS Source: Change notation of EXTINCT_2, but not functionality

mcgratta:


mcgratta:
Randy, Jason -- take a look at the Tech Guide section on Extinction. I have implemented a notation that is more succinct and easier to understand, I hope. I modified the code accordingly. There should be no change in functionality.

PR #6158, opened 2018-03-16
Comprehensive revision of data structures and adding of more multi-level solvers

SusanKilian:
The data structures for the administration of different solver hierarchies and grid levels with different resolutions are substantially revised and expanded. 

A two-level Krylov-method and a Krylov-method with preconditioning by a global Multigrid method are added, as well as corresponding pressure_iteration-test cases for 2D and 3D.

See https://github.com/firemodels/fds/issues/1105 for more details.


mcgratta:
Susan, you are pushing up many unversioned files. Also, why is there an impi_intel_osx_64 target? I do not believe that OSX allows Intel MPI. `impi` means Intel MPI. 

Do you push too many files at once. We cannot test all of this.

SusanKilian:
Sorry, it's not clear to me what going wrong. I thought that I only would push read/scrc and some pressure_iteration-files. Let me have a look to it ...

SusanKilian:
I am still not familiar with the pull requests, sorry. I suppose, I must do a rebase now to restrict to only the required files?

rmcdermo:
Susan,
Now that you have committed all these files that are untracked, you have to do a hard reset (or else delete your repo and reclone).  Recloning requires you to purge your fork and start over, not the best option.  Here is a way around it.

1. Make a copy of all the source files and input files you have modified that you want to send.  Put these in a different directory, outside the repo.
2. Do a [git reset --hard](https://git-scm.com/docs/git-reset) to the commit prior to all your changes.
3. Do a [git push -f origin master](https://git-scm.com/docs/git-push).
4. Do a remote update and merge, push to origin.
5. Now you can copy your source changes back to the clean, up-to-date repo.  Do a git diff.  You should only see the changes you intended to make.
6. Push to your fork and send a PR.  You should only see your changes.

sbenkorichi:
I had some issues before. What I would prefer to do, is to fork a new fresh repo and then clone it locally, then modify each file that my work is related to (from your current repo). This way, all your commits will be up to date, which makes it easier for reviewing.

Another way, I would copy the fds repo in your OS to a different folder, then, in the original /fds, I would do
`$ git reset --hard firemodels/master`
This will make your local repo up to date with the firemodels/fds 
then you can make all the changes you want, and then you push it to the github, by
`$ git push -f origin master`
Then you can send a new pull request.

There are other ways as well, Randy can give other alternative ways.

PR #6160, opened 2018-03-16
FDS Source: new twolevel-Krylov and Krylov-GMG-version of ScaRC

SusanKilian:
Next try:

The data structures for the administration of different solver hierarchies and grid levels with different resolutions are substantially revised and expanded. 

A two-level Krylov-method and a Krylov-method with preconditioning by a global Multigrid method are added, as well as corresponding pressure_iteration-test cases for 2D and 3D.


rmcdermo:
@SusanKilian Good job!  Thanks for the contribution.  Now, are these test cases in the verification suite?  Can they be added to FDS_Cases.sh?  Are all the SCARC parameters documented in the FDS User Guide?

SusanKilian:
Currently, these cases are not included in the FDS_Cases.sh.

Earlier, we had the former scarc2d- and scarc3d-cases there, but I changed some input parameters so that they didn’t work anymore and replaced them by the new cases.

If you don’t mind, I would like to continue with the documentation of these test cases and the ScaRC parameters now ...

rmcdermo:
OK, just to be clear, you mean the cases have been removed(?)  I do not see any ScaRC cases in Pressure_Solver in FDS_Cases.sh now.  It is really important that we get test cases committed and running.  Thanks.

SusanKilian:
Yes, I mean the former scarc2d- and scarc3d-cases, which I renamed/modified on 21-th November because they are not only related to ScaRC, but all pressure solvers, see b9a9d9d11.

I’ll add the new versions to FDS_Cases.sh right at the beginning of next week.

Have a nice weekend, Susan

PR #6167, opened 2018-03-21
FDS Verification: activate pressure_iteration cases and add corresponding Matlab scripts

SusanKilian:
Based on the former scarc2d- and scarc3d-cases I created some pressure_iteration test cases in 2d and 3d which I already added to FDS_Cases.sh. For a simple square/cube-domain (each with 8 meshes) they compare the different pressure solvers:

- default FFT on corresponding 1-mesh case
- default FFT
- tight FFT (tol = 0.00001)
- GLMAT
- ScaRC-CG 
- tight ScaRC-CG (tol = 0.00001) 
- ScaRC-CG with GMG-preconditioning
- ScaRC-GMG
- ScaRC-CG with additive coarse grid
- UScaRC-CG

I added corresponding Matlab scripts which compare the results for the pressure device, the number of pressure iterations, the velocity errors. For the ScaRC cases also the number of ScaRC iterations and resulting convergence rate is plotted. There is also a zoom plot for the pressure device in the 2d-case. I removed the former script scarc2d.m.


rmcdermo:
Thanks, Susan.

Is there a reason these plots cannot be made with dataplot.m?

If so, then the scripts need to be added to the main FDS_verification_script.m.

Please read the documentation at the top of dataplot.m.  You can use an "f" flag to "follow" the previous plot and add comparions.

SusanKilian:
Okay, I'll have a look to it and test it ...

rmcdermo:
Here are some notes on how to use it:

https://github.com/firemodels/fds/wiki/Using-the-Matlab-script-dataplot.m

But also see the top of the script for comments on the flags in the switch_id

https://github.com/firemodels/fds/blob/master/Utilities/Matlab/scripts/dataplot.m

Thanks.  I know this seems like a pain.  But once you do one, it becomes easy.

SusanKilian:
Yes, I believe that and am really a fan of such scripts. It just the startup ... I am just reading the stuff. 
So, is it right that I should specify the User Guide as destination for the plots (like the dancing_eddies)?

rmcdermo:
I would expect the plots to be in both user guide and verification guide.  In cases where there is already a GLMAT example in the user guide, add your ScaRC example in a separate plot in that section.

But remember that these will not go to the guide until you are ready to release.  Any case that is in dataplot must also have a section in the guide (this is checked by firebot).  So, until ScaRC is ready for release, you can just have the cases running in FDS_Cases.sh.  But it would be good to try to get ScaRC in as Beta release for upcoming version 6.7.0.

SusanKilian:
Sounds good, I will push it as soon as possible. So, for the moment - until I have dataplot running - I would like to have the pressure_iteration cases running via FDS_Cases.sh. I'll do another pull request to include them (without the Matlab scripts)

PR #6170, opened 2018-03-21
FDS Verification: activate pressure_iteration cases

SusanKilian:
Based on the former scarc2d- and scarc3d-cases I created some pressure_iteration test cases in 2d and 3d which I already added to FDS_Cases.sh. For a simple square/cube-domain (each with 8 meshes) they compare the different pressure solvers:

default FFT on corresponding 1-mesh case
default FFT
tight FFT (tol = 0.00001)
GLMAT
ScaRC-CG
tight ScaRC-CG (tol = 0.00001)
ScaRC-CG with GMG-preconditioning
ScaRC-GMG
ScaRC-CG with additive coarse grid
UScaRC-CG


rmcdermo:
Susan,
In this case you want to use a reset, not a revert.  If I accept this PR, then I'm adding the cases and removing the cases to our main trunk, increasing the size unnecessarily.  It may be minor for this case, but I want to be vigilant about unnecessary commits.  Thanks.

SusanKilian:
I see, that's right ... I'll clean it tomorrow

PR #6177, opened 2018-03-22
FDS Source: Generalize processing of SMOKE3D

mcgratta:


mcgratta:
Glenn -- these changes should not affect old functionality. But the order of the quantities matters because of the logic in SMOKE3D_TO_FILE.

PR #6186, opened 2018-03-23
FDS Verification: add unstructured ScaRC-CG version for duct_flow case

SusanKilian:
This is the ScaRC-CG version with PARDISO-preconditioning for the unstructured pressure matrix with matching volume flows and  only 1  pressure iteration.


rmcdermo:
Nice work!

PR #6192, opened 2018-03-25
FDS Verification Guide: Fix bad reference

mcgratta:


mcgratta:
When adding a verification case to the User's Guide or Verification Guide,, label the section where the case is described with exactly the name of the case which is in the second column of the `FDS_verification_dataplot_inputs.csv` file.

PR #6194, opened 2018-03-25
minor edit to qfds.sh, when using the -v option, don't output '-v' when outputting the command line to the script

gforney:


gforney:
missed this change in original pull request for this feature

PR #6198, opened 2018-03-26
FDS Validation: add 864-mesh scaling cases for ScaRC

SusanKilian:
I also added the 864-cases for default ScaRC.


rmcdermo:
Susan,
We do not have that many cores to test this.

PR #6203, opened 2018-03-27
FDS Source: fix unused warning compile issue for AS in scrc

SusanKilian:
Use it only if the MKL-libraries are used


rmcdermo:
Works.  Thanks!

PR #6204, opened 2018-03-27
FDS Validation: Fix a few input files and docs

mcgratta:


drjfloyd:
For the FSE section, was there a reason we skipped three tests? If so I think we should state it.

For the cases where we had to define the fuel, we should probably also add RADCAL_ID and RADIATIVE_FRACTION.

mcgratta:
Not sure what the RADCAL_ID should be. I'm going with default. RAD_FRAC is 0 for all  cases -- we're predicting it.

PR #6207, opened 2018-03-28
FDS Source: fix further MKL compile warnings and add STOP_CHECK in main after SCARC_SETUP

SusanKilian:
Put all MKL-related stuff into compiler directives and add error message if MKL-solvers are called without MKL library compiled using STOP_CHECK-mechanism


rmcdermo:
Great work!  Thanks!

PR #6211, opened 2018-03-29
Making wake drag reduction apply to both liquid droplets and solid particles

btrettel:
Title is self-explanatory. I do not know if wake drag reduction was not applied to solid particles for a reason. I can think of one: Solid particles are not necessarily spherical as the wake drag reduction model assumes. It may be desirable later to add some sort of "shape factor" for particles which are not spherical, but I suspect the model is sufficiently accurate for non-spherical particles as-is.


Source/part.f90
```diff
@@ -1649,13 +1649,11 @@ SUBROUTINE MOVE_IN_GAS
...
-      ! Drag reduction model for liquid droplets
+      ! Drag reduction model for liquid droplets and solid particles
 
       WAKE_VEL=1.0_EB
-      IF (LPC%LIQUID_DROPLET) THEN
-         DROP_VOL_FRAC = MIN(1._EB,AVG_DROP_DEN(IIG_OLD,JJG_OLD,KKG_OLD,LPC%ARRAY_INDEX)/LPC%DENSITY)
-         IF (DROP_VOL_FRAC > LPC%DENSE_VOLUME_FRACTION) CALL WAKE_REDUCTION(DROP_VOL_FRAC,LP%RE,C_DRAG,WAKE_VEL)
-      ENDIF
+      DROP_VOL_FRAC = MIN(1._EB,AVG_DROP_DEN(IIG_OLD,JJG_OLD,KKG_OLD,LPC%ARRAY_INDEX)/LPC%DENSITY)
+      IF (DROP_VOL_FRAC > LPC%DENSE_VOLUME_FRACTION) CALL WAKE_REDUCTION(DROP_VOL_FRAC,LP%RE,C_DRAG,WAKE_VEL)
```
rmcdermo, line 1656: @btrettel Are you sure that `LPC%DENSE_VOLUME_FRACTION` is defined for `.not.LPC%LIQUID_DROPLET`?

btrettel: I believe yes, as DENSE_VOLUME_FRACTION is on the PART line and doesn't seem to have anything specifically to do with liquid droplets or solid particles. I'll look at what the FDS code does, however.

rmcdermo:
Ben,
Don't just change the code.  First get a test case that works.  Your goal is to demonstrate that LPC%DENSE_VOLUME_FRACTION affects the transport and WAKE_REDUCTION algorithm for solid particles.  Get all that working on your end.  Send us a section of the guide that shows it works and then send us the code changes and verification case.

PR #6212, opened 2018-03-29
FDS Source: Add TAU_DEFAULT to MISC line and remove NO_RAMPS

mcgratta:


mcgratta:
Instead of `NO_RAMPS=T` on `MISC`, set `TAU_DEFAULT=0`.

PR #6220, opened 2018-04-02
FDS user guide: add text related to total mass output

bmralph:
Following confusion from a student, adding some text here regarding the output of total mass.


rmcdermo:
Thanks, Ben!

PR #6223, opened 2018-04-03
FDS Source: rework input parameters for ScaRC/UScaRC to set simple defaults

SusanKilian:
The default versions of ScaRC can now be simply called by SOLVER='SCARC' or SOLVER='USCARC' in the &PRES-namelist. By default, the first one calls a global CG-method with meshwise FFT-preconditioning (based on Crayfishpak) working on the structured pressure matrix, the second one a global CG-method with meshwise Pardiso-preconditioning (based on Intel MKL) working on the unstructured pressure matrix.


rmcdermo:
Susan,
Looking good.  Please take a quick look at the source conventions for logical statements here:
https://github.com/firemodels/fds/wiki/Developer-Commit-Guidelines
and modify accordingly.  Then we can merge.  Thanks.

SusanKilian:
Hopefully, it's okay now ...

rmcdermo:
Thanks!

Marcos and I were wondering: how is SCARC handling mesh refinement interfaces now?  Does one use USCARC with mesh refinement or does it matter?

SusanKilian:
Currently, ScaRC needs the same refinement level on adjacent meshes. Some time ago, I already worked on a corresponding matrix generation for different refinement levels, but this isn't incorporated in the current code (because there still were some cases for which it didn't work at that time). I would have to revive it ... Please note, that I really need the global matrix including the corresponding complex stencils. But it would be great to discuss that at a favorable time.

rmcdermo:
OK, this needs to be mentioned in the documentation.  And also you need to make sure to exit the code gracefully if someone tries to use SCARC with mesh refinement.

But I agree, let's get uniform interface working well before moving on to refinement interfaces.

PR #6224, opened 2018-04-03
FDS Verification: provide simple default ScaRC versions for duct_flow and dancing_eddies

SusanKilian:
The input parameters for the dancing_eddies- and duct_flow-cases are now reduced to the default settings (structured/unstructured).


rmcdermo:
Very good.  Thanks!

PR #6227, opened 2018-04-03
FDS Source: Issue #6222. Correct minor coding problems that were affe…

mcgratta:
…cting Intel 18 compiler


mcgratta:
This commit also adds hydrogen to the two-step chemistry model.

PR #6239, opened 2018-04-05
Matlab: Correct title in figure 6.8 of UG (pressure iterations for dancing_eddies)

SusanKilian:
I fixed the title again. It’s related to the current figure 6.8 of UG where the title ‚CPU Time (dancing_eddies)‘ is used both in the plot for the pressure iterations and the plot for the CPU time.


rmcdermo:
Got it.  Thanks and sorry for this.

SusanKilian:
No problem at all ...

PR #6241, opened 2018-04-06
FDS Source: add single precision version of MKL preconditioning in ScaRC

SusanKilian:
I introduced a compiler directive WITH_MKL_FB (at the top of scrc.f90) which is currently undefined.
By defining it, the MKL-preconditioning is only done in single precision mode whereas the surrounding CG-method still uses double precision mode. 

As discussed with Marcos, the "_S"  versions of PARDISO and CLUSTER_SPARSE_SOLVER were added to geom.f90.

The default accuracy for CG was set to 1E-10 and the error messaging for ScaRC was refined.


marcosvanella:
Good work Susanne! I'll test USCARC when the cluster becomes available.

PR #6270, opened 2018-04-16
FDS Source: Add experimental BLOCKAGE_FACTOR for radiation

mcgratta:


mcgratta:
This feature may or may not stay. I have not committed any validation or verification cases for it.

PR #6274, opened 2018-04-16
FDS Utilities: Clean up qfds.sh

mcgratta:


mcgratta:
Note that when using Intel MPI with more than one OpenMP threads, qfds will automatically map one MPI process per socket, 2 per node, and grant the job exclusive rights to the entire node. When using Open MPI, only the cores you request will be allocated.

PR #6296, opened 2018-04-18
FDS Source: Define VVLES parameters

mcgratta:


Source/read.f90
```diff
@@ -1739,45 +1739,36 @@ SUBROUTINE READ_MISC
...
+   VN_MAX = 0.5
+   VN_MIN = 0.4
+   FLUX_LIMITER = CHARM_LIMITER
+   IF (TURBULENCE_MODEL/='null') THEN
+      WRITE(MESSAGE,'(A,A,A)')  'ERROR: TURBULENCE_MODEL,',TRIM(TURBULENCE_MODEL),', is not appropriate for DNS.'
+      CALL SHUTDOWN(MESSAGE) ; RETURN
+   ENDIF
 ELSEIF (RESEARCH_MODE .OR. SIMULATION_MODE=='LES') THEN
    SIM_MODE = LES_MODE
+   FLUX_LIMITER = SUPERBEE_LIMITER
```
rmcdermo, line 1754: @mcgratta I think LES should use CHARM (FLUX_LIMITER=2).

mcgratta:
Will do. I'm testing again.

rmcdermo:
Of course, I meant FLUX_LIMITER=CHARM_LIMITER (which is actually 4).

Do we want to take this opportunity to make FLUX_LIMITER a character string for input, instead of the cryptic integer designation?

drjfloyd:
I think that would be a good idea.

PR #6299, opened 2018-04-18
FDS Source and Manuals: Miscellaneous changes.

marcosvanella:
- Add GET_CUTCELL_VERBOSE to &DUMP line, to allow for cut-cell computation progress monitoring.
- Increase size of local arrays used in SET_CUTCELLS_3D. Still need to make some of these allocatable.
- Use \texttt instead of \ref for references to Verification Guide in geom_notes. This is to avoid warnings.


rmcdermo:
Marcos, you have not merged.

PR #6307, opened 2018-04-20
FDS Source: Reduce error for HVAC node temperatures, fix node tempera…

drjfloyd:
…ture update for downstream vent nodes


bmralph:
Excited to test this out. I think this may be causing some problems in a
few of my test cases...

mcgratta:
Any reason why this is not being merged?

drjfloyd:
Just forgot to click merge.

PR #6315, opened 2018-04-23
FDS tech guide: add extra explaination

bmralph:
Added extra definition of particle diameter D (2r), following confusion from a student.


mcgratta:
Thanks, K

PR #6317, opened 2018-04-23
FDS Source: All STRATIFICATION and CONSTANT_CP

mcgratta:


rmcdermo:
Kevin,
On a PR like this, please give some explanation for why you are making this change.  Thanks.

mcgratta:
For some simple compartment validation sets involving HVAC, like the `LLNL_ENCLOSURE` , the pressure stratification does matter. The error associated with the equation of state is much smaller than the error associated with the assumption of diatomic molecules. Running the validation cases will demonstrate this.

PR #6325, opened 2018-04-24
Add verification case for Frankman, Combust. Sci and Tech, 2010 on Fi…

tmilac:
…ne Fuel Heating by Radiant Flux

As described, FDS does not accurately reproduce this case because of the way it handles the calculation of the convective cooling coefficient.  FDS should probably be extended to allow for a more flexible handling of this coefficient, minimally so as to be able to choose natural convection only. We can submit code that does the latter and closely reproduces the Frankman results.

The cases could not be run simultaneously because of the large variation of radiant panel emissivity reported.

### DIFF
--- a/Manuals/Bibliography/FDS_general.bib
+++ b/Manuals/Bibliography/FDS_general.bib
@@ -1531,6 +1531,15 @@
   publisher    = {Cambridge Univ Pr},
 }
 
+@ARTICLE{Frankman:2010,
+  author       = {Frankman, D. and Webb, B. W. and Butler, B. W. and Latham, D. J.},
+  title        = {{Fine fuel heatin g by radiant flux}},
+  journal      = {Combust. Sci. and Tech.},
+  volume       = 182,
+  pages        = {215--230},
+  year         = 2010
+}
+
 @TECHREPORT{Friday:1,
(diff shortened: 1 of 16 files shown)


mcgratta:
Tom, send me the compiled pages from the Verification Guide. Thanks

rmcdermo:
Does the Rayleigh model work for this?
```
&SURF ..., HEAT_TRANSFER_MODEL = 'RAYLEIGH'/
```
This is what we are likely moving toward for convective heat transfer.

mcgratta:
These input files are very complicated, and it is difficult to know if the problem lies with the convective heat transfer coefficient. I just created this input file (copy to frankman.fds): 
[frankman.txt](https://github.com/firemodels/fds/files/1948600/frankman.txt)
I make very simple assumptions about the vegetation. Here are the results:
[frankman.pdf](https://github.com/firemodels/fds/files/1948604/frankman.pdf)
This case runs in a few seconds. I have not had the chance to do grid sensitivity, etc. 

I'd like to rethink where this ought to go. Now that I am more familiar with it, it really does fall in the category of validation. But before setting up plots, play with the file a bit. Make sure the results are not just a fluke.

PR #6364, opened 2018-05-02
FDS Validation Guide: Fix Cup Burner write up, adjust front matter

mcgratta:


mcgratta:
Jason, whatever editor you are using to edit the FDS manuals turns "fl" into some kind of weird character. It appears to be a non-recognized single character. Make sure you use a simple text editor or TeX specific editor.

drjfloyd:
Was the fl in the table? I am scanning the changes you made on github  trying
to see where this happened.

Generally I use a Tex specific editor.  For the table of cup burner data, I
built that using my Excel workbook of the experimental data using
CONCATENATE and then pasted it into the tex editor.

mcgratta:
It was in the text part. The word "flame" became &*^%ame. coflow because co&&^&%ow. When I deleted the fl, it deleted as a single character. It is probably the remnant of a special character that sets fl as a single special joined character.
[FDS_Validation_Guide 71.pdf](https://github.com/firemodels/fds/files/1968189/FDS_Validation_Guide.71.pdf)

drjfloyd:
I am not sure how that happened. I just tried typing fl into the editor I
use and it stayed as two characters when saving and reopening.

mcgratta:
It does not appear in the current release version of the Guide. Just one of those things.

PR #6383, opened 2018-05-08
FDS Source: fix error related to combination of periodic and Neumann …

SusanKilian:
I fixed an error related to the condensed system which only occurred when periodic and Neumann BC's are combined (in the absence of Dirichlet BCs).
Furthermore I temporarily deactivated an array only needed for the Algebraic Multigrid which isn't used currently.

### DIFF
--- a/Source/scrc.f90
+++ b/Source/scrc.f90
@@ -387,7 +387,7 @@ INTEGER, PARAMETER :: NSCARC_STACK_ZERO              =  0, &         !> root sta
                       NSCARC_STACK_MAX               = 10            !> maximum number of consecutive solvers
 
 REAL(EB), PARAMETER:: NSCARC_THRESHOLD_CONVERGENCE   = 1.0E-15, &    !> threshold for convergence
-                      NSCARC_THRESHOLD_DIVGERGENCE   = 1.0E+15, &    !> threshold for divergence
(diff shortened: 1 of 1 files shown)


Source/scrc.f90
```diff
@@ -387,7 +387,7 @@ MODULE SCRC
                       NSCARC_STACK_MAX               = 10            !> maximum number of consecutive solvers
 
 REAL(EB), PARAMETER:: NSCARC_THRESHOLD_CONVERGENCE   = 1.0E-15, &    !> threshold for convergence
```
rmcdermo: @SusanKilian All your (EB) parameters should be eight byte reals, for example, 1.0E-15_EB

rmcdermo: Also, what is your intention for 1.0E-16?  Is this machine precision?  If so, we use use TWO_EPSILON_EB, which is a global parameter.

SusanKilian:
Oh, sorry, typing error with the missing _EB. I use TWO_EPSILON_EB from now on.

rmcdermo:
OK, looks good.  Last nit pick (for next PR), you should not use >= when comparing with epsilon, just use >.

SusanKilian:
Okay, makes sense ... I will take it into account.

PR #6395, opened 2018-05-10
FDS Validation: Remove explicit reference to CFT in Cup Burner inputs

mcgratta:


drjfloyd:
Are we changing data.f90 to use defaults based on cup burner experiments or we will still be computing values based on a correlation?

mcgratta:
Yes, we're going to use the cup burner MEC with N2 agent as the basis for CFT. I will change the data.f90 file.

drjfloyd:
Here are the values for N2:

Acetone N2 1670.0
Acetylene N2 1312.0
Benzene N2 1682.7
Butane N2 1594.0
Dodecane N2 1577.0
Ethanol N2 1529.3
Ethylene N2 1441.0
Heptane N2 1616.5
Hexane N2 1646.0
Hydrogen N2 870.0
Methane N2 1662.7
Methanol N2 1395.3
Octane N2 1564.0
Propane N2 1572.7
Propanol N2 1608.0
Toluene N2 1754.5

mcgratta:
Do you have values for ethane, decane and propylene?

drjfloyd:
No I don't.

OCTANE (C8) and DODECANE (C12) both have N2 at 33 %.  Assuming the same
holds for DECANE (C10), we could compute the CFT assuming 33 % (probably
will wind up near the midpoint of 1564 and 1577).

ETHANE I've only found data so far for Halon-Alt agents
METHANE has an MEC of 24.2% and PROPANE has an MEC of 32.2 %. The other
hydrocarbons range from 29 to 33 %.

ETHYLENE has an MEC of 45 %.  PROPYLENE is probably close to that.

PR #6396, opened 2018-05-10
FDS Source: Add default CFT based on MEC

mcgratta:


mcgratta:
Jason, can you check these, and if you can find values for the commented fuels, add them. Also, what should we set for default?

PR #6403, opened 2018-05-11
FDS Tech Guide: Correct equation

mcgratta:


mcgratta:
Good catch. That mistake's been in the Appendix for quite some time.

PR #6419, opened 2018-05-15
FDS Source: fix error for non-Dirichlet case with evaluation of undefined level in scrc

SusanKilian:
I fixed a bug which is related to the application of the multilevel variants (multigrid or twolevel-cg) w.r.t. to a missing specification of the evaluation level for the Dirichlet-BC-counter.


rmcdermo:
Thanks, Susan!

PR #6441, opened 2018-05-23
Check on too small edge length and face area for GEOMs

emanuelegissi:
This are the first quality checks on GEOMs edges and faces.
I tested the checks with a broken Vettori sloped case.


marcosvanella:
@emanuelegissi have you run the GEOM verification tests on these commits?

marcosvanella:
Ema, GEOMEPS is recomputed taking into account the domain size (line 1903 in geom.f90). 
We should  recompute GEOMEPSSQ accordingly after GEOMEPS is set. Take a look.

PR #6454, opened 2018-05-30
FDS Source: Fix thermophoretic dT/dx in DNS mode

drjfloyd:


rmcdermo:
Ah... there's the factor of two.  Well done.

PR #6458, opened 2018-05-31
Added trace flag to qfds.sh for use with Intel Trace Analyzer

PatrickFURI:


mcgratta:
Thanks. Double check to make sure that nothing special happens when the trace flag is not set.

PatrickFURI:
I tested it with -v and it had the expected behavior both with and without
the flag.

PR #6472, opened 2018-06-05
FDS Source: revert changes to pres

rmcdermo:


rmcdermo:
@marcosvanella These variables show up as unused when compiled without MKL.  I think they need to be behind a directive.
```
 static ../../Source/pres.f90
../../Source/pres.f90(2278): remark #7712: This variable has not been used.   [IERR]
INTEGER :: IERR
-----------^
../../Source/pres.f90(2277): remark #7712: This variable has not been used.   [MB_FACTOR]
INTEGER, ALLOCATABLE, DIMENSION(:,:) :: MB_FACTOR
----------------------------------------^
```

marcosvanella:
Should be fixed in PR #6474.

PR #6475, opened 2018-06-05
FDS Verification: Restore dancing_eddies case

mcgratta:


mcgratta:
Patrick -- just an FYI, this dancing_eddies case is used for other verification purposes, so we don't want to change it. To test the Intel tools, just copy repository cases into new files and rename them by changing the CHID at the top of the input file.

PR #6487, opened 2018-06-07
New MISC parameter FIREBOT_NO_ERROR, reload bad geometry verification cases, matlab script

emanuelegissi:
New MISC parameter FIREBOT_NO_ERROR, when T, ERROR label is replaced with SUCCESS
Update UG MISC table with FIREBOT_NO_ERROR (commented out), and compiled for testing
Update bad GEOM verification cases with FIREBOT_NO_ERROR
Tested every case in debug mode
Reload these bad GEOM cases in Firebot
Added tentative matlab script for checking error conditions


rmcdermo:
Can we change FIREBOT_NO_ERROR to POSITIVE_ERROR_TEST?

emanuelegissi:
Yes, of course!

PR #6502, opened 2018-06-13
Revert pois.f90 changes

PatrickFURI:


mcgratta:
If you change FDS source code (.f90 file), make sure your commit message is
```
FDS Source: Blah blah blah.
```
We grep on the string `FDS Source` to look for source code bugs.

PR #6534, opened 2018-06-20
FDS User Guide: clarify text re back side boundary conditions

bmralph:
Clarification regarding what a "wall" means in this context.


bmralph:
Added a few words following confusion from a colleague regarding whether the "wall" was a `SURF` definition or a `OBST` definition.

mcgratta:
Thanks, K.

PR #6540, opened 2018-06-21
Matlab: IMPORTANT, switch back to TeX interpreter for *most* dataplot cases

rmcdermo:
In this PR, I converted dataplot back to using TeX as Font_Interpreter, but left open the possibility to use LaTeX for special cases, like Waterloo.  There is now a column in FDS_validation_dataplot_inputs.csv for Font_Interpreter.

This needed to be done for the incorrect Times font, but the reason it had to be done was that plots with LaTeX interpreter are about 43 KB and plots with TeX are about 3 KB.  This had ballooned the guides to over 250 MB, as pointed out by Salah.


sbenkorichi:
Nice work Randy,
You've brought the Validation guide alone from around 150MB to 40MB only.

PR #6601, opened 2018-07-06
FDS Source: Add detailed CFL output file option

mcgratta:


mcgratta:
This is a handy output option. Read more in the User's Guide that is committed here.

PR #6613, opened 2018-07-12
FDS Source: allow CHECK_MESH_ALIGNMENT without SETUP_ONLY

rmcdermo:
If you had set CHECK_MESH_ALIGNMENT=T on a MESH line, but you were finished with your setup and wanted to run the case, you had to go back and remove the CHECK_MESH_ALIGNMENT.  Now, you can leave it in case you need to go back and check again.  I do this frequently.


rmcdermo:
Hang on, I am fixing this!

PR #6619, opened 2018-07-12
FDS Source : Add READ_CATF routine to concatenate input files.

marcosvanella:
Randy, as discussed, I added the basis for this capability in read.f90. 
Check it out with these two files:

[sphere_helium_1mesh.fds.txt](https://github.com/firemodels/fds/files/2190419/sphere_helium_1mesh.fds.txt)
[sphere_helium_1mesh.geom.txt](https://github.com/firemodels/fds/files/2190420/sphere_helium_1mesh.geom.txt)


rmcdermo:
@marcosvanella Great job!  My only comment, and we can discuss this more tomorrow, is that we already have the concept of setting OVERWRITE=F on the MISC line, which users use to prevent overwriting the input file.  So, I would change the default behavior to go ahead and overwrite _cat.fds unless OVERWRITE=F.  I would not want to have to save or erase this file every time I make a change to the base input file.

marcosvanella:
Ok, I also still need to make this work on parallel jobs. That should be simple.

On Jul 12, 2018 6:51 PM, Randy McDermott <notifications@github.com> wrote:

@marcosvanella<https://github.com/marcosvanella> Great job! My only comment, and we can discuss this more tomorrow, is that we already have the concept of setting OVERWRITE=F on the MISC line, which users use to prevent overwriting the input file. So, I would change the default behavior to go ahead and overwrite _cat.fds unless OVERWRITE=F. I would not want to have to save or erase this file every time I make a change to the base input file.

PR #6626, opened 2018-07-14
FDS Verification: remove ^M from sphere_radiate files

rmcdermo:


mcgratta:
Ack, missed that. I run the Windows SMV which writes out those dos files. One other thing I overlooked -- the case sphere_radiate is automatically entered in the table of cases at the end of the verification guide, with a link to the section in either in the U or V guide that describes the case. Either we cut that link somehow or just add the case to the V guide in the radiation section with a header indicating that this is under development. I guess that this is the reason for the two lists of cases.

rmcdermo:
No problem.  I am also cleaning up my PRIMARY_BREAKUP mess.

Yes, I will just move the cases to the complex geometry list.  These now get run in firebot.  We'll move over after next maintenance release.

PR #6627, opened 2018-07-14
Matlab: remove sphere_radiate from dataplot for now

rmcdermo:
@marcosvanella 

Will you see about adding this case to the geom_notes?  I am not sure how this process works at the moment for cases that live in dataplot.

Also, I can see now that we have many Complex_Geometry cases in FDS_Cases.sh.  But we also maintain GEOM_Cases.sh.  Do we still need both?  I would rather have the normal firebot run all the geom cases.  How we post-process them is another issue, I understand.


marcosvanella:
I'll look into it. I don't think we have plots in geom_notes from dataplot (we have scripts that generate these after the call to dataplot).

About the GEOM_Cases.sh, the reason for those is being able to run the geometry cases (./Run_FDS_cases -g) and process the smokeview figures for them without needing to run the verification suite. In generating verification cases to test FDS, having this extra option is not critical as the work is done in a one by one basis. 

Glenn can chip in on this.

gforney:
Marcos said what I would have said - we want to be able to run cases
needed to build geom_notes without running the entire vv suite.  Other
solutions I could think of for solving this problem ( add if statements to
FDS_Cases.sh) would break firebot I think and make it more complicated.

rmcdermo:
It's fine for now to have GEOM_Cases.sh.  But very soon we need to be integrating everything into one version to prepare for FDS 7 beta.  In my view, it is time to cut the cord where ever possible.  This is one example.

PR #6645, opened 2018-07-17
FDS user guide: changes to MTR guidance

bmralph:
I've found the guidance on MTR to be confusing and am proposing the following changes. I think that the use of STATISTICS=MEAN is not required (we don't need spatially averaged velocities for the outputting of TKE) - instead we need time-averaged velocities. I've added some more description also regarding nomenclature and the post-production of velocity components to output TKE.


rmcdermo:
Thanks!

PR #6654, opened 2018-07-19
Update fds2ascii.f90

trgdang:
Resubmit pull request: multi blocks vtk converter.


rmcdermo:
I do not use fds2ascii much, and have never used it to look at PLOT3D.  Do you actually have to manually do through and write out the .vtk for each time step?  Or is there a way to automate this?  Would it not make more sense to write a Python script to convert the .q into .vtk?  Isn't that what they call a Paraview "reader"?  Thanks.

trgdang:
I actually have another version of (the idea is somehow similar to this version) which can automate the process, but I think it shall damage other tools integrated into fds2ascii. Of course you can view PLOT3D files by Tecplot (in this case we need both .xyz file and .q files if I remember well). Nevertheless, for those who don't want to install smokeview, and can't afford Tecplot, Paraview is their choice.

rmcdermo:
My problem with accepting this PR is that I can't perform a good test without (it seems) manually creating a time file for each time step, something like 500 files.  I do not see this as useful, which is why I am asking if I am just missing a trick in fds2ascii to make this practical.

Why not just write the .vtk files directly from FDS?

PR #6658, opened 2018-07-20
FDS user guide: added more words re INERT

bmralph:
I'm proposing some more words regarding INERT. Many times I've had discussions with practitioners in the UK and students who use INERT as the default with the justification "I'm not concerned about heat transfer at the walls, it's not important to my case". I have repeatedly found a misunderstanding of what INERT means by users. Therefore I've made things a bit more explicit. Feel free to edit, change, etc. my words.


rmcdermo:
Thanks, Ben.  Good clarification.

PR #6661, opened 2018-07-21
Manual: update activation energy unit.

sbenkorichi:
Correct the Ea as mentioned in the reference.
From` 211,000~kJ/mol` to `211KJ/mol`

### DIFF
--- a/Manuals/FDS_User_Guide/FDS_User_Guide.tex
+++ b/Manuals/FDS_User_Guide/FDS_User_Guide.tex
@@ -5850,7 +5850,7 @@ Note: The sign of {\ct NU} corresponds to whether that species is consumed (-) o
 
 \subsection{Soot Surface Oxidation}
 
-If soot deposition is being computed, the soot present on surfaces can undergo oxidation if the local temperature is high enough and oxygen is present. Soot oxidation is enabled with the logical flag {\ct SOOT\_OXIDATION} on the {\ct MISC} input. Oxidation is computed as a single step, Arrhenius reaction using the soot surface density and the local oxygen mole fraction. The reaction uses a rate constant of 4.7x10$^{12}$ and an activation energy of 211,000~kJ/mol. These values are based upon work performed by Hartman, Beyler, Riahi, and Beyler ~\cite{Hartman_2012}.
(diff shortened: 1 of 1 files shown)


sbenkorichi:
I do have a request if it's trivial thing to accomplish without having adverse effect on other works.
Can we make the default units of `Ea` to `KJ/mol`, instead of `J/mol`. Most of the literature references are using such unit, and we saw users getting this wrong time to time. So, just to be in alignment with other practices.
Thanks !

rmcdermo:
@drjfloyd Can you please review this change?

drjfloyd:
Thanks for catching the error.  

In terms of your unit request, the units for E are already kJ/mol; see table 21.24 in the User's Guide.

sbenkorichi:
Be patient with me here,
But I doubt it. I believe it's set for both gas and solid phase in `J/mol`. 
As Kevin told me before, it's best to check the code when it comes to this. I think table 21.24 needs updating.
But, what confuses me is this line 
` 5668  ML%E(:)                 = 1000._EB*E(:)` from read file. It's already multiplied here.

The one that I'm sure about is the solid phase it's in `J/mol`.
See section "Single step reaction" **16.3**. It's in `J/mol`, Ea is defined on the `&MALT` line.
And the in the input file examples of pyrolysis, we do have it in `J/mol`. --Look for instance to the example` birch_tga_1step_2.fds` 	the activation energy is in J/mol `E = 1.0E5` <--- this to me it's in `J/mol`
Also see section **11.5.2** -- (the activation energy, is prescribed via E(j) in units of` kJ/kmol`)
I've done before some tests and it's taking values in `J/mol`, so one will always need to multiply by 1000 to input it into fds, which quite confusing for some users.

Any thoughts on this?
Thanks !

sbenkorichi:
Right, further to what I said above, the value that's dumped it's gets divided by 1000., which will returns unit to KJ/Kmol.
`2780      WRITE(LU_OUTPUT,'(A,ES9.2)')'        E (kJ/kmol): ',ML%E(NR)/1000.`
So, can we change the units to` KJ/mol` instead.
Should I open a new feature request for this or send a Pull request with the necessary changes?
Thanks !

drjfloyd:
I'll take a closer look at this.

drjfloyd:
My first reply was in error (misread the unit in the guide).  Currently the E units for REAC and MATL are kJ/kmol.  They have been that way for a very long time.  I'm not sure we would want to change that as we would suddenly break everyone's old inputs.  Since I misread it (I first read it is kJ/mol and missed the second "k"), one clarifying change could be to call it J/mol rather than kJ/kmol in the guide.

rmcdermo:
I am used to seeing J/mol in the kinetics literature.  That would be my vote.

drjfloyd:
I'll change the guide and the units in source comments.

drjfloyd:
https://github.com/firemodels/fds/commit/2dca463dbae55fb1a8cd059290aa2e0547159dcb

sbenkorichi:
https://github.com/firemodels/fds/pull/6692

PR #6692, opened 2018-07-26
FDS Validation: Change kJ/kmol to J/mol

sbenkorichi:
More changes to #6661


drjfloyd:
In a couple of cases you changed kJ/mol to J/mol without changing the numerical value. Did you first verify using the source literature that the original unit was in error?

sbenkorichi:
Actually those were an original error, so I just found about them while doing this PR. 
I did check it, that was Stas's work paper, he was reporting J/mol.

drjfloyd:
Thanks for checking.

Issue #6714, opened 2018-07-30
Correct cross correlation of u and T in Waterloo Methanol


rmcdermo:
![screen shot 2018-07-30 at 11 20 41 am](https://user-images.githubusercontent.com/4418497/43406604-9fb2b4c2-93ea-11e8-99b8-d293b719b2d5.png)

@mcgratta I think the difference between the data and cross correlation with u*T here (left side of r=0) is just a matter of taking an absolute value.

drjfloyd:
TIME_AVERAGED = FALSE writes the instantaneous value of the output, i.e. it is not averaged over DT_DEVC.  ABSOLUTE_VALUE writes the absolute value of the output, i.e. if the DEVC output were -1 it would be written as 1.

Er9y714:
Sorry, I realized that just after posting question.

mcgratta:
Have a look at the latest Validation Guide. I replaced the horizontal velocity u with a `'RADIAL VELOCITY'` such that a positive value points away from the origin.

PR #6715, opened 2018-07-30
FDS Source: Issue #6714. Add ABSOLUTE_VALUE for DEVC.

mcgratta:


drjfloyd:
Would this be better as math control function to make it more broadly accessible?

mcgratta:
It could be. But it's easy just to add `ABSOLUTE_VALUE=.TRUE.` on the `DEVC` line rather than extra `CTRL` and `DEVC` lines.

drjfloyd:
Should the ABS be done after any CONVERSION?

mcgratta:
I can't think of a good example.

PR #6729, opened 2018-08-01
MATLAB to Python Results

PatrickFURI:
Scripts and README's for converting MATLAB scripts to Python scripts.


mcgratta:
Patrick, there is an existing directory called `Utilities/Python`. Move the existing scripts in this directory to a new sub-directory called scripts, and then move all other python scripts there so that we end up with the `Python` folder having the same basic structure as the Matlab folder. In other words, leave Matlab as is, and just create its clone in Python.

PR #6821, opened 2018-09-12
 FDS Source : explicitly define various quantities. 

marcosvanella:


drjfloyd (commented):
Why are we computing a temperature dependent specific heat when the user has defined that cp is constant?

rmcdermo:
Good catch Jason.  Marcos is working on a fix.

marcosvanella:
Jason, correct. We changed that to use CPOPR on the KP equation when constant CP is selected. In such case the CP pointer hasn't been assigned.
Check the open PR #6822.

PR #6822, opened 2018-09-12
FDS Source : Remove computation of CP when CONSTANT_SPECIFIC_HEAT_RATIO=T, use CPOPR.

marcosvanella:


rmcdermo:
This should work.  I'll test.

PR #6882, opened 2018-10-12
FDS source: added global substepping to HVAC_MASS_TRANSPORT

bmralph:
Refer to issue in issue tracer. I added more information there so that if my solution is not adopted the issue can be tracked over there.

This addresses the potential for the breaking of a CFL limit on the global HVAC solution and errors in mass conservation.


drjfloyd:
In line 931 you call the solver with DT_MT but it doesn't appear to get initialized anywhere.  I think you need to set to DT?

bmralph:
It needs to be DT_MT as set in UPDATE_HVAC to make sure we're not advancing
too far. I've made DT_MT module wide (i think... not at desk just now)

drjfloyd:
Line 856: SUBROUTINE HVAC_CALC(T,DT,FIRST_PASS)

The calculation routine is called with DT.  

Line 913: IF (HVAC_MASS_TRANSPORT) CALL UPDATE_HVAC_MASS_TRANSPORT(DT_MT)

DT_MT has not been defined yet.

bmralph:
I define DT_MT in the HVAC_ROUTINES module on line 18; meaning it's
available across subroutines within the module

The call to HVAC_CALC on line 856 populates DT_MT which is then useable for
the call on line 913 to UPDATE_HVAC_MASS_TRANSPORT

To be honest I didn't know if this was best practice but I saw it done with
DT_HV so adopted this module-level definition method

drjfloyd:
DT_MT is defined as a REAL(EB) on line 18. At this point it has no value assigned to it.  The first time it is referenced is when it is used to call UPDATE_HVAC_MASS_TRANSPORT.  Since it had no initial value assigned to it, it still has no value assigned to it. If you look at DT_HV you will see that it is assigned the value DT at the start of HVAC_CALC.  You need to do something similar with DT_MT.

bmralph:
Okay cool. Thanks. I thought the population within hvac_update was enough.
I'll add a dt_mt = dt along with dt_hv = dt at the beginning of hvac_calc

Thanks Jason

bmralph:
Definition is now added. I realise I should have sent this PR from a new branch of my origin... things got _sliiiightly_ messy when I was trying to add a new commit via git for Windows. In the end I just added it directly on GitHub which worked well.

PR #6884, opened 2018-10-12
An Error in FDS_EVAC User Guide

godisreal:
I tested an example and it shew that herding agents cannot follow others before their detection and reaction times are reached.  So I suggest that the sentence in FDS_EVAC User Guide should be deleted or modified because it is not consistent with the testing result.  

The discussion is listed in the google forum as below.  
https://groups.google.com/forum/?fromgroups#!topic/fds-smv/qNAbpZ8Y2-A

Best Regards

### DIFF
--- a/Manuals/FDS_Evacuation_Guide/FDS+EVAC_Guide.tex
+++ b/Manuals/FDS_Evacuation_Guide/FDS+EVAC_Guide.tex
@@ -1897,9 +1897,7 @@ are heading towards an exit, they start to follow these other agents.
 A herding agent will remain still even after detecting the fire and
 its reaction time has passed if it has not been able to select any
 exit either by being familiar with it or by observing other agents.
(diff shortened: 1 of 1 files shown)


rmcdermo:
@tkorhon1 Please check this and accept or close.  Thanks.

godisreal:
@tkorhon1
Dear Timo, If you do not mind, please let me know your workflow so that I can follow your schedule to contribute something there.  Many Thanks.  Peng WANG.

tkorhon1:
Hi Peng!

I have been a little bit busy, but now I added your change to the FDS+Evac manual LaTex file that I have in my own computer. I will try to update the Github .tex file and add the V&V .fds files to the Github also and my shell scripts that I use to run the V&V cases. The figures are mainly done using Excel, so most of the figures wll be missing in Github.

TimoK

PR #6893, opened 2018-10-15
FDS Source: Fix POINTER inconsistency to be compatible with Intel 19

mcgratta:


mcgratta:
I ran the verification suite to check this. Firebot will test tonight.

PR #6900, opened 2018-10-16
FDS source: add new device for descritised ducts using hvac_mass_transport

bmralph:
Creating first of devices to output cell quantities within descritised ducts when using `HVAC_MASS_TRANSPORT = .TRUE.`.

Only added one device at the minute. To get the go ahead regarding correct set up before I add devices for density, mass fraction.

Some notes:
- I used `CELL_N` for the user input for which cell number was of interest. Considered using `DUCT_CELL_N` but seemed long.
- Moved allocation of duct cell arrays within hvac.f90 as they were needed to initialise devices.
- Added two new arguments to `GET_QUANTITY_INDEX`. Added dummy arguments wherever the subroutine is called.
- I caught all user input errors I could imagine (e.g. missing `DUCT_ID`, `CELL_N` or `CELL_N` outside of the duct).
- Compiled in debug with no errors.
- I haven't alphabetised things yet but will do if/when I know the method is in line with the project.


drjfloyd:
Ben, thanks for adding the outputs. I think we would want to be able to use length along the duct either instead of or in addition to the node index. If we get to having heat transfer into a duct from the domain, then I could see a user with a duct going through multiple rooms just wanting to give the length along the duct.

bmralph:
Jason. Great point. Also makes things easier if the user is changing the
number of cells. Saves them changing the DEVC line. Cheers, I'll make that
change when i add the other output quantities.

PR #6934, opened 2018-10-29
FDS Source: MAJOR CHANGE: Allow DEViCes to cross mesh boundaries.

mcgratta:


mcgratta:
This PR allows integrated devices to extend across mesh boundaries, but for now it only handles STATs like `VOLUME INTEGRAL`. I plan to extend it to work for all integrated outputs. Try it and let me know if it works. I ran it through the verification cases. 

Jason -- if you have a complicated DEVC/CTRL case, try it.

Er9y714:
Thank you, I have been hoping and waiting for this.

mcgratta:
Feel free to test it, but just be aware that for now it only works for `VOLUME INTEGRAL` and `SURFACE INTEGRAL`.

PR #6938, opened 2018-11-01
FDS+Evac dump.f90: remove unused variables

tkorhon1:
The "device" activation writes to .smw was moved from dumpf90 to main.f90
yesterday. So, there were now some unused variables in dump.f90.


mcgratta:
Timo -- do not make any more changes to dump.f90 until I am finished overhauling devices. Thanks.

PR #6940, opened 2018-11-02
FDS User Guide: Describe changes to DEViCes

mcgratta:


mcgratta:
Some notes on changes to devices. All spatially integrated devices should span multiple meshes now -- means, max, min, integrals. The changes passed the verification tests, but I have not extensively tested multiple meshes yet.

`STATISTICS` is replaced by `SPATIAL_STATISTIC` and `TEMPORAL_STATISTIC`

`STATISTICS_START` and `STATISITICS_END` play the roles of `DT_DEVC_LINE` etc. You can specify integration time intervals on each `DEVC` line.

All deprecated inputs still work, but have been removed from the Guide.

PR #6943, opened 2018-11-02
FDS User Guide: Edit to beam_detector example

mcgratta:


mcgratta:
There is only one type of spatially integrated device that I cannot process with the current restructuring of devices. That is `LAYER HEIGHT` and layer temps. These would require two MPI exchanges, one to add integrals of T and 1/T, and the other to integrate T from z_i to H through potentially more than one mesh. I have to think about this one.

PR #6946, opened 2018-11-04
FDS Source: add WALL PRESSURE TEST

rmcdermo:
Here a new SOLID_PHASE_OUTPUT QUANTITY called 'WALL PRESSURE TEST' is added to compare with original wall pressure.  The idea is that KRES is, by no slip, zero at a wall.  In DNS calculations of drag coefficient, we are finding better results by interpolating H to the wall without KRES and interpreting p = rho*H at the wall as the pressure on the wall.


rmcdermo:
@marcosvanella Please double check that this should work for CFACES.  Thanks.

marcosvanella:
@rmcdermo change the PRESS assignment line in GET_PRES_CFACE_TEST (line ~1567 in geom.f90) for:

      PRESS = PRESS + CUT_FACE(IND1)%INT_COEF(INPE)* &
      CUT_FACE(IND1)%INT_CVARS( INT_RHO_IND,INPE)*CUT_FACE(IND1)%INT_CVARS( INT_H_IND,INPE)

This will provide the pressure at the wall using rho_g and H at the external normal point for the CFACE. This is consistent with ONE_D%RHO_G*H(IIG,JJG,KKG) done on wall cells. 
I did the change on my repo, but have other changes, so about to firebot test now (blaze). 

Also, to engage the new 3D interpolation of velocities for forcing set IBM_PLANE_INTERPOLATION=F in the &MISC line. This will become the default eventually.

rmcdermo:
OK, thanks.

PR #6971, opened 2018-11-12
FDS source: add new HVAC fan model (beta)

bmralph:
Adds a new quadratic fan model to HVAC (beta). It is currently hidden behind a `QFAN_BETA` logical flag on the `HVAC` line and doesn't affect normal workings.

The new fan was developed to address problems with the current quadratic fan related to "background" flow conditions within the duct (e.g. fire-induced flows, pressure induced flows).

The fan model is based upon the generation of a steady state system curve for the relevant ductrun for that time step. This system curve is equated to the fan curve to output an operation point.

The fan model is behaving well for tested verification cases with no `ZONE`s and agrees with the existing model for such cases. The following points related to this beta version:
- doesn't agree with existing quadratic model for cases with `ZONE`s (e.g. fan_test.fds)
- `DUCT` in which the fan is placed requires a loss (either `LOSS` or `ROUGHNESS`)
- currently uses a slow bisect method to equate system and fan curve; faster methods e.g. Newton weren't converging

Compiled in win db with no errors or warnings.


rmcdermo:
Ben, thanks.  Let's have Jason review.  Do you have a verification case?  Or does this improve one that already exists?  If so, please shoot us an email with the plots generated with this feature.  Thanks!

bmralph:
Hi Randy. No verification cases yet per se. Work in progress. I'm still working on this. Want to get it to work with `ZONE`s and I can't yet figure out why it can't... when it's working it will (at least) use the original fan verification cases (fan_test and ashrae). I will make another verification case to show the accounting for "background" flow.

bmralph:
@drjfloyd two things to bear in mind. The new fan model is agreeing with the existing quadratic for cases where the ducts are suitably rough (the solution of the system curve does not exist with `TOTAL_LOSS = 0`) and where there are no ducts with `REVERSE=.TRUE.` (the operation point method is dependent on direction, the steady state solution is direction-dependent due to the accounting of the pressure at up and downstream nodes).

rmcdermo:
Ben, That's fine.  In that case, I will let Jason merge when he is ready.  Thanks!

drjfloyd:
I'll take a look at this sometime today.

drjfloyd:
I'd like to see this hidden better while we test it. Right now we are always doing the new algorithmc even if there is no new fan. Also if the flag is set, all fans become the new fan even if they don't have inputs to support it.  I'd add a flag to the HVAC input for the new fan type and in the logic for selecting the fan type only set the fan type to 4 if it is already going to a fan type 2 AND the flag is set so we don't force types 1 and 3 to become type 4.  If a type 4 is defined then automatically set the global flag. Only allocate variables and execute lines of code for the new fan if the global flag is set (wrap everything in IF blocks)

bmralph:
@drjfloyd take a look at 2e5de6a. Hopefully addressed your comment. Don't _fully_ understand what you mean re make all fans become type 4. But hopefully the new selection addresses your comment. Cheers :+1:

bmralph:
Just realised that taking `DU%TOTAL_LOSS` as the loss for the output of the points of the system curve is not correct if there is a roughness and f is calculated. f depends on the velocity i.e. for the two points on the system curve: the two system velocities and not the actual velocity in the duct.

PR #6998, opened 2018-11-20
 FDS Source : Add SOLID_PHASE point devices for CFACES.

marcosvanella:


marcosvanella:
@mcgratta give it a try.

PR #7018, opened 2018-12-01
FDS+EVAC_Guide: Update FDS+EVAC_Guide.tex

godisreal:


mcgratta:
Thanks

tkorhon1:
Hi Kevin!

I made a GitHub project for the FDS+Evac Guide in my GitHub (tkorhon1):

https://github.com/tkorhon1/FDS-Evac-Guide

So, I will put there my V&V cases, LaTeX source, figures, etc. I thought that it might be
easier to make a separate Git repo for the FDS+Evac manual. This repo is going to be
quite much like "a work in progress" type one. There will not be automated scripts
that will produce figures from the V&V cases and similar stuff.

So, one could take the .tex file of the FDS+Evac manual away from the firemodels/fds/manuals
part and write there a readme file (or something like that) saying that the manual is
in an another GitHub place.

When I have some time, I try to update (make new ones) the "verification cases" in the
Verification/Evacuation directory. I will put there simple (and fast to run) test cases that
the evacuation stuff is running, which means that one gets "STOP: FDS completed successfully"
message, i.e., the calculation does not crash. I do not put there "real" verification cases, where
one should check that the evacuation things work as they should.

Wbr,
Timo

mcgratta:
Would you like to put the repo under firemodels, like fds, smv, out, etc? Or, Randy, do you want to limit the number of repos under firemodels? Maybe work the repo under your own GitHub account and at some point we can migrate it to firemodels.

godisreal:
Hi, Timo.  Thanks for creating the repo.  I do not have access to the VTT webpage.  So your new repo in Github offers an alternative way to reach your VTT examples of FDS+EVAC.  

I also agree with Kevin's suggestion.  There should be a proper way to merge new revisions into firemodel.  So it will be better if your VTT files are also uploaded to firemodel/fds. 
Best Regards
Peng WANG

tkorhon1:
Well, we could put the repo under firemodels in future. But for now, I have figures as binary files (most of them are pdf) and no instructions/script to make the figures automatically after the FDS+Evac runs.
I do have not anything against that FDS+Evac manual git is put in the Firemodels. 

So, I should do (some of them at least):

1) make .sh (well, I use bash shell scripts in our Linux cluster) script files that run the example and V&V cases (I have already scripts, but should edit them to be nice and easy to read etc.)

2) Put the .fds files where they should be (well, already in /tkorhon1/FDS-EVAC-Guide).

3) modify the .sh scripts that I use to make (some) post processing of the results.

4) Make the plotting of the figures automatic (e.g. gnuplot or Matlab). Now I use Excel, where
I have to copy the .csv file (and sometimes _evac.out file) information.

Points 1-3 should be done before the fds+evac manual repo could be put under Firemodels. Well, it could be put there earlier, but it would be very ugly (= no use to anybody except me). 

Well, it could be a good strategy to take the Firemodels/fds/Manuals/FDS_Evacuation_Guide directory away at some point (e.g., when there is a separate Firemodels/fds-evac-guide project or tkorhon1/fds-evac-guide is not so much "under construction"). And Firemodels/fds/Verification/Evacuation should be updated (by me). I should put there test cases to see, if the evacuation part is running or crashing. So, just to check if one gets "FDS completed successfully" => firebot runs easy to check. The actual verification cases are in the fds-evac-guide git part. Here I should usually see results in Smokeview to see, if the agents are doing things what they should do (and things show up in Smokeview as they should).

Wbr, Timo

rmcdermo:
@tkorhon1 I do not have any problem with adding Evac to firemodels organization.  The way we handle binary or other larger files is to store them in the fig or cad repos.  The idea here is that we are not going to worry about version control of these files.  If the repo ever gets too big, we will purge it and start over.  But the files are still publicly available.

If you want to be the only person with owner permissions on your repo, then keep it in tkhornon1.  Since firemodels is a NIST-blessed org, we have to maintain control of all the repos in that org.

tkorhon1:
Well, I do not need to be the only owner of evac manual repo. So, we could put it in firemodels. 
So, "firemodels/fig/evac" could be the place to put the pdf/png figures of the FDS+Evac manual.
And "firemodels/out/Evacuation" and some subdirectories, where I put the output files that
are needed to make the figures of the V&V cases.

firemodels/fds/Verification/Evacuation/  I put here just some inputs that check if the cases crash or not

firemodels/FDS-Evac-Guide  I put the manual LaTeX here etc (pdf/png figs are in fm/fig/evac). And I put here the example and V&V input files that are in the manual (and some shell scripts, .ssf .ini etc files).

mcgratta:
OK. We should be able to compile the FDS-Evac_Guide so long as we have updated all repos. We would not need to run any cases, unless those  cases are part of the standard FDS verification suite.

PR #7055, opened 2018-12-16
FDS+EVAC_Guide: Update Two Paragraphs

godisreal:
I suppose that evacuees are told to use emergency exits in fire drills so that they become familiar with these special exits for emergency use.  Please double check.  @tkorhon1


rmcdermo:
@tkorhon1 I thought you were setting up a separate repo for the Evac guide?

rmcdermo:
Feel free to reopen if that is not the case.  But I don't want a PR hanging around otherwise.  Thanks.

godisreal:
Is there a new repo yet?

tkorhon1:
Well, yes. I have a https://github.com/tkorhon1/FDS-Evac-Guide for the LaTeX source and pdf/png figures etc. But might be that I (and Peng) are not yet always remembering to use it. 

The idea (for now) is that we (Peng and I) use  https://github.com/tkorhon1/FDS-Evac-Guide to make updates to the FDS+Evac manual. And I, sometimes, update the LaTeX source in firemodels/fds/Manuals. So there should not be pending pull request then. I will commit new LaTeX source, merge it and confirm it.

At some point, there will be in firemodels/fig a subdirectory for evacuation manual and also in firemodels/out. And, perhaps, firemodels/fds-evac-guide, where are LaTeX source(s), V&V fds input files needed for manual, etc. So, then the firemodels/fds/Manuals/FDS_Evacuation_Guide directory could be removed. I.e, at some point we move tkorhon1/FDS_Evac-Guide => firemodels/.

When this move is going to be? Well, I first need to have some time to make the tkorhon1/FDS-Evac-Guide almost "ready" (= has the things needed to make the pdf from the LaTeX source and figs and has the V&V input files that are needed by the manual).

tkorhon1:
> 
> I suppose that evacuees are told to use emergency exits in fire drills so that they become familiar with these special exits for emergency use. Please double check. @tkorhon1

Yes, I read your changes. They are ok. In some parts of the world the evacuation drills are very common => emergency exit doors are known. In some parts of the world (e.g. Finland) evacuation drills are not common => emergency exit doors are not known and we have seen (we have been video recorded some drills some years ago) that the emergency exit doors are not used if there is an available "known, every day used door". Well, in schools drills are more frequent, but there also the doors used are usually known doors. The teachers guide the class rooms out in an orderly manner. I.e., the pupils go out and in like they do every hour: pupils spend their breaks (15 min break, 45 min lesson) at the school yard. This is very fast and effective. They have trained it in every day life for years and some 200 days per year and some 5-7 times per day.  So, in schools. almost all doors are used in everyday life.

PR #7078, opened 2018-12-21
Doc: Updated particle output options.

marcfehling:
According to the following lines of code, there are some more output options for particle quantities available than the FDS User Guide describes. This pull requests completes the documentation by the remaining options.

https://github.com/firemodels/fds/blob/78a399d9b56d802b6de4e0d9c931e89d03e2f814/Source/dump.f90#L7814

Follow-up to #6022.

### DIFF
--- a/Manuals/FDS_User_Guide/FDS_User_Guide.tex
+++ b/Manuals/FDS_User_Guide/FDS_User_Guide.tex
@@ -7879,6 +7879,9 @@ The parameter {\ct QUANTITIES} on the {\ct PART} line is an array of character s
 {\ct 'PARTICLE MASS'} (kg) \\
 {\ct 'PARTICLE AGE'} (s) \\
 {\ct 'PARTICLE WEIGHTING FACTOR'} \\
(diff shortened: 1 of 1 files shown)


rmcdermo:
Thanks, Marc!

marcfehling:
I just checked if FDS complains when I use these additional quantities. It didn't, which is why I submitted this pull request. However, the values of the quantities do not show up in `smokeview`, i.e. have a default value -1 x 10^20.

I will check the content of the `prt5` file. Maybe `smokeview` can not interpret these quantities yet. I'll keep you informed on that.

marcfehling:
The output to the `prt5` files happens here:
https://github.com/firemodels/fds/blob/78a399d9b56d802b6de4e0d9c931e89d03e2f814/Source/dump.f90#L3621
The quantities regarding position and velocity that I added to the documentation in this pull request are thus not available in those files.

I will match the quantities of the `DUMP_PART` routine to the ones of the `PARTICLE_OUTPUT` function in a separate pull request. Sorry for any inconvenience.

PR #7094, opened 2018-12-24
FDS Verification: fix input files in terminal velocity cases

rmcdermo:


marcfehling:
I should have checked if any tests were effected by request #7081. Sorry about that :(

Issue #7103, opened 2019-01-07
RAMP: Improve error message


sbenkorichi:
Hi Kevin,

Have a look at this case. 
```
&HEAD CHID='test' /

&MESH IJK=3,3,4, XB=-0.15,0.15,-0.15,0.15,0.0,0.4 /

&TIME T_END=20. /

&REAC SOOT_YIELD=0.01,FUEL='PROPANE'/

&SURF ID='TEST1'
	  COLOR='RED'
      HRRPUA = 1000.
      RAMP_Q = 'HRR1'
      THICKNESS=0.01 . /

&SURF ID='TEST2'
	  COLOR='BLUE'
      HRRPUA = 1000.
      RAMP_Q = 'HRR2'
      THICKNESS=0.01 . /

&RAMP ID = 'HRR1', T = 0,  F = 0. /
&RAMP ID = 'HRR1', T = 1,  F = 1. /
&RAMP ID = 'HRR1', T = 10, F = 1. /
&RAMP ID = 'HRR1', T = 11, F = 0. /

&RAMP ID = 'HRR2', T = 0,  F = 0. /
&RAMP ID = 'HRR2', T = 1,  F = 1. /
&RAMP ID = 'HRR2', T = 10, F = 1,2 /
&RAMP ID = 'HRR2', T = 11, F = 0. /

&VENT XB = -0.10,-0.05,-0.05,0.05,0.0,0.0, SURF_ID = 'TEST1' /
&VENT XB =  0.05,0.10,-0.05,0.05,0.0,0.0, SURF_ID = 'TEST2' /

&TAIL /
```
This is the error message that gives : 
`ERROR: Problem with RAMP HRR1 (CHID: test)`
The error is with 2nd RAMP but it's outputing the 1st one. 
Put the 2nd SURF before the 1st SURF in the input file, it will output the 2nd RAMP name.

I tried putting a simple write message and this is what it gave.

`too many values for NAMELIST variable, unit 5, file /home/salah/1_issues/ramp/test.fds, line 29, position 35`

Something is not right with the RAMP_ID(N).
```
SEARCH_LOOP: DO
         CALL CHECKREAD('RAMP',LU_INPUT,IOS)
         IF (IOS==1) EXIT SEARCH_LOOP
         READ(LU_INPUT,NML=RAMP,ERR=56,IOSTAT=IOS)
         IF (ID/=RAMP_ID(N)) CYCLE SEARCH_LOOP
[...]

PR #7104, opened 2019-01-07
FDS Source: Issue #7103. Fix bugs related to ERROR statements

mcgratta:


mcgratta:
In this commit, I added an optional argument to the SHUTDOWN subroutine. Now, by default, if you
```
CALL SHUTDOWN(MESSAGE)
```
the `MESSAGE` is written only once to the .err file by MPI process 0. However, if you do this
```
CALL SHUTDOWN(MESSAGE,PROCESS_0_ONLY=.FALSE.)
```
then the `MESSAGE` will be written by all processes. Why do this? In some cases, only some of the processes might get tripped by the error, in which case we need to have that particular process write the error message, along with the process ID so we know what process failed.

Also, when using `SHUTDOWN` and then a `RETURN`, make sure that no MPI calls are made after it in the same subroutine, or else the processes that have returned will cause the MPI call to hang. I found an MPI call that I had added at the end of `READ_DATA` that would hang up if any processes returned to the main routine prematurely.

PR #7108, opened 2019-01-12
FDS Source: Change UW to U_NORMAL

mcgratta:


mcgratta:
Cosmetic change only, except that UW0 was moved from WALL and CFACE to WALL%ONE_D and CFACE%ONE_D to be consistent with UW, UWS. Now these are ONE_D%U_NORMAL, U_NORMAL_S, U_NORMAL_0. Same functionality as before.

PR #7128, opened 2019-01-29
FDS Validation: remove BNDE from Lattimer cases

rmcdermo:


rmcdermo:
These cases also need to be updated for new GEOM constructs.

PR #7138, opened 2019-01-31
FDS Source: Skip input lines with no space after namelist variable

mcgratta:


drjfloyd:
Should this throw an error instead?

I think if someone typed "&OBSTXB=" they meant for that line to be read and just made a typo.

mcgratta:
I'll give it a try. The problem with ERRORs is that all the MPI processes must return to main and shutdown properly. I want to avoid adding lots of RETURNs or what have you.

drjfloyd:
Could there just be a check after each time we call CHECKREAD in read.f90 to see if STOP_STATUS was set and if so RETURN back to the top of read.f90. Then the normal stop status logic would kick in.

PR #7176, opened 2019-02-18
FDS Source: Add EVAP_MODEL plus changes to humidity calc

drjfloyd:


rmcdermo:
@drjfloyd Could you give some context for these changes.  Thanks.

Source/part.f90
```diff
@@ -2322,22 +2346,87 @@ SUBROUTINE PARTICLE_MASS_ENERGY_TRANSFER(T,DT,NM)
...
                ELSE SOLID_OR_GAS_PHASE_2
-
-                  NUSSELT  = 2._EB + NU_FAC_GAS*SQRT( LP%RE )
-                  SHERWOOD = 2._EB + SH_FAC_GAS*SQRT( LP%RE ) ! Sh matches Li&Chow FT 2008
-                  H_HEAT   = NUSSELT *K_AIR/(2._EB*R_DROP)
-                  H_MASS   = SHERWOOD*D_AIR/(2._EB*R_DROP)
+                  LENGTH = 2._EB*R_DROP
+                  RE_L = RHO_AIR*VEL*LENGTH/MU_AIR
+                  SELECT CASE (EVAP_MODEL)
+                     CASE(-1) ! Ranz Marshall
```
rmcdermo, line 2365: Perhaps a reference and equation number?

Source/part.f90
```diff
@@ -2322,22 +2346,87 @@ SUBROUTINE PARTICLE_MASS_ENERGY_TRANSFER(T,DT,NM)
...
+                           ZZ_GET(Z_INDEX) = 1._EB
+                           CALL GET_SPECIFIC_HEAT(ZZ_AIR,C_GAS_DROP,TMP_FILM)
+                           ZZ_GET(1:N_TRACKED_SPECIES) = ZZ_INTERIM(II,JJ,KK,1:N_TRACKED_SPECIES)
+                           CALL GET_SPECIFIC_HEAT(ZZ_AIR,C_GAS_AIR,TMP_FILM)
+                           THETA = C_GAS_DROP/C_GAS_AIR/LEWIS
+                           B_NUMBER = (1._EB+B_NUMBER)**THETA-1._EB
+                           NUSSELT  = 2._EB + NU_FAC_GAS*SQRT(RE_L)*LOG(1._EB+B_NUMBER)/B_NUMBER
+                           H_HEAT   = NUSSELT*K_AIR/LENGTH
+                        ENDIF
+                     CASE(2) ! Shazin M1
```
rmcdermo, line 2404: Shazin M2?

drjfloyd: At this point the EVAP_MODEL case structure may not remain for a future release. This commit mostly gives me the ability to test other models. If it makes sense after testing to change the default or to have multiple. I'll add them to the tech ref. If you want I can add the paper citation to the current source for M1 and M2. The other thing contained in this commit is a change to how we do the C-C relation for the equilibrium vapor fraction.  The prior formula is a derivation that assumes H_V is constant with temperature; it isn't. The change pushed up is a better approximation that I have derived.

rmcdermo: OK, sounds good.  In general, I like to have a short ref next to formulae, for example, Incropera and Dewitt, Eq. X.  This just serves as a reminder without having to cross check Tech guide when looking at the code.  Thanks.

PR #7187, opened 2019-02-20
Manual typos

imnasnainaec:
Fixed two errors in the Manuals that had caused me to look for things that don't exist.


drjfloyd:
The CRAYFISKPAK is a typo that should be fixed. Thanks for finding it. I'll do that.

The change from Examples to Verification is not correct. This is referring users to the standard installation of FDS using the installer which places the files in a folder called examples.  For example in windows this is in C;\Program Files\firemodels\fds6\Examples. 

I going to close this request and make the change for CRAYFISHPAK

imnasnainaec:
I'm glad at least one of the two was useful. Thank you for the clarification of the other, @drjfloyd .

PR #7236, opened 2019-03-08
FDS User Guide: Describe CFL_FILE

mcgratta:


mcgratta:
FYI, I added some extra diagnostics to the CFL_FILE. This is helpful in debugging big, ornery cases.

PR #7238, opened 2019-03-09
FDS Source: Correct precision issue related to Poisson solver.

mcgratta:


mcgratta:
Hopefully this commit will fix the Poisson initialization error. It's just a precision thing -- 10e-11 vs 10e-16 in cases with periodic boundary conditions.

PR #7320, opened 2019-05-03
Updated particle output quantities in user guide.

marcfehling:
The section `20.12 Summary of Frequently-Used Output QuantitiesSummary of Frequently-Used Output Quantities` was lacking some particle output quantities that this PR provides.

Further, I sorted the particle quantities in section `20.9.4 Coloring Particles and Droplets in Smokeview` alphabetically, to be in accordance with the register.


rmcdermo:
Thanks, Marc!

PR #7339, opened 2019-05-15
FDS Verification: Explicitly specify SHAPE_FACTOR for particle absorp…

mcgratta:
…tion test


mcgratta:
Note to all --- it is no longer assumed that the SHAPE_FACTOR (ratio of projected to surface area) for solid particles is 0.25, as it is for spheres. Plates are 0.5, cylinders are 1/pi, and spheres 0.25.

PR #7362, opened 2019-05-25
FDS Source: add FLAME_INDEX_MODEL for testing

rmcdermo:


rmcdermo:
This model automatically sets the value of ZETA_0 based on the "flame index".  The flame index is a measure of the alignment of the gradients of fuel and oxidizer.  If aligned, then flame index is 1 and the flame is considered premixed.  If flame index is -1 then the flame is a diffusion flame.  So, here FI = -1 maps to ZETA_0 = 1 and FI = 1 maps to ZETA_0 = 0.

Invoke the model with FLAME_INDEX_MODEL=T on the COMB line.

PR #7399, opened 2019-06-06
First addition of OMP pragma to pois.f90

ksuarez1423:
Minor upgrade to performance, often negligible, but good practice for OMP pragma.


mcgratta:
To be consistent with the rest of the code, use capital letters for the OpenMP directives.

PR #7410, opened 2019-06-12
Advisor Makefile Update

ksuarez1423:
Most advisor requirements are at the linking stage, and the makefile is now updated to ensure that this occurs correctly. Furthermore, pois.f90 received some now commented-out marking up for use with Advisor, along with parallelization of VRADB5.


mcgratta:
What do the advisor compiler directives do? Are they just for profiling, or are they part of OpenMP?

klockheart:
In the code, the Advisor pieces are commented out - they cannot effectively be in the code at the same time as OpenMP directives, and are actually function calls instead of directives. They are left in as comments, due to further testing.

In the makefile, the flags are required for Advisor's annotation, but were needed for the linking step instead of the compilation step.

PR #7419, opened 2019-06-14
FDS Verification: restore tight tolerance for mwtest_cfl

rmcdermo:


rmcdermo:
I tested this with 19u3 and it works now.

PR #7463, opened 2019-06-28
FDS Source: Fix error in assigning ZONE for removable OBST at a bound…

drjfloyd:
…ary.

### DIFF
--- a/Source/func.f90
+++ b/Source/func.f90
@@ -1426,12 +1426,21 @@ SORT_QUEUE: DO
          END SELECT
       ENDDO
 
-      ! If an obstruction is found, assign its cells the current ZONE, just in case the obstruction is removed
+      ! If an obstruction is found and it has DEVC, CTRL, or is consumable, assign its cells the current ZONE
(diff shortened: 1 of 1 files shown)


drjfloyd:
@mcgratta can you take a look at this change to see if you think it this fix is just going to cause other issues?

The attached case illustrates the issue (raised in this thread https://groups.google.com/forum/?fromgroups=&nomobile=true#!topic/fds-smv/CaHcwuzLUJc). There is an OBST that is initially not present and then added later that sits at a mesh boundary. When zones are assigned during initialization the OBST is assigned the correct zone but the ghost cell values are not changed. The ghost cells around the OBST stay 0. At the first timestep with the OBST not present CONNECTED_ZONES sees the gas cell with a zone of 1 and the ghost with a zone of 0 and says that 1 is connected to 0. As a result no pressure rise happens until the OBST is turned on and disconnects 1 from 0.

I modified the pressure zone search algorithm handling of OBST to set the ghost cell to the OBST zone if the OBST is something that is controllable or can burnaway. 

[Mechanical_obst.txt](https://github.com/firemodels/fds/files/3339504/Mechanical_obst.txt)

PR #7464, opened 2019-06-28
Add Hasemi Localized fire data on steel column, HT3D

cnz2nist:
Hasemi et al. conducted a number of localzied fire tests on steel column.  A case was considered here. HT3D predicted steel temperatures were compared with the measured value. In the FDS model, TIME_SHRINK_FACTOR=10 was used. FDS model need to check. Discussion needed.


mcgratta:
I am on leave for a week. Do the following -- compile the FDS Validation Guide completely with your additional case. Then send or share the document with Randy so he can check it for style and conflicts.

rmcdermo:
Please resubmit after discussion.

PR #7465, opened 2019-07-02
Bugfix and parallelization

ksuarez1423:
Repaired an error in QFDS, added new parallel code to radi.f90, and changed part of pres.f90 to use SIMD lanes.


klockheart:
Tested successfully with Firebot.

PR #7467, opened 2019-07-04
Update Layer Height usage with multiple meshes

sbenkorichi:
Can someone please review.
Updating this section as there is an existing verification case in the repo. `layer_4mesh.fds`


rmcdermo:
@mcgratta I merged, but please double check when you get time.

mcgratta:
This is correct. The layer height feature should work on multiple meshes.

sbenkorichi:
I tested it earlier. This has apparently included only in the release 6.7.1 as 6.7.0 works only with a single mesh.

PR #7479, opened 2019-07-12
New Script for Inspection + Updated INSPECT_Cases

ksuarez1423:
Changed INSPECT_Cases.sh to use a more rigorous test, and created a new script for use in Firebot and by users that wish to automate basic error parsing.


mcgratta:
If we get a clean firebot run, let's remove any input files and scripts from Thread_Check that we are not using. Less is better.

PR #7480, opened 2019-07-16
Shortening race_test_4.fds

ksuarez1423:
Reducing the time of the Inspection case will reduce accuracy to some unknown extent, but 1.2 s tends to be adequate in my personal testing. This is a necessary edit so Firebot can finish its nightly run in a timely manner.

### DIFF
--- a/Verification/Thread_Check/race_test_4.fds
+++ b/Verification/Thread_Check/race_test_4.fds
@@ -7,7 +7,7 @@
 &MESH IJK= 6, 6,32, XB=-0.15, 0.15,-0.45,-0.15, 0.00, 1.60 /
 &MESH IJK= 6, 6,32, XB=-0.15, 0.15, 0.15, 0.45, 0.00, 1.60 /
 
-&TIME T_END=10. /
+&TIME T_END=1.2 /
 
 &REAC FUEL='N-HEXANE'
       CO_YIELD=0.100


rmcdermo:
Does this solve the out of tolerance problem?

klockheart:
I am still uncertain what caused that error. Due to the long-running nature of Firebot, I wanted to push my solution to the most simple problem first.

rmcdermo:
So, what does "timely manner" mean?  Are we under 2 hours with this PR?

klockheart:
In this case, timely manner means that this test should not take more than an hour once it is running, as opposed to the timings seen with the full 10 second test. I can leave it to run tonight to find the exact time difference, but I know that the 10 second case takes a good deal longer.

As for the final speed of Firebot, that will depend heavily on when the job gets scheduled. If it is scheduled at the beginning or in the middle of the testing phase, the inclusion of the job will have less of an effect than if it is scheduled towards the end.

klockheart:
Upon deeper review of the process, I now see that race_test_4 is included in FDS_Cases - thus, it was being run twice, likely simultaneously and resulting in instability in the results. Thus, the solution is likely a matter of removing race_test_4 from FDS_Cases, or making a copy of it with this shorter time, renaming it, and running that in INSPECT_Cases.

I'll test this, and make a new PR with that solution.

rmcdermo:
Thanks.  Yes, it is suspicious that the long run is also happening on the job that is out of tolerance.  Best to fix the whole thing.  Also, we want to make sure firebot finishes in 2 hours, so you need to test the whole firebot on blaze to see if this is true, not just your case.  Thanks.

PR #7483, opened 2019-07-22
FDS Source: Overhaul pressure ZONE specification

mcgratta:


mcgratta:
Timo -- FYI, this PR might break Evac stuff. We now prefer to use a single XYZ triplet to specify a pressure ZONE, even if it crosses mesh boundaries.

PR #7578, opened 2019-09-05
Add CSV output for runtime diagnostics.

JakeOShannessy:
Currently the time step, step size, current time, and simulation time are outputted in the CHID.out file. I find this a very useful source of information but the CHID.out file is not particularly machine readable. It's been [discussed before](https://groups.google.com/d/msg/fds-smv/yS_wRwCOsME/5e_2xCoDBwAJ) that it's probably better to keep the CHID.out file flexible and unstructured and output other information into structured files (such as CSV).

This PR adds CSV output for the four items above (time step, step size, current time, and simulation time). In particular this relates wall time to simulation time, which I'm particularly interested in.


rmcdermo:
Jake, please discuss with Kevin.  I think there is already an option for this.

mcgratta:
There doesn't seem to be any harm or extra time to write out this file. It might be useful for automatically determining how long the case has to run, like "Estimated time to completion..."

JakeOShannessy:
Yes, that's exactly what I use it for, it's a very simple and uniform way to track the progress of simulations.

PR #7608, opened 2019-09-18
FDS Source: Fix SHERWOOD calc in part.f90

drjfloyd:


rmcdermo:
I'm confused.  I thought I fixed this several days ago.

When I look at part.f90 history, I see my change.  Curious why it let you merge.

drjfloyd:
Had pushed up a change but didn't do a pull request. My git was giving me fits trying to update your changes to the solid phase and an easy solution was to do a pull request even though it meant no actual change.

rmcdermo:
OK, that makes more sense.  Thanks

PR #7798, opened 2019-11-18
FDS Source : Fix pointer to CFACE, add more detail in ERROR messages …

marcosvanella:
…for broken GEOMs.


emanuelegissi:
Great improvement of GEOM error messages. This makes debugging easier.

marcosvanella:
Ciao Ema, we have @drjfloyd to thank for this suggestion.

PR #7831, opened 2019-11-27
FDS Source: Add hooks in level set routine for IBM geometry

mcgratta:


mcgratta:
Marcos, there are two places in `vege.f90` to add IBM info, both indicated by `CC_IBM` if statements.

PR #7842, opened 2019-12-04
Energy_budget_gas

MortezaHaghighi:
Hi,
This is Morteza, from Aalto University. I'm working on an energy budget verification test case based on the HT3D routine. While I was investigating solid pyrolysis inside the solution domain, I found that my main problem comes from the enthalpy change of the gas phase. Removing the solid from the domain, I tried to satisfy energy conservation (in the unit of "kJ") between the exterior hot walls and the interior gas phase. Unfortunately, I couldn't even do that; the inward energy doesn't match the enthalpy increment of the gas. Therefore, we decided bringing up this issue here in the hope of finding a solution.


mcgratta:
Create an issue for the Issue Tracker, not a pull request. Just upload the input file. We do not need the output files because we will run the case ourselves.

Issue #7744, opened 2019-11-03
FDS 6.7.3 with FDS+Evac: Evacuation simulation without fire are not longer possible


gjaeger:
Revision         : FDS6.7.3-0-g9a07c366b-release
Revision Date    : Wed Oct 30 13:27:03 2019 -0500
Compiler         : Intel ifort 19.0.4.233
Compilation Date : Oct 30, 2019  22:26:38

**problem:**
For the example [evac_memory_test0.fds](https://github.com/firemodels/fds/blob/master/Verification/Evacuation/evac_memory_test0.fds) with

```
&MISC EVACUATION_MC_MODE=.TRUE., NO_EVACUATION=.FALSE. /
```

instead of ([evac_memory_test0.fds, line 24](https://github.com/firemodels/fds/blob/43221bb324668629b0169ddb39fcf3ae361100e3/Verification/Evacuation/evac_memory_test0.fds#L24)):

```
&MISC EVACUATION_MC_MODE=.FALSE., NO_EVACUATION=.TRUE. /
```

I get the following message:

```zsh
% fds evac_memory_test0.fds

 Starting FDS ...

 MPI Process      0 started on zam433

 Reading FDS input file ...

 EVAC: Emesh 1 FF1stFloor has 4 door flow fields

ERROR: The domain appears sealed. Specify one or more pressure ZONEs. (CHID: evac_memory_test0)

STOP: FDS was improperly set-up (CHID: evac_memory_test0)
```

It looks like _Evac_ simulations without fire are no longer possible.

rmcdermo:
OK, thanks for the report.

@tkorhon1 can you please look at this?

We will have to do another release asap.

rmcdermo:
@gjaeger Did you try specifying a pressure ZONE?

gjaeger:
> Did you try specifying a pressure ZONE?

No. That wasn't necessary until now.

rmcdermo:
Well, given that Timo is not responding, your options are either not use 6.7.3 or try to address the error message FDS is giving.

gjaeger:
As a preliminary solution, the simulation _evac_memory_test0.fds_ runs with the addition:

```
&ZONE XYZ= 25.0,20.0, 1.5 /
```

godisreal:
Below is a quote from a previous issue.  Maybe it is helpful.  
There are different issues that address a relavent problem. 
In brief the coupling of fds and evac is not very good in current mode.  

[...]

PR #7887, opened 2020-01-07
Utilities, Training: Add some FDS+Evac examples and IMO test cases

gjaeger:
Add a few FDS+Evac input files of [examples](http://virtual.vtt.fi/virtual/proj6/fdsevac/documents/FDS+Evac_textbased_homepage.txt) and [IMO test cases](http://virtual.vtt.fi/virtual/proj6/fdsevac/imo_fds6.html) to the _/Utilities/Training/_ directory. It's a first step based on the discussion in issue #7744.


rmcdermo:
Thanks Gregor.  Please note I made minor name changes to the directories.  You may need to update any processing scripts you have.

So, I suppose the plan right now is to simply run the cases and grep on a FDS Completed Successfully?  If so, we will need a list for running.  Can you put together something like fds/Verification/FDS_Cases.sh for the Evac tests?  Call it Evac_Cases.sh or something.  Thanks!

gjaeger:
Thanks, no problem. 

I will create a script for running.

PR #7942, opened 2020-01-27
FDS Manuals: Fix multiply defined citation problem

mcgratta:


mcgratta:
This hopefully fixes the firebot warnings about multiply-defined labels.

rmcdermo:
OK, let's see if this fixes the problem.  But I would like to try to find a better solution.  It is not ideal to have "UG-" prepend the names in the casename column.  I expect that when I want to run FDS_verification_script for, say, pine_needles, that I will get the pine_needles case.

mcgratta:
This is actually the proper way to do things. The errors were due to the fact that we were getting away with something in an earlier version. The way it works is if you're in the Verification Guide, and you want to invoke a label from the User's Guide, you append UG- to the label name to tell LaTeX that the reference is to the external document.

rmcdermo:
That much is fine.  I object to needing to remember which cases will have UG- prepended when I want to run FDS_verificaiton_script.  The point behind that casename column is to make it easy to make individual plots without opening the file and figuring out which row to plot.  This UG business breaks all that.  The casename column needs to be the CHID.

PR #7944, opened 2020-01-27
Verification Test for 1D tunnel flow with linear cp

Cactilio07:
Notes:

the FDS_Verification_dataplot_inputs file gave problems while merging, it should be exactly the same file as before, just with 4 lines more that can be searched by searching linear_cp, it should be fine though

light modifications to one formula in the cons_gamma case, pressure, so that it fits exactly the given results.

The fds file for the test is the same for the const gamma just introducing a linear cp using a ramp.


rmcdermo:
@Cactilio07 we need to get the dataplots inputs issue sorted out.  It is not ok that I can't see the diff.  There should only be a couple of lines that are added.

PR #7945, opened 2020-01-27
Verification case for 1D tunnels with linear cp

Cactilio07:
fixed (rerun with fds debug before sending this request, and works)

Introduction of a Verification case for 1D tunnel flow with linear cp and its subsection in the Verification guide. Minor modifications to the constant gamma case to make the pressure formula match the results given.


rmcdermo:
Nice work @Cactilio07 !

Issue #7943, opened 2020-01-27
DEVC output for PATH_RAMP not working with multiple meshes


rmcdermo:
This concerns the new PATH_RAMP feature when outputting DEVC QUANTITIES with multiple meshes.  In the single mesh case, all seems fine.  In the attached input file, comment the 2 mesh and uncomment the 1 mesh to see the results.  In the single mesh case, the particle path is correct in Smokeview and also in the _devc file (see 1 mesh plot).

With 2 meshes, the particle path is Smokeview is still correct (should be a clue), but the _devc location is not (see 2 mesh plot).

[test.txt](https://github.com/firemodels/fds/files/4119258/test.txt)

[part_path_ramp_jog_1mesh.pdf](https://github.com/firemodels/fds/files/4119260/part_path_ramp_jog_1mesh.pdf)
[part_path_ramp_jog_2mesh.pdf](https://github.com/firemodels/fds/files/4119261/part_path_ramp_jog_2mesh.pdf)

rmcdermo:
This is somehow related to SDV or "SUBDEVICE"s.  The SUBDEVICE values are correct, but the DEVICE values that are getting written to _devc file are not when we have multiple meshes.

mcgratta:
Commit #7959. I also fixed the test case.

PR #7946, opened 2020-01-27
FDS User Guide: add discussion of PATH_RAMP

rmcdermo:


rmcdermo:
I changed part_path_ramp_jog to a single mesh case until the 2 mesh is worked out.  See Issue #7943

PR #7958, opened 2020-01-28
FDS Validation: a few mods to Juelich SETCOM cases

rmcdermo:


drjfloyd:
I don't understand why we would want to change NFRAMES to only DT_BNDF. This will greatly expand the size of all non-BNDF output files.

rmcdermo:
Because I would like to see more detail in the slice files.  I reduced frequency of devices.

PR #7998, opened 2020-02-07
qfds.sh: Add ppn option for SLURM

mcgratta:


mcgratta:
This commit adds the ability to control the number of processes per node (ppn) in SLURM. This is the -n option in the qfds.sh script.

PR #8072, opened 2020-02-23
Source vege.f90

ruddymell:
Simple changes for consistency of LS1 with WFDS9977


rmcdermo:
@ruddymell 

Sorry, but we need to rework this PR.  A few things:
 - You inadvertently added an exe that is 15 MB binary file.
 - There are too many merges commits with things added and removed.  I don't accept these because it is too likely that something we do not want gets injected into the repo.
 - If we are going to comment code, just remove it.  The whole point of the version control system is that the code is still there if we ever need it.  We are going for clean, readable source on the master branch.

Thanks!

PR #8083, opened 2020-02-26
Update evac_memory_test0.fds

godisreal:
When I run this example, I find that 3D smoke cannot be visualized.  Because SMOKE3D=.TRUE., DT_SLCF should be effective.  There is also SLCF defined in the input file.  This example may be improved in several other aspects.


rmcdermo:
I am not sure what we are testing here.

PR #8084, opened 2020-02-26
FDS Source: Add MEAN_FORCING output quantities

mcgratta:


Verification/Atmospheric_Effects/wind_example.fds
```diff
@@ -7,6 +7,8 @@
 
 &WIND SPEED=5., RAMP_DIRECTION='theta', Z_0=0.01, L=-100. /
 
+&REAC RADIATION=F /
```
rmcdermo, line 10: Did you mean this to be a REAC line?  Shouldn't it be RADI?

mcgratta: Yes, I added at the last minute to reduce time.

PR #8094, opened 2020-03-02
revised ScaRC version with OpenMP support and two different matrix st…

SusanKilian:
Completely reworked ScaRC version with:
- more modular redesign and revised communication structures
- OpenMP support for long loops
- two different matrix storage techniques (compact and bandwise, the latter in combination with OpenMP, but only applicable for structured discretizations)
- further, partly still experimental procedural options (which are explained if necessary and are hidden by directives)


rmcdermo:
I will test then merge

SusanKilian:
Thanks

rmcdermo:
Thanks, Susan.  Things seem to be working fine with the scarc verification cases.  Given that MKL has a memory leak, if ScaRC can step in and be a workable solution, that would be fantastic.  I'm going to test on a real case today.  I'll keep you posted.

SusanKilian:
Hi, Randy, 

that sounds fine. Please note that the baroclinic pressure correction should also work with ScaRC now.

I already tested the EVS-09-20 tunnel-case from https://github.com/firemodels/fds/issues/7929 with

&PRES SOLVER='SCARC', PRESSURE_TOLERANCE=100., MAX_PRESSURE_ITERATIONS=50 /

which seemed to work.

If you use the structured variant - that is simply ‚SCARC‘  which based on local Crayfishpak FFT's for preconditioning - it uses the banded matrix storage technique by default which should now be applicable in combination with OpenMP. 

Best, Susan

PR #8095, opened 2020-03-02
add Scarc variants for tunnel_demo case with one or two grid levels

SusanKilian:
I would like to add two ScaRC variants for the tunnel_demo case:  
- tunnel_demo_scarc.fds uses the default SCARC version (data-parallel CG with local FFT preconditioning and bandwise matrix storage technique).  
- tunnel_demo_scarc_twolevel.fds additionally uses a coarse grid solution per preconditioned CG iteration, which consists of applying the global CLUSTER_SPARSE_SOLVER on the **once** coarsened overall grid.

Here, the use of the additional coarse grid preconditioning leads to a considerable reduction in the convergence rate of the CG method (from an average convergence rate of 0.61 with about 52 CG-iterations per pressure solution to a convergence rate of 0.16 with about 14 CG-iterations).

Nevertheless, the computing time may be longer, since the ratio of computing operations to communication is rather unfavorable for the coarse grid problem. I'm currently in the middle of testing a master-slave combination that outsources the solution of the coarse-grid problem to a master process, but at this point I'm not sure if it will really bring a gain. I'll keep you informed.


rmcdermo:
Susan,
The good news is that this are running and producing good results!
The bad news is that things are running pretty slowly.  So... couple of questions.  All I am doing is using
```
&PRES SOLVER='USCARC', MAX_PRESSURE_ITERATIONS=1/
```
I need `USCARC` for this case.  But are there other options I can use to try to improve speed?

Next, when I look at the `chid_cpu.csv` file the `PRES` column claims I am only taking 0.2 % of the calculation.  Do you have SCARC timings hooked up to the _cpu file?  If not, let's add that asap.  If so, then I need to figure out what else is wrong.  Thanks

SusanKilian:
Randy, yes, I know it's the same old story and it seems to be in the nature of things (more global information, more runtime). I have already tried so many ways to speed things up, I'm still at it and I haven't given up hope yet.

For USCARC it's harder to get further acceleration because only the compact matrix storage technique can be used and the use of OpenMP is more limited. USCARC in its default version is based on local PARDISO preconditioning and I am currently in the process of re-enabling the single precision variant (I have only now noticed a bug in the currently checked in version related to the **single-precision** Pardiso solver, sorry ... but this **does not** concern the default double precision Pardiso).

I will address the chid_cpu.csv issue shortly!

Could you please send me the geometry file for the test case which you mentioned above?

rmcdermo:
Hi Susan,

We understand that the global solver is slower.  But it it now clear where we can adjust parameters to loosen the tolerance.  Please provide some guidance on this.

It will be very good to get the timings into the cpu file.

Here is the test case I was running.
[test_uscarc.fds.txt](https://github.com/firemodels/fds/files/4282146/test_uscarc.fds.txt)

Cheers,
Randy

PR #8096, opened 2020-03-02
FDS Source : Update to FDS cut-cell definition scheme.

marcosvanella:


emanuelegissi:
Does this pull request mean that the current code can deal with intersected
and self-intesected GEOMs?
Emanuele

PR #8097, opened 2020-03-03
Dump ScaRC timings to chid_cpu.csv and enable single precision for MKL preconditioned and coarse grid solver

SusanKilian:
The time measurements for the ScaRC solver in chid_cpu.csv should work now (see the addition of T_USED(5) at the end of the SCARC_SOLVER routine).

Also, there is now a new parameter SCARC_MKL_PRECISION which can be used to control the accuracy of the MKL solver, either to 'SINGLE' or to 'DOUBLE' (the latter is the default value). This means that you no longer need to compile with the WITH_MKL_FB directive to change the accuracy.

You can also control the accuracy of the CG solver using SCARC_KRYLOV_ACCURACY and SCARC_KRYLOV_ITERATIONS.

**Example:**
&PRES SOLVER = 'USCARC'                               / ---> Default: 'PARDISO' preconditioner
            SCARC_MKL_PRECISION = 'SINGLE'      / ---> Default: 'DOUBLE'
            KRYLOV_ACCURACY = 1E-6                   / ---> Default 1E-8
            KRYLOV_ITERATIONS = 100                   / ---> Default 1000

Here, the Pardiso preconditioning would only be executed in single precision and the CG solver would stop after 100 iterations or after reaching an accuracy of 1E-6, whichever is reached earlier. 

If SOLVER = 'SCARC' is set for a structured discretization, the default preconditioner is 'FFT', but via:
   
&PRES SOLVER = 'SCARC'   
             SCARC_PRECON = 'PARDISO'
             SCARC_MKL_PRECISION = ...
             KRYLOV_ACCURACY = ...
             KRYLOV_ITERATIONS = ...

the MKL preconditioner can be selected instead with the desired changes of the iteration parameters, if needed for test purposes.

Before 'Run Time Diagnostics' in chid.out the selected ScaRC parameters should be listed.

All this is hot-needle knitted as we say in Germany. I hope it all works.


rmcdermo:
Thanks for the explanation.  This is very helpful!

SusanKilian:
Great! I have started your test case in different variants and will get back to you tomorrow.

SusanKilian:
I have calculated your test case test_uscarc.fds in different constellations. 

Apart from your 20 mesh decomposition into 10x1x2 meshes, I have considered further mesh decompositions into 2x1x1, 5x1x1, 5x1x2, 10x1x1 meshes.

For all constellations I considered the single and double precision version for the PARDISO preconditioning of USCARC.

To keep the computing times manageable, I only calculated up to time T_END=50.

Attached is an overview of the resulting TIME STEPPING WALL CLOCK TIMES and the UScaRC convergence statistics.

Obviously, the convergence rates for the single-precision version become worse (and thus the number of iterations higher). Nevertheless, the single-precision version's computing times are significantly shorter than the double-precision version's - with apparently the same result - on average only 75% or less.

This is of course only a small snapshot. And one would probably also have to look at how it behaves when simulating larger T_END times. But maybe this is an option for saving time.
What do you think about this?

In the end, however, the basic convergence rate must be further improved, which in my opinion is only possible by adding coarse grid information. When using multiple grid levels (for the structured variant), convergence rates are significantly lower and usually in the range of 0.1 up to 0.2 (but surely at the price of higher communication effort). Unfortunately, it is always extremely difficult to find a good balance between numerical and parallel efficiency.

[UscarcComparison.pdf](https://github.com/firemodels/fds/files/4287649/UscarcComparison.pdf)

rmcdermo:
Susan,
Thanks for looking at this.  It would also be helpful to have the total simulation time in the chart.  Since you have these all run, can you please add this?
Based on these results, however, 25% to 30% speed up on average is significant, and would be enough for me to make the single precision smoothing the default.

SusanKilian:
Randy, enclosed you will find the table extended by the Total Elapsed Times. Since ScaRC hardly needs any initialization times, this does not change the basic relationship.

I must admit that I was also surprised that the difference is so big. 

What also becomes clear for me in these times is the fact that doubling the meshes really halves the time. At least in this constellation.

Do you think it makes sense if I run the same calculations again overnight at a longer T_END time to check if this trend continues (which I expect)?

[UscarcComparison2.pdf](https://github.com/firemodels/fds/files/4287892/UscarcComparison2.pdf)

rmcdermo:
Thanks, but I think I was not clear.  I think we need a column for PRES and a column for TOTAL.  Or, just attach for the 20 mesh case your cpu file.  Thanks

rmcdermo:
I do not think running to T_END will drastically change the conclusions for this case.

SusanKilian:
Oh, I got that wrong, sorry. Attached are now the two CPU files for the 20 mesh case in single and double precision and for the sake of completeness also the corresponding FDS files. 

[uscarc20_double.fds.txt](https://github.com/firemodels/fds/files/4287985/uscarc20_double.fds.txt)
[uscarc20_single.fds.txt](https://github.com/firemodels/fds/files/4287986/uscarc20_single.fds.txt)

[uscarc20_double_cpu.csv.txt](https://github.com/firemodels/fds/files/4287970/uscarc20_double_cpu.csv.txt)
[uscarc20_single_cpu.csv.txt](https://github.com/firemodels/fds/files/4287971/uscarc20_single_cpu.csv.txt)

SusanKilian:
and finally also the whole overview ...
[UscarcComparison3.pdf](https://github.com/firemodels/fds/files/4288088/UscarcComparison3.pdf)

rmcdermo:
For the same mesh arrangement (20 meshes) UGLMAT takes 65% of the total cpu time.  SCARC takes 91% for single precision (93% for double).  The goal is to get SCARC under 65% (for now) without OpenMP.  When we run large cases, OpenMP is not a useful strategy, we allocate all one core per mesh and fill up the cluster.  So, now that we have accuracy sorted out, the focus shifts to efficiency.  Let's see if we can make it work.

SusanKilian:
Yes, I really hope so, and I'm juggling lots of different approaches right now. 
From my point of view one should also take care not to split a whole grid into unnecessarily many meshes just to have many processors at work, because the block preconditioning converges worse with an increasing number of blocks and the total time does not necessarily get shorter. Of course, it is difficult to find a suitable balance for any test case a priori.

I am currently looking very hard for a way to reduce the communication costs of using a coarse grid. Because, as I said, the total number of iterations is usually reduced considerably. If only the costs for a single iteration (or its coarse grid portion) could be reduced even partially, we would probably be in business. I am working on it!

rmcdermo:
I understand.  But it is not realistic to expect users not to use cores for domain decomposition.  We must still maintain strong scaling.  "Unnecessary" meshes are only ones that are smaller than, say, 16^3.

SusanKilian:
Yes, I am aware of that and the decompositions are certainly often geometrically motivated. Therefore I don't see any other possibility than to use global information, because the purely blockwise approach numerically reaches its limits when increasing the number of meshes.  

Nevertheless, I found it rather disillusioning how much global communication or the bad relation of computational part to communication (for an additional coarse grid solver) weighs. As mentioned before, I had the opportunity to do some bigger tests on the Lunarc cluster in Lund. The considered test cases (geometric multigrid with different numbers of grid levels / CG with additive coarse grid solver) showed convergence rates which were smaller by a factor of 4 or more than with pure CG, but still needed twice the computing time. With Scalasca this could clearly be traced back to the coarse grid solver.

PR #8105, opened 2020-03-04
FDS Source: Add SURFACE_DIAMETER to control drop diameter on solid

mcgratta:


mcgratta:
Randy -- this commit does not change the current code functionality, but it does give you control over the droplet size when it hits a surface.

PR #8203, opened 2020-03-31
FDS Source: Reduce memory requirements for VENTs.

mcgratta:


mcgratta:
Before this commit, all VENT lines for all meshes would be read and allocated by each MPI process. In most cases, this is not a problem. However, in some WUI cases, there can be thousands of VENTs describing different kinds of vegetation. With this commit, we now only read the VENTs for the meshes that are controlled by a given MPI process. 

Jason -- note that there is now an MPI call in PROC_HVAC because when we search for VENTs that connected to nodes, we can now only search the meshes under a given MPI process control, and then collect the results (for FOUND) using an MPI_ALLREDUCE. 

Timo -- check your Evac cases. This might affect them.

PR #8215, opened 2020-04-03
FDS Source: Fix MASSFLUX calc for MLRPUA with different MATL HoC than…

drjfloyd:
… REAC HoC


mcgratta:
How did firebot pass last night without this fix?

rmcdermo:
Looks like it has not been merged.

But I suspect the verifications have not been added yet, which explains the pass.

drjfloyd:
Was interrupted. Merged now. Verification case will be added soon.

PR #8226, opened 2020-04-08
FDS Source: Fix bug related to MP5 flux limiter

mcgratta:


mcgratta:
Randy -- take a look at this. The immediate problem with the soborot case was the fact that all 4 values of RHO*ZZ are used in `SCALAR_FACE_VALUE`, but at boundaries we only supply 3 values. The other value is probably NaN and causes crash. You might want to take a look at all calls to `SCALAR_FACE_VALUE` in mass.f90, divg.f90, and velo.f90. I only fixed what was need to get past the current crash of the soborot case.

rmcdermo:
I dialed back the code to before your fix and tried to reproduce this with 2020 update 1 on my mac and the case runs in release with -ipo.  Very weird.

mcgratta:
Either this is a compiler  bug, or different OS and optimization levels handle the seg faults differently.

rmcdermo:
Well, I'm going to leave this alone for now.  The mp5 cases are running with your fix.  It's a pain to debug on blaze if you are not vi proficient, which I'm not.  I'll revisit if we see more problems when we return.

PR #8229, opened 2020-04-09
FDS Source: Change names of MASSFLUX variables only

mcgratta:


Source/dump.f90
```diff
@@ -7911,15 +7911,15 @@ REAL(EB) FUNCTION SOLID_PHASE_OUTPUT(NM,INDX,Y_INDEX,Z_INDEX,PART_INDEX,OPT_WALL
       ENDIF
    CASE( 7) ! BURNING RATE
       IF (N_REACTIONS>0) THEN
-         SOLID_PHASE_OUTPUT = ONE_D%MASSFLUX_SPEC(REACTION(1)%FUEL_SMIX_INDEX)
+         SOLID_PHASE_OUTPUT = ONE_D%M_DOT_G_PP_ACTUAL(REACTION(1)%FUEL_SMIX_INDEX)
```
rmcdermo, line 7914: @mcgratta This is a good start.  But to me these names are still ambiguous.  Does "actual" refer to pyrolyzate or the surrogate gas species?

mcgratta: `ONE_D%M_DOT_G_PP_ACTUAL(N)` is the actual mass flux (kg/m2/s) of tracked species N. 

`ONE_D%M_DOT_G_PP_ADJUST(N)` is the adjusted mass flux (kg/m2/s) of tracked species N. 

Both of these refer to the generation rate of gas species N at a boundary. The word `ADJUST` means that it is the effective generation rate that is consistent with the single gas phase reaction HoC.

rmcdermo: What always confuses me is that I think of "actual" as what is "actually" put into the gas phase simulation.  I know it is too long of a variable name, but what I'm after is something that conveys:

ONE_D%M_DOT_G_PP_BASED_ON_HOC_MATL
ONE_D%M_DOT_G_PP_BASED_ON_HOC_REAC

mcgratta: If we come up with a better name, we can easily do a global replace.

PR #8244, opened 2020-04-13
  FDS Validation: Correct reference for DelCo Trainers report.

rpeacoc:
Referenced NIST TN has only one author
[NIST.TN.1953 Cover.pdf](https://github.com/firemodels/fds/files/4471019/NIST.TN.1953.Cover.pdf)


rmcdermo:
Thanks, Rick!

PR #8247, opened 2020-04-14
FDS Source: Remove unused variable and add array bound to prec.f90

mcgratta:


mcgratta:
FYI -- I made `MAX_INPUT_ID=40` a `prec.f90` parameter.

PR #8255, opened 2020-04-16
hadibordbar

hadibordbar:
The soot radiation model (of RADCAL) is modified by implementing the correlation given for complex index of refraction of soot by the following reference. By this, alfa in calculation of K_eta__soot=alfa*volume fraction*eta would be function of eta. In RADCAL it was fixed to 7.

Chang, H. and Charalampopoulos, T.T. (1990) Determination of the Wavelength Dependence of Refractive Indices of Flame Soot. Proceedings of the Royal Society of London. Series A, 430, 577-591.
https://doi.org/10.1098/rspa.1990.0107


mcgratta:
Is this change documented in the FDS Technical Reference Guide? We cannot add new features to the code until they are documented at the very least, and verified and validated if possible.

hadibordbar:
Not yet, I did run the verification suite but didn’t remake the verification guide.
I will write a short description for the technical reference guide. Shall I send it to you or …? Sorry it’s my first pull request and still am in learning phase of the process…

Take care and stay healthy,
Hadi

drjfloyd:
Looking at the abstract it discusses premixed propane flames.  Is this equation based solely on that data? If so what is the confidence that this is an equation that would apply generically to non-premixed flames? Have you plotted the result as a function of lambda and do values make sense?

mcgratta:
It appears that you are proposing an improvement to the current method. We
need to consider the costs and benefits:

Costs

   1. Is there increased CPU time?
   2. More variables to input/maintain?
   3. More complex description and documentation?
   4. More lines of code?

Benefits

   1. Lower CPU time?
   2. More accurate results?
   3. Better theoretical justification?

If this change effects only the start-up time, no problem. But if we call
this routine every time step, this may be an extra CPU cost. More accurate
results---you need to run the verification cases and ensure that all
results remain within tolerance. This is the very least to do. It might
also be worth looking at an appropriate set of validation cases and
making sure that the change does not adversely affect the results. Similar
results are OK if the other costs are minimal, but if the costs are high,
we would want to see a noticeable improvement.

>

hadibordbar:
> Looking at the abstract it discusses premixed propane flames. Is this equation based solely on that data? If so what is the confidence that this is an equation that would apply generically to non-premixed flames? Have you plotted the result as a function of lambda and do values make sense?

Thanks for the notice.
Here are the changes of complex index of refraction with wavelength in various studies (from Modest book)
image

I separately plotted the Change and Charalampopoulos correlations as
n-k-1

I did check the Planck weighted c used in (K_eta=c * volumefraction * eta) calculated by the Chang correlations of complex index of refraction. 
c_2

For example for the soot volume fraction of 10e-6, I compared the spectral Kappa calculated by a constant c and the spectral c  
soot_abs

The constant values for c are suggested as c = 6.3 for combustion oil, 4.9 for propane, 4.0 for acetylene, 3.7–7.5 for coal combustion, and 4.1 for methane flames (S. Yagi, H. Iino, Radiation from soot particles in luminous flames, in: Proceedings of the Eighth Symposium (International) on Combustion, Pasadena, CA, 1961, pp. 288–293.). RADCAL uses c=7 but I couldn't find any reference to confirm it.
In brief, c is a spectral variable and the proposed changes account for this. Using constant c is Ok, but c should be adjusted for different fuels and fire scenarios. C=7.0 used by RADCAL perhaps overestimates the soot emission for most of the cases.

hadibordbar:
> It appears that you are proposing an improvement to the current method. We need to consider the costs and benefits: Costs 1. Is there increased CPU time? 2. More variables to input/maintain? 3. More complex description and documentation? 4. More lines of code? Benefits 1. Lower CPU time? 2. More accurate results? 3. Better theoretical justification? If this change effects only the start-up time, no problem. But if we call this routine every time step, this may be an extra CPU cost. More accurate results---you need to run the verification cases and ensure that all results remain within tolerance. This is the very least to do. It might also be worth looking at an appropriate set of validation cases and making sure that the change does not adversely affect the results. Similar results are OK if the other costs are minimal, but if the costs are high, we would want to see a noticeable improvement.
> […](#)

No substantial CPU increment is expected as it is used only in producing look up table of Kappa in the beginning. Currently it is done by using a constant c over the whole spectrum as discussed above. The new proposed changes makes the c spectrally varying. The additional CPU cost is very small. I did the verification cases. However, I can not perform the validation run until next week. Will update you then.

drjfloyd:
I found the paper that is in the original RADCAL manual as being the reference for the soot values and I don't see where the 7 came from either.  I've plotted the new and old FF equation as a function of Lambda and attempted to overlay that on what appears to be a similar plot from the paper referenced in RADCAL.  It does appear that RADCAL trends high. If we do decide to accept this pull request, it should also be done in the RADCAL repository with the change in documentation for soot made there.  

image

hadibordbar:
> I found the paper that is in the original RADCAL manual as being the reference for the soot values and I don't see where the 7 came from either. I've plotted the new and old FF equation as a function of Lambda and attempted to overlay that on what appears to be a similar plot from the paper referenced in RADCAL. It does appear that RADCAL trends high. If we do decide to accept this pull request, it should also be done in the RADCAL repository with the change in documentation for soot made there.
> 
> image

Thanks. I agree. 
Could you please share that old paper of Dalzell and Sarofim? The one I have is missing one page. thanks.
Hadi

drjfloyd:
Hadi,

First, thanks for submitting this pull request. This approach is not a significant CPU burden and to me seems a better grounded approach theoretically than the constant value we have now. 

Your pull request started some discussion about the best way to maintain radcal. As a result we made some changes to fds and to radcal so that both programs could use the exact same rcal.90 file with the official version being the one in the radcal repository (https://github.com/firemodels/radcal). I am going to close your pull request for fds. We'd like for you to make your change using radcal, document the new soot approach in the radcal guide, and do some verification to demonstrate what impact this has in radcal and that the impact is what is expected. This might be by running a number of cases with differing amounts of soot and varying gas temperatures with the old and new versions of radcal and comparing them. I think we also want to satisfy ourselves that the specific fit you have is broadly applicable for different fuel types (or at least no worse than the current approach in terms of its range of error). Once that is done, then please resubmit this as a pull request first to the radcal repository.

hadibordbar:
Hi Jason,
Maintaining Radcal in an integrated manner with FDS is a great idea. 
I will run some cases soon and will update you then by submitting another pull request to Radcal repository. It may take a week or two due to other pending tasks.
Take care,
Hadi

Issue #8269, opened 2020-04-18
Allow different I,J,K for DEVC QUANTITY's


mcgratta:
Now that `FACE_CENTERED` device `QUANTITY`s interpolate differently than `CELL_CENTERED`, we must allows `QUANTITY` and `QUANTITY2` to have separate IJK values.

mcgratta:
These changes passed firebot and appear to be working.

RIO4FDS:
> These changes passed firebot and appear to be working.

Everything working fine with fds2ftmi verification cases either! :)

mcgratta:
Good to hear. The functionality is internal---hopefully it does not alter inputs and outputs.

PR #8278, opened 2020-04-20
FDS Source: Issue #8269. Allow for multiple QUANTITYs on a DEVC line

mcgratta:


mcgratta:
This change allows a single point DEVC to have two QUANTITYs with different I,J,K coordinates. This comes up when we have cross correlations involving velocity components. For a given XYZ, we might have different I,J,K for the U, V and W components. This change does not change the inputs to DEVC, but it leaves open the possibility of having more than 2 QUANTITYs later on because now these things are stored in allocatable arrays. In other words, we no longer use QUANTITY2 internally, even though the input parameter is still used.

PR #8306, opened 2020-05-01
FDS Source: Rework 1-D heat transfer routine to allow substeps

mcgratta:


mcgratta:
This PR restructures the 1-D heat conduction routine in wall.f90 so that the time step can be automatically reduced if the flux or reaction source terms warrant. Details are in the Appendix of the FDS Tech Guide where the 1-D conduction solver is described. There is also some discussion of a few new parameters in the User's Guide where currently WALL_INCREMENT is described.

PR #8340, opened 2020-05-19
FDS Source: Improve level set fire coupling

mcgratta:


emanuelegissi:
If I understand correctly, the RGB setting is now embedded into the 13
Anderson conditions, right?
Emanuele

rmcdermo:
Yes, these are now set by default, though looks like they can be reset if desired.

mcgratta:
Ruddy Mell and I are working to streamline the level set parameters so that QGIS2FDS can create a simple decoupled level set input file. That is, the case would run only the level set calculation and freeze the CFD calculation. The wind would be the specified wind, not the CFD calculated wind. No fire. Just the level set spread similar to what you would see in Propagator. We are not done yet, so wait a few days.

emanuelegissi:
Okay

PR #8359, opened 2020-05-23
Source: handle firebase presence in grid cell

usfs-fds:
Account for the subgrid fuel gas mass flux from level set modeled firebase (resolved or subgrid).


drjfloyd:
Double check this. I think the change at line 2155 will break wall.f90.  No TAU_MF or RAMP_MF will be evaluated with this change unless a user sets BURN_DURATION but that would not be set for anything other than a WUI case.

usfs-fds:
I think you are right. I'll fix this and resubmit.

mcgratta:
This new code seems far more complicated than it needs to be. Could you write a short set of notes explaining what you are doing and I'll see if I can code it up more simply.

mcgratta:
I will commit alternative code that does something similar.

PR #8361, opened 2020-05-24
FDS Source: Relabel HTC model flags; separate BOUNDARY FUEL model

mcgratta:


mcgratta:
This commit mainly just changes the flags for convective heat transfer models. We were using, for example, `H_ABL` as a flag for the ABL model. But we also use `H_FORCED` as the actual heat transfer coefficient value. This was a bit confusing, so now `H_ABL` is `ABL_HTC_MODEL`. This doesn't affect code input, just internal flags. They are more clear now.

In addition, I created a separate `BOUNDARY_FUEL_HTC_MODEL` for the `BOUNDARY_FUEL_MODEL`. It's the same as the `DEFAULT_HTC_MODEL` except it uses a geometry appropriate for the vegetative fuel element, not the overall surface.

PR #8364, opened 2020-05-26
Change wall surface temp to first node, not surface. Test doxygen in radi.f90

shostikk:


rmcdermo:
@shostikk Can you give some more explanation of this commit?  Jason, Kevin, and I have been discussing energy conservation of the 1D solid phase solver in regard to remeshing.  The node versus face issue has been discussed.  Is this at all related?

shostikk:
I consider this a temporary attempt to fix. Not solving the main problem, which is the energy conservation of moving meshes. Surface temperature value based on the virtual node seems over sensitive. 

You can revert it back if it clashes with what you are doing.

rmcdermo:
I worry that what you have done is first-order accurate.  The surface temperature should obey a continuity of flux boundary condition, which you write out in the tech guide.  I would like for all of us to agree on making such a big change.  Please write up notes on the new numerical method and circulate.  Thanks.

mcgratta:
I agree. Focus for now on the change I made a few weeks ago to allow for sub-iteration of the source term. This should not have changed the algorithm in a case where there is no need for sub-iteration, which is most of the time. I want to make sure that by rearranging the computation of various terms I did not cause a problem.

drjfloyd:
Do we have too many spoons in the pot now in this routine?

Kevin's changed introduce some minor issues related to computing the convective flux and heat transfer coefficient that needs fixing. I have the changes but did not have time this weekend to test and push up. 

I've been working on an approach for the surface temperature issues that is energy conserving and it includes a routine for doing energy conservation for moving meshes. It hasn't been committed. Rather than interpolate temperature like is done currently, the routine apportions the enthalpy and mass from the old mesh to the mesh. Temperature is then extracted from the mass and enthalpy. For constant specific heat materials this is trivial. For temperature dependent materials, this is an iterative process.

shostikk:
I think I tested Kevin's modification to remeshing, and I think it is working correctly. 

But let's cancelt that commit for now till the other changes affect. (How did I do that?)

mcgratta:
Simo -- just edit wall.f90 and restore the previous lines of code. No need to do a revert, because your changes to radi.f90 are fine.

PR #8384, opened 2020-05-28
FDS Source: Include AREA_ADJUST in calculation of some solid outputs

mcgratta:


mcgratta:
This commit ensures that quantities per unit area that are computed by SOLID_PHASE_OUTPUT include the AREA_ADJUST when appropriate. This mainly affects mass fluxes and burning rates. Checked with the verification cases, but watch out for problems in cases where the grid and vent are not perfectly aligned.

PR #8387, opened 2020-05-29
FDS Source: Allow user input for solid-to-droplets heat transfer coeffcient

shostikk:
Source code update
Brief documentation in user's guide.


shostikk:
Btw, I cannot compile guides at the moment. I haven't been able to run all verification tests in the new computer (some issues with the queuing system), and don't have all the figures or data for tables available.

drjfloyd:
You can get a zip of the figures from the nightly builds page.

https://pages.nist.gov/fds/manuals.html and go to nightly builds and then the figures folder.

drjfloyd:
Would a fixed h for a droplet be better as a PART parameter so it is tied to the particle class rather than all particles of all types?

mcgratta:
I'll merge the changes, but take a look at Jason's comment. The particle class derived type does not take up that much memory, and it's generally a good idea to store properties there rather than globally.

shostikk:
Don't know if I missed something, but the H_SOLID_TO_DROPLET is LPC property, not constant.
read.f90:
   LPC%H_SOLID_TO_DROPLET  = H_SOLID_TO_DROPLET*1000._EB

part.f90:
                     H_WALL   = LPC%H_SOLID_TO_DROPLET

Isn't this what you meant?

drjfloyd:
Your pull request had H_SOLID_TO_DROPLET on both MISC and PART and I only saw the one one MISC.

shostikk:
OK. MISC only contains the flag for using the correlation rather than the constant.

drjfloyd:
On MISC it forces all droplet classes to have the same behavior. In most cases this is probably OK since we generally have just one class for sprinklers, but we could also have the default value for  LPC%H_SOLID_TO_DROPLET be a negative value and then in part.f90 check to see if it has been overidden with a positive value.

shostikk:
You are right. I did not realize that. Thanks.

PR #8389, opened 2020-05-29
FDS Source: Change interpolation during REMESH to be energy conserving

drjfloyd:
This replaces the previous linear interpolation of temperature when a wall is remeshed. The new method reapportions the enthalpy (h) and the average specific heat (c) from the old wall noding to the new wall noding. For nodes where all materials have a constant c, the new temperature is h/c. If there is a temperature dependent c then a newton method is used to find the new temperature.


rmcdermo:
Looks nice.  Problem solved?

drjfloyd:
I still need to go back to the case I was running back when this all started.

PR #8401, opened 2020-06-03
FDS Source: Fix errors and warnings from GCC 9.1 compiler

mcgratta:


mcgratta:
Jason, some of these warnings/errors from GCC 9.1 compiler are related to changes you made last week. Look them over and make sure I have not changed any logic.

PR #8446, opened 2020-06-15
bug fix: file id was sometimes wrong before in the variable number portion of the name.

BYUignite:


rmcdermo:
@BYUignite Hi David, Thanks for the contribution!  You might notice I'm being pretty liberal with merges without review.  The reason is that we don't use these Python scripts yet in any continuous integration testing.  They have mainly been added for my own education in Python.  As we move forward, if these scripts get integrated into your workflow to any significant degree, I'd like to consider a unit test or two.  For the time being, we can keep things casual as long as the codes are not too big.  Thanks again!

BYUignite:
Hi Randy,

Yes, I definitely noticed and figured it was due to the relative independence of the scripts. I am very happy to provide any unit tests needed and I’ll take direction on that. I created a parallel github repo for these scripts that includes the case files that these were tested for and some additional documentation  in the spirit of what you describe. In fact, the bug fix was needed due to not having a sufficiently complicated case the first time around. The recent fix was tested using a 16x4x4 mesh with 2 planes in y, 5 planes in z, and 4 planes in x (in that order) with 8 variables in each. About half the planes live on mesh boundaries. The domain is assumed to be rectangular, with no oddly shaped mesh connectivity (like L-shaped, etc.). I don’t actually envision too much development at this point, but I have found this tool useful.

David

David O. Lignell
Professor, Chemical Engineering
Brigham Young University
801-422-1772 | http://ignite.byu.edu

On Jun 15, 2020, at 5:48 PM, Randy McDermott <notifications@github.com<mailto:notifications@github.com>> wrote:



@BYUignite<https://github.com/BYUignite> Hi David, Thanks for the contribution! You might notice I'm being pretty liberal with merges without review. The reason is that we don't use these Python scripts yet in any continuous integration testing. They have mainly been added for my own education in Python. As we move forward, if these scripts get integrated into your workflow to any significant degree, I'd like to consider a unit test or two. For the time being, we can keep things casual as long as the codes are not too big. Thanks again!

PR #8459, opened 2020-06-18
Take droplet cooling into account in AST

shostikk:
The main capability is there, but not working accurately.

New verification test input adiabatic_surface_temperature_spray.fds has three stages: 
heating, heating + spray cooling, only spray cooling. In the end of heat stage, wall temperature and AST should converge to same value, but they do not accuratelly.


shostikk:
[adiabatic_surface_temperature_spray.pdf](https://github.com/firemodels/fds/files/4798067/adiabatic_surface_temperature_spray.pdf)

At the moment, looks like this, but it will be easier to improve if we merge now.

rmcdermo:
@shostikk I am fine with merging.  Let's get this verification case up and running and see what the problem is.

PR #8470, opened 2020-06-20
FDS Source : Change output conditionals for reg solid faces.

marcosvanella:


Source/dump.f90
```diff
@@ -5187,12 +5187,12 @@ SUBROUTINE GET_SOLIDREGFACE_SCALAR_SLICE(X1AXIS,I,J,K,VAL_CF,&
 CCSUM = SUM(CC1(LOW_IND:HIGH_IND))
 IF( CCSUM > 0._EB ) CC1(LOW_IND:HIGH_IND)=CC1(LOW_IND:HIGH_IND)/CCSUM
 
-IF (CC1( LOW_IND)==1._EB) THEN
+IF (CC1( LOW_IND)>0._EB) THEN
```
rmcdermo, line 5190: Should we make this TWO_EPSILON_EB?

marcosvanella:
They are set to 0._EB originally, but it wouldn't hurt. It is to make sure we don't call gas phase output if we don't need to. I'll change it.

PR #8486, opened 2020-06-24
FDS Source: Fix definition of KAPPA_PART for solid particles

mcgratta:


mcgratta:
Simo, can you take a look at this change when you have a chance. I am restoring the expression for KAPPA_PART in radi.f90. Originally, KAPPA_PART=A/(4\*V), where A is the surface area of the particle and V is the volume of the grid cell where the particle resides. I changed the 1/4 factor to SHAPE_FACTOR, thinking that this factor was related to the shape of the particle, and 1/4 is the SHAPE_FACTOR for a sphere (pi*r^2/4\*pi\*r^2). However, I don't think that this is correct, because if you look at the expressions for QR_W and Q_RAD_IN and Q_RAD_OUT, you find that the 4 is not because of the shape factor of a sphere, but rather it's the factor of 4 you find in U related to the integration of I over 4\*pi. I am assuming that the particle is convex; that is, its outward normals do not intersect the surface.

Can you confirm this please. I've already committed this, but if you can double check, that would be great. Thanks.

shostikk:
KAPPA_PART should be particle CROSS SECTION area divided by cell volume. Therefore, 
KAPA_PART = LPC%SHAPE_FACTOR * LP%ONE_D%AREA / (DX * DY * DZ) should be just right.

mcgratta:
Yes, if the radiation is coming from a certain direction, and the particles are oriented a certain way. However, only when the SHAPE_FACTOR is 0.25 does the energy balance work; that is QR_W*VOL_CELL = Sum (Q_RAD_IN-Q_RAD_OUT)

shostikk:
What is the test you are looking at?

mcgratta:
Look at `Verification/Sprinklers_and_Sprays/particle_heating_radiation.fds`. If this were a small cylinder instead of sphere, and I changed the SHAPE_FACTOR to 1/pi or something different than 1/4, I do not see how the energy would balance in terms of QR_W and Q_RAD_IN and Q_RAD_OUT. Note that I have already changed the code, so the SHAPE_FACTOR is no longer in the KAPPA_PART formulation, but even if it were, I do not see how we could achieve an energy balance.

shostikk:
Did you see the mismatch in some of the DEVC outputs? That is, should some of the DEVC outputs match, or is the verication based on the analytical temperature, as in the guide?

mcgratta:
Let me create a simple case that demonstrates.

shostikk:
I see the problem in the code, though.

mcgratta:
My test case is hard to formulate because in these ADIABATIC cases with RADIATION=T, it is difficult to maintain perfect exterior boundary conditions. In any case, I'd like you to just look at the formulation.

usfs-fds:
For what it's worth, I find that the absorption coefficient, kappa, for a single subgrid, perfectly absorbing, sphere in a bulk volume, Vb, is  kappa = (surface area of sphere)/4/Vb. For a collection of subgrid spheres, all assumed to subjected to the same intensity (direction and magnitude) this becomes  kappa = [(surface area of sphere)/(volume of sphere)] * [(volume of all spheres)/Vb]. I think this is consistent with what Kevin is writing.

shostikk:
I created a test case that demonstrates this problem: particle_absorbing_radiation.fds
DEVC 1 and 2 should be the same
DEVC 3 and 4 should be the same
DEVC 5 and 6 should be the same
[particle_absorb_rad.pdf](https://github.com/firemodels/fds/files/4856676/particle_absorb_rad.pdf)

shostikk:
[particle_absorbing_radiation_isotropic.fds.txt](https://github.com/firemodels/fds/files/4857755/particle_absorbing_radiation_isotropic.fds.txt)


There are multiple issues here:
1) WHen using wall enthalpy for cartesian (plate) particles, the area is counted twice, as we seen from line 8682-84 of dump.f90:
         SELECT CASE (SF%GEOMETRY)
            CASE DEFAULT
               VOL = ONE_D%AREA*(ONE_D%X(I)-ONE_D%X(I-1))
Therefore, the wall enthalpy output over there is two times too high for the plate. 

2) Calculation of radiation interaction with particles assumes that the particle is as efficient absorber / emitter in all directions. Only particle for which this is actually true, is a sphere. Therefore, the only ratio of true and cross sectional areas is 0.25. Making the calculation more precise would require definition of the particle orientation, and calculation of QR_W term using the actual integrals of orientation_vector \dot intensity vector.

3) Surface emissivity is not taken into account. I think it should be taken into account in KAPPA_PART. Would also need allocation of reflected enerty into scattering term.

mcgratta:
1. For plate particles, area is counted twice, but the sum of (x_i-x_(i-1)) is only half the true thickness of the plate. So I think that two wrongs make a right here. Do you agree?
2. In most particle cases, we assume that the particles have random orientation. Under this assumption, do you agree that only 0.25 as a "shape factor" in the expression for KAPPA_PART is appropriate?
3. I'll look into this.

shostikk:
1. I think this sum takes into account the whole thickness.
```
DO I=1,NWP
         VOL = ONE_D%AREA*(ONE_D%X(I)-ONE_D%X(I-1))
ENDDO

```
2. At the moment, yes I think 0.25 is the appropriate shape factor.
3. Thanks, I try to follow.

mcgratta:
For a plate particle, the THICKNESS is really the half thickness, and the gridding goes from the surface to the middle.

shostikk:
Aha. Can you help me find where that setting is done?

Which particles is this code in wall.f90 for?
```
         CASE(-1) ! Particle "backside conditions" are assumed to be from the same gas cell

            TMP_BACK  = ONE_D%TMP_G
            DTMP = TMP_BACK - ONE_D%TMP_B
            HTCB = HEAT_TRANSFER_COEFFICIENT(DTMP,SF%H_FIXED_B,SURF_INDEX,PARTICLE_INDEX=PARTICLE_INDEX)
            Q_CON_B = HTCB*DTMP
            Q_RAD_IN_B  = ONE_D%Q_RAD_IN

```

mcgratta:
Very good question. Let me look into this. I do not know why we need these
lines if we assume a plate particle is symmetric.

>

mcgratta:
Usually, a particle plate is assumed to have an INSULATED back boundary
condition, because the "back" is really the middle of a symmetric plate.
The backside conditions for a particle that you write above are for EXPOSED
back conditions. I guess this is when you have a plate particle with an
ORIENTATION.

```
   SELECT CASE(GEOMETRY)
      CASE('CARTESIAN')
         SF%GEOMETRY       = SURF_CARTESIAN
         IF (SF%WIDTH>0._EB)                 SF%BACKING = INSULATED
...
```

>

shostikk:
I cannot find this change of THICKNESS and enforcement of 'INSULATED' backing.

mcgratta:
The lines are from read.f90

>

shostikk:
Ah. It was my input mistake. I ALWAYS forget this thing that USER must input only half-thinckness for the cartesian particles. 

Anyway, now the isontropic radiation case looks like this. 0.25 gives consistent energies. 
[particle_absorbing_rad_iso.pdf](https://github.com/firemodels/fds/files/4858406/particle_absorbing_rad_iso.pdf)

mcgratta:
Great. Is everything OK now? Thanks.

shostikk:
Not sure. The unisotropic case wasn't as good. I will check tomorrow.

shostikk:
Both cases (particle_isotropic_radiation and particle_unisotropic_radiation) are working fine with 0.25

Tried adding these to Ver dataplot input, but had some problem with d1 columns. Could not yet figure out what. There is open [pull request](https://github.com/firemodels/fds/pull/8539) on this.

PR #8492, opened 2020-06-25
Source Level Set

ruddymell:
Corrected errors in computation of Rothermel RoS parameters. Improved clarity. Added namelist parameters for user defined vegetation. Added level set mode 5 (uses static wind to determine RoS but allows buoyancy generated flow).


mcgratta:
Are there any input files that use VEG_LSET_ROS? If so, we have to change them.

Source/dump.f90
```diff
@@ -3030,7 +3030,7 @@ SUBROUTINE INITIALIZE_DIAGNOSTIC_FILE(DT)
       IF (SF%VEG_LSET_IGNITE_T<1.E6_EB) &
       WRITE(LU_OUTPUT,'(A,ES9.2)')  '     Ignition Time (s)           ', SF%VEG_LSET_IGNITE_T
       WRITE(LU_OUTPUT,'(A,ES10.3)') '     Burn Duration (s)           ', SF%BURN_DURATION
-      WRITE(LU_OUTPUT,'(A,ES10.3)') '     Rate of Spread (m/s)        ', SF%VEG_LSET_ROS
+      WRITE(LU_OUTPUT,'(A,ES10.3)') '     Rate of Spread,R00 (m/s)    ', SF%VEG_LSET_ROS00
```
rmcdermo, line 3033: Can this be VEG_LSET_ROS_0?

Rate of Spread, R_0 (m/s)

mcgratta:
I originally intended VEG_LSET_ROS to be the no-wind, no-slope spread rate.
What does ROS00 mean?

>

Source/dump.f90
ruddymell: The double zero, _00, denotes zero wind, zero slope. The zero wind, zero slope, rate of spread shows up in a number of models, the Rothermel model being one. Papers have the _00 notation. For this reason, I'd like to keep the _00 notation.

ruddymell:
For clarity and direction connection to spread rate models (such as the Rothermel model), it's best to have the rate of spread input for zero wind, zero slope, be explicit. This makes it clear in the input file when a head fire rate of spread is to be derived (in the code) from the zero wind, zero slope, rate of spread (by multiplying it by a factor) or the head fire rate of spread is directly specified by the user. This makes the code easier to follow. This is being done in preparation for adding additional models for the rate of spread.

ruddymell:
I didn't find an input file in the repo where VEG_LSET_ROS is defined (I'll look again). I think they all specify VEG_LSET_FUEL_INDEX, which doesn't require that VEG_LSET_ROS is also specified. Note, that with these recent changes the user only has the choice of entering VEG_LSET_ROS00, VEG_LSET_ROS_HEAD, VEG_LSET_ROS_BACK, VEG_LSET_ROS_FLANK.

Source/dump.f90
rmcdermo: OK, thanks.  Can we change to ROS_00 (underscore), add code comments, and add space to output `Rate of Spread, ROS_00`?

PR #8495, opened 2020-06-25
6/25/20 Source: Correct error in LS1 (BT10m_2x2km_LS) when not using …

usfs-fds:
…Anderson fuels models.


mcgratta:
These changes need to be documented in the User's Guide

PR #8502, opened 2020-06-26
FDS Source: Treat thermally-thick and thin solids the same

mcgratta:


mcgratta:
The surface temperature (TMP_F) is set the same way whether NWP=1 or NWP>1.

PR #8509, opened 2020-06-29
Add compilation flag required by GCC 10

rouson:
`gfortran` 10.1.0 enforces argument type matching on calls to external procedures.  Compiling with MPICH requires passing the new `-fallow-argument-mismatch` flag to downgrade a resulting compile-time error to a warning (see the [gfortran online documentation]).  This pull request adds that flag in all GNU builds.

[gfortran online documentation]: https://gcc.gnu.org/onlinedocs/gfortran/Fortran-Dialect-Options.html

### DIFF
--- a/Build/makefile
+++ b/Build/makefile
@@ -351,7 +351,7 @@ mpi_xlf_aix : $(obj_mpi)
 
 #*** GNU Compilers ***
 
-mpi_gnu_linux_64 : FFLAGS = -m64 -O2 -std=f2008 -ffpe-summary=none -fall-intrinsics $(GITINFOGNU) $(GNU_COMPINFO) $(FFLAGSMKL_GNU_OPENMPI)
+mpi_gnu_linux_64 : FFLAGS = -m64 -O2 -std=f2008 -ffpe-summary=none -fall-intrinsics -fallow-argument-mismatch $(GITINFOGNU) $(GNU_COMPINFO) $(FFLAGSMKL_GNU_OPENMPI)
 mpi_gnu_linux_64 : LFLAGSMKL = $(LFLAGSMKL_GNU_OPENMPI)
 mpi_gnu_linux_64 : FCOMPL = mpifort
 mpi_gnu_linux_64 : FOPENMPFLAGS = -fopenmp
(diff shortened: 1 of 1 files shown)


gforney:
the  option, -fallow-argument-mismatch, option is not recognized by our
version 9 gfortran compiler on a Linux system.  I get the following error

gfortran: error: unrecognized command line option
‘-fallow-argument-mismatch’; did you mean ‘-Wno-argument-mismatch’?

does  -Wno-argument-mismatch  do what you want

marcosvanella:
Hi Damian, thank you for the contribution. Seems adding this flag will break compilation for gcc 9 and presumably lower releases. This is a problem. Does this MPICH issue also extend to other MPI distributions? 

Note we use gcc 9 + openmpi on Mac OSX and have that compile option also on our linux cluster.

marcosvanella:
Damian,
I setup an Ubuntu VM with gcc/gfortran 10 and compiled mpich3.3.2 with these. The argument mismatched error is fixed configuring mpich with the flag you used. My configure line is:
`./configure FC=gfortran-10 FFLAGS=-fallow-argument-mismatch CC=gcc-10 --prefix=/usr/local/mpich3.3.2 --enable-mpirun-prefix-by-default --enable-mpi-fortran --enable-static --disable-shared`

The issue is tied to the MPICH distribution, when compiling FDS with this library, mismatched arguments are reported for MPI routines calls (reduce, gather, send/recv). 
I also compiled openmpi3.1.6 with gcc/gfortran 10. Configure line:

`./configure FC=gfortran-10 CC=gcc-10 --prefix=/usr/local/openmpi3.1.6 --enable-mpirun-prefix-by-default --enable-mpi-fortran --enable-static --disable-shared`

In this case there is no need for this argument mismatch flag, which tells us the problem is related to MPICH. I would recommend using openmpi, as it is also the library of choice when intel mpi is not available (i.e. MacOS Bundle).

Either way, @gforney implemented the logic to detect if gfortran-10 is being used at compilation time and add the -fallow-argument-mismatch flag. This way things will not break for the rest of the gnu compiler users. Do an update on your repo and try the functionality.

I'm going to close this PR now.

PR #8524, opened 2020-06-30
FDS Source: Minor fix to char pyrolysis

mcgratta:


Source/wall.f90
```diff
@@ -3587,7 +3587,7 @@ SUBROUTINE PYROLYSIS(N_MATS,MATL_INDEX,SURF_INDEX,IIG,JJG,KKG,TMP_S,TMP_F,IOR,RH
 
       ! If there is char oxidation, save the HRR per unit volume generated
 
-      IF (ML%NU_O2_CHAR(J)>0._EB) THEN
+      IF (ML%NU_O2_CHAR(J)>0._EB .AND. SIMPLE_CHEMISTRY) THEN
```
rmcdermo, line 3590: I don't understand why char oxidation should only work with SIMPLE_CHEMISTRY.

mcgratta:
It doesn't only work with SIMPLE_CHEM. The term in the IF statement should only be set when we are using SIMPLE_CHEM. It is just a source term for the hrr output file that adds the oxidation of char to the HRR that one would measure in an oxygen consumption calorimeter. In SIMPLE_CHEM, we assume that the AIR contains the oxygen. In my test cases, I explicitly declare OXYGEN as a species and do not declare any reactions. That will break this term.

Issue #8360, opened 2020-05-24
3D pyrolysis model does not work for a special case


rmcdermo:
Hi

In my pyrolysis model, SURF goes to two MATLs: 1. complete evaporation of volatiles 2. a solid ... however, is a charring case (does not burn away at the end)
The case runs fine with the 1D PYRO model.

A) When I activate the 3D PYRO model with HT3D=T, FDS asks for turning on BURN_AWAY (maybe because of MATL 1), which does not represent the charring physics.
what did I do wrong?

B) Let's activate BURN_AWAY to get rid of the error and run sth on 3D PYRO (add BURN_AWAY=T to 3D.fds)... no pyrolysis reaction happens!
any suggestion to fix this problem?

TNX

[3D.fds.txt](https://github.com/firemodels/fds/files/4673449/3D.fds.txt)
[1D.fds.txt](https://github.com/firemodels/fds/files/4673450/1D.fds.txt)

MortezaHaghighi:
Randy,

I think it's enough to change the place of this (commented out) IF statement on the read.f90 and put it outside the DO loop:

DO NR=1,ML%N_REACTIONS
   IF (ABS(SUM(ML%NU_RESIDUE(:,NR)))>TWO_EPSILON_EB) PYRO3D_RESIDUE=.TRUE.
ENDDO
       **!IF (.NOT.PYRO3D_RESIDUE .AND. .NOT.OB%CONSUMABLE) THEN
           !   WRITE(MESSAGE,'(A,A,A)') &
           !      'ERROR: MATL ',TRIM(ML%ID),', PYRO3D requires residue (NU_MATL) or BURN_AWAY'
           !   CALL SHUTDOWN(MESSAGE,PROCESS_0_ONLY=.FALSE.) ; RETURN
        !ENDIF**
ENDIF        
OB%RHO(:,:,:,NNN) = ML%RHO_S ! TEMPORARY -- must be reinitialized after PROC_WALL is called
ENDDO
                     
      **IF (.NOT.PYRO3D_RESIDUE .AND. .NOT.OB%CONSUMABLE) THEN
      WRITE(MESSAGE,'(A,A,A)') &
      'ERROR: MATL ',TRIM(ML%ID),', PYRO3D requires residue (NU_MATL) or BURN_AWAY'
       CALL SHUTDOWN(MESSAGE,PROCESS_0_ONLY=.FALSE.) ; RETURN
       ENDIF**

MortezaHaghighi:
[...]

PR #8526, opened 2020-06-30
Fix minor problems regarding issue #8360

MortezaHaghighi:


MortezaHaghighi:
https://github.com/firemodels/fds/issues/8360

PR #8539, opened 2020-07-02
FDS Verification: Add particle radiation cases to Ver script

shostikk:


shostikk:
For some reason I get errors, so it is not yet working well.

Error: Problem with dataplot row 319 (particle_isotropic_radiation); check syntax of analytical/expected/experimental (d1) columns. Skipping case.
Error: Problem with dataplot row 320 (particle_unisotropic_radiation); check syntax of analytical/expected/experimental (d1) columns. Skipping case.

rmcdermo:
Looks like you have repeated the _devc file where the "expected" data file should be.  Also, I do not see that you have committed something like "particle_isotropic_radiation.csv".

shostikk:
The verification compares two different DEVC outputs, which should be the same. There is no exact result (or at least I haven't calculated one)

rmcdermo:
OK, well, I'm just going to merge it and sort it out from there (easier than testing a separate branch).

rmcdermo:
I am editing things, I think I see the problem, please don't touch for now.

PR #8543, opened 2020-07-02
FDS Verification: fix issues with particle radiation dataplot entries

rmcdermo:
In this PR,
* d1 entries for dataplot need to be only one x and y quantity in order to handle statistics
* rename input file
* fix duplicate DEVC name


rmcdermo:
@shostikk Things should be working now.  But note that the "ball" cases are out of tolerance.  Please have a look at these.  You need to double check the `FDS_Verification_Guide/SCRIPT_FIGURES/Scatterplots/verification_statistics.tex` file to see whether firebot will pass.

shostikk:
Thanks @rmcdermo. I will look at the tolerance, once I get the whole verification suite run. 

Also, I will add the remaining temperature comparisons for the particle_(an)isotropic_radiation. Outputs are there, but not yet in the dataplot.

PR #8569, opened 2020-07-09
FDS build: add compiler option to Intel debug linux target so unused …

gforney:
…variable will be flagged


gforney:
firebot was ok.  the problem was a missing compiler option in the intel linux debug target

PR #8581, opened 2020-07-10
FDS Tech Ref: Fix equation formatting

drjfloyd:


mcgratta:
The write-up is really nice. I think it helps alot to explain the routine.

>

drjfloyd:
Thanks, I can't take all the credit as it is mostly from the paper Randy and I wrote for IAFSS.

PR #8588, opened 2020-07-16
FDS Source: apply diffusion term corrections for all simulation modes

rmcdermo:


rmcdermo:
The inclusion of Jason's merge commit is part of my reset to fix Susan's PR.

PR #8602, opened 2020-07-20
FDS Source: Move gas H_HEAT and H_MASS to a subroutine in func.f90

drjfloyd:


mcgratta:
Jason -- could you add the Doxygen like comments to these routines. I still have to add comments to other routines in func.f90, but at least I'll understand what these do. Units are especially important.

drjfloyd:
Sorry I forgot. I'll add.

PR #8616, opened 2020-07-21
FDS Source: Avoid endless loop in 1-D heat transfer routine

mcgratta:


mcgratta:
Jason -- take a look at this. The validation cases called SP_Wood_Cribs all hung up in SOLID_HEAT_TRANSFER_1D in a search loop with no backdoor exit.

drjfloyd:
I'll take a look at it

mcgratta:
The cases are working now. I added a backdoor. But just see if my max number of iterations is OK

drjfloyd:
That number should be fine.

drjfloyd:
I modified the backdoor to average the prior and current temperature. This is what is done elsewhere (see 1498 in hvac.f90). I checked this case and it is bouncing around between 782.9932 and 782.9938.

PR #8663, opened 2020-08-10
FDS Verification: Add Error statements to scaling_tests

mcgratta:


mcgratta:
Assuming that firebot picks up on the word "Error" in a Matlab script, these error statements should catch cases when the MPI scaling test results are way out of whack.

rmcdermo:
Based on some of the other Matlab scripts, I think we expect firebot to grep on both "Error" and "Matlab Warning", so you should be good.  Thanks!

gforney:
firebot greps on "Error", "Out of Tolerance" and  "Matlab Warning" in the matlab verification step and "Error" in the matlab validation step

Issue #8666, opened 2020-08-12
Pressure drop in still compartment


mcgratta:
I can't figure out why the pressure near the floor drops about 3 Pa in this case:
```
&HEAD CHID='leak' /
&MESH IJK=26,26,26, XB=0.0,26,0.0,26,0.0,26.0 /
&TIME T_END=50.0 /
&SURF ID='ROOF' /
&ZONE XYZ= 12,12,12 /
&VENT ID='IN', SURF_ID='ROOF', XB=0.0, 26, 0.0, 26, 26, 26, COLOR='BLUE' /
&HVAC ID='LEAKAGE', TYPE_ID='LEAK', VENT_ID='IN', VENT2_ID='AMBIENT', AREA=1.0 /
&DEVC ID='Pres', QUANTITY='PRESSURE', XYZ=13,13,0.5 /
&TAIL /
```
If you remove the leak in the roof, there is no pressure drop. Is there some assumption about the height of the AMBIENT vent?

drjfloyd:
I'll take a look. Local leakage uses the local pressure rather than just ZONE pressure so it includes the effects of any local velocity. Initial noise might result in some initial flow due to some random net flow to or away from the VENT, but I wouldn't expect to see such a large pressure drop.  The AMBIENT vent should have the same elevation as the average elevation of the VENT but maybe I am doing something wrong for IOR=-3/3.

drjfloyd:
I figured out the problem. The underlying issue was the RAMPs for temperature and pressure.   The RAMPs span from ZS_MIN to ZF_MIN, in this case 0 to 26. In the HVAC routine when we go to get the domain pressure at a horizontal vent we do it as 0.5*(PBAR(KK) + PBAR(KK+1)). With the VENT on the ceiling the center of the node KK+1 is > 26 but the RAMP tops out at 26 so we don't have the right pressure.  I need to adjust how we determine the span for the RAMPs when we have HVAC present.

drjfloyd:
Should be working now. There can still be a slight initial flow depending on grid resolution since the pressure inside the domain is done by averaging PBAR over K and K+1 but the ambient node pressure is done by using the average Z with a RAMP lookup. With the fix there is now a few cc of air that flow out and the initial pressure drops ~0.03 Pa.

mcgratta:
Yup, works, thanks.

PR #8667, opened 2020-08-13
FDS Source: Fix initialization of localized leakage ambient nodes Iss…

drjfloyd:
…ue #8666


mcgratta:
This does not need to be included in the current maintenance release.

PR #8672, opened 2020-08-14
This PR prevents validation cases from being run in a dirty repo or by a dirty executable unless you override (say while testing cases before being run for real)

gforney:


rmcdermo:
Glenn,
Dirty repo is ok.  We will almost always be running with something modified when we do validation runs.  The input files need to be clean, not the whole repo.

PR #8689, opened 2020-08-20
FDS Validation: add more outputs for UMD line burner methane case

rmcdermo:


mcgratta:
Did we get new data for these experiments, or did we miss it previously?

rmcdermo:
James White added the centerline TC profiles and RMS data for the "y" profiles sometime last year.  I was reminded of it because I got pinged on one of Bart Merci's papers where they had compared to this new data.  I want to make sure we are up to date.  I am not going to process the data until after this release.

Also, I want to create a verification case (unless you know of one that exists) for RMS output when the DT_LINE interval is in the middle of the simulation.

mcgratta:
OK, no I don't have a verification case like that, but this should now work with the new stats features. Let me know if it doesn't.

PR #8690, opened 2020-08-20
FDS Source: Allocate and exchange geometry arrays earlier in startup

mcgratta:


mcgratta:
This change was needed to allocate and exchange the arrays WALL_INDEX, CELL_INDEX, OBST_INDEX_C, and SOLID earlier in the start-up process because they are now used in the setting up of WALL boundary types. We need to know about the geometry of neighboring meshes to decide what external WALL BCs to apply. This change just allocates and exchanges these arrays earlier and does not change any of the logic.

PR #8698, opened 2020-08-24
FDS Source: Prevent hang if one MPI process fails during startup.

mcgratta:


mcgratta:
This bug fix is important for all. When we implemented MPI, we would do all MPI calls in main.f90 to ensure that all MPI processes where "back" from their various subroutines when the calls were made, especially blocking calls or collective calls that require all MPI processes to be present. As time has gone on, I've introduced some MPI calls directly in the startup and read routines. That is OK as long as all MPI processes run through the same code in the start-up routines. But more and more we have to limit some code only to those MPI processes that control a particular mesh. For example, only processes that control a particular mesh will read OBSTs in that mesh. That opens up the possibility that a process will hit an ERROR, write a MESSAGE, and then RETURN, but all other processes will continue on through the set up, and if these processes hit an MPI call, things will hang because one process has already returned and will never arrive at the MPI call. 

This is what happened here. There is a loop in PROC_HVAC that is restricted to only those MPI processes that control a particular mesh. There are ERROR statements in the loop, along with RETURNs. Then after the loop, a collective MPI call. So in this instance, one MPI process (process 1) hit an error, returned, and never arrived at the MPI_ALLREDUCE. I fixed it (since I broke it in the first place) and I'll look for other situations like this. Be on the lookout as well. Best to put MPI calls in main.f90 in places where we know all processes must pass. But occasionally there is a need to put the MPI call outside of main, in which case we have to be careful.

PR #8704, opened 2020-08-25
Matlab: Use unicode characters for dataplot inputs

mcgratta:


mcgratta:
All -- I converted various symbols in `FDS_verification_dataplot_inputs.csv` to be unicode. Matlab uses the UTF-16 character set, and some of the previous symbols like \circ and various other TeX symbols are no longer handled properly, which is why when we run the Matlab scripts in linux we get odd spacings. I want to see if git and firebot are OK with this before changing other files. If you use Excel to edit this file, the symbols will look natural, that is, x^2 will look like it should. The degree symbol, etc. If you want to introduce new symbols in a plot, test it before committing.

PR #8717, opened 2020-08-28
FDS Source: updates to BLOWING model, use DN as length scale for Cart…

rmcdermo:
…esian geometry, include low Re correlations for Sh and Nu

### DIFF
--- a/Source/wall.f90
+++ b/Source/wall.f90
@@ -3423,6 +3423,7 @@ MATERIAL_LOOP: DO N=1,N_MATS  ! Tech Guide: Sum over the materials, alpha
             IF (ML%H_R_I(1) > 0) THEN
                CALL GET_EQUIL_DATA(SPECIES_MIXTURE(SMIX_INDEX)%MW,TMP_F,PBAR(KKG,PRESSURE_ZONE(IIG,JJG,KKG)),&
                                          H_R,H_R_B,T_BOIL_EFF,X_W,RAMP_INDEX=ML%H_R_I(1))
+
             ELSE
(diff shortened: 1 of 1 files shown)


drjfloyd:
The two correlations are flat plate correlation where Nu_L or Sh_L is defined using L from the edge of the plate. Using DN doesn't match the basis for the correlation. SF%LENGTH I think was correct but for a fixed pool the average distance from the edge is 1/3 r for a circular pool.

rmcdermo:
I understand.  This is temporary, but works still for the only case that uses BLOWING, which is the water evap case.

In the Rayleigh model for HTC, which is essentially a wall function, we take a Nu_x correlation and change x with distance from the surface.  This works rather well, and only needs an adjustment of the coefficients.  The form of the correlation is the same.  I do not see a way in which these large scale correlations converge to a DNS unless the distance form the wall is part of the formula.  As I have it coded now, things are looking better.  I want to run a series to see where things stand with the other cases.

drjfloyd:
The water evap case is SPHERICAL which uses a correlation based on radius so your change didn't impact it.  For a drop it is pretty easy to get the free stream velocity and vapor concentration. As long as the drop size is small compared to the grid cell, the drop relative velocity and cell gas concentration is what we need. The plate correlations are also based on free stream conditions, but as we make the grid cell smaller, the velocity in that cell at the wall is going to vary along with the gas concentration. Maybe we need another approach altogether for CARTESIAN?

rmcdermo:
I'm not talking about the drop cases.  I'm talking about the ASTM case in Pool_Fires---it is a pan.  Yes, my thinking is that Cartesian surfaces (pools) would be treated differently than droplets for the "correlations" or "wall functions" or whatever.  We can easily keep it so that if someone specifies a convective length scale, then it can be used.  I just did not go that far yet with the implementation.

PR #8725, opened 2020-08-31
FDS Source: Shut down code if MPI timeout detected

mcgratta:


mcgratta:
This commit should cause all processes to end gracefully if there is an MPI timeout.

PR #8730, opened 2020-09-05
FDS Source: Create MPI communicators for each mesh neighborhood

mcgratta:


mcgratta:
This commit cleans up warnings from the option -check_mpi in debug mode. This option looks for errors in the MPI calling routines. The most persistent warning was about MPI persistent sends/receives in which arrays like `M%SOLID` are send from each mesh to its neighbors. This is not perfectly kosher because if the send buffer is the same for multiple sends, the MPI standard does not guarantee that the buffer won't be tampered with.

PR #8747, opened 2020-09-16
Correction of non-manifold geometries in V&V (from BlenderFDS V&V)

emanuelegissi:
Several validation cases had the same `GEOM` line representing an ico-sphere. I shared that line in a `sphere_geom.dat` file called by a `CATF` line. I do not think the dat file deserves being moved to the `cad` repository, being so simple. While at that, I corrected the sphere `GEOM` removing many loose vertices.
I did not replace the `GEOM VERTS FACES` sphere, with a `GEOM SPHERE`, because I did not want to influence the verification results.

The arch `GEOM` in the `geom_arch.fds` validation case had many loose vertices. I removed them.

Loose `VERTS` are vertices not used by any of the `FACES` of the `GEOM` line.
They are currently not detected by FDS. At the moment they do not seem to adversely influence the cut-cell algorithm.


rmcdermo:
@emanuelegissi @marcosvanella Do we even need a sphere_geom.dat file?  Why not just use the internal code to create the sphere for these cases?

Do we actually need geom_arch?  This was a very old case from the days of Charles Luo working with us.  I'd rather purge these things if they are not getting used.

marcosvanella:
@emanuelegissi thank you. There shouldn't be an issue with unconnected nodes, as long as the triangulation is correct. Things are done on a triangle basis. We don't have a test for unconnected nodes in IBM_INIT_GEOM, we should add it and throw a warning. 

I think we can switch to the analytical sphere definition on these cases:

&GEOM ID='SPHERE',SURF_ID='SPHERE',SPHERE_ORIGIN=0.0,0.0,2.0,SPHERE_RADIUS=1., N_LEVELS=3/ 

About the arch, at this point is being used to make an SMV plot that is used in the guides to show a more complicated geometry made with VERTS and FACES. We can either get rid of it, or move the arch to a .dat file to exercise CATF with a geometry in firebot.

I'm going to close this pull request and we can discuss these changes further.

emanuelegissi:
Ok

PR #8751, opened 2020-09-18
FDS Source: Automatically assign PRESSURE_ZONEs

mcgratta:


mcgratta:
This PR eliminates the need to declare pressure ZONEs. Now the code will automatically figure out where the ZONEs are. However, if you are leaking from one ZONE to another, you still have to declare the ZONEs. 

Current input files still work. If you declare a few ZONEs and the code finds more, it will automatically re-allocate the P_ZONE derived type array and add the new ZONEs. 

Eventually, say with FDS 7, we might be able to eliminate the ZONE line altogether. I believe, and Jason tell me if I'm wrong, that leaks are actually done via HVAC machinery in the background. The leakage parameters on the ZONE line are maintained for backward compatibility. I should be able to just tack some VENTs to various locations and hook an HVAC system to it without having to declare a ZONE line.

I have checked the verification cases and they are working. I'm now going to gradually remove ZONE lines that do nothing more than indicate a pressure zone. I'll leave the HVAC stuff for later, once we're happy that the zone finding algorithm works.

Jason -- just think about how we might remove ZONE completely. Don't do anything until the dust settles on the latest change.

Marcos -- this stuff works with GEOM, but we have only a few examples. We might have to exercise GEOM stuff more.

drjfloyd:
For localized leakage we wouldn't need any changes. Bulk leakage is a different matter. Consider two adjacent rooms where one room had an alcove that bumped out into the second. If we simply used HVAC, then we would have multiple vents for each segment of the wall where each would need its own HVAC duct. That could get very tedious for complicated geometries. We may still want some kind of SURF input. Some thing like LEAKAGE_ID (or make LEAK_PATH a string). Then we could have an &HVAC TYPE of 'BULK LEAKAGE' where you give two LEAKAGE_IDs instead of two VENT_IDs and reserve the name AMBIENT.

mcgratta:
OK, just think about it for now. I'm going to eliminate ZONE lines from the verification cases where they just indicate a sealed cavity. We'll leave leakage stuff for later.

emanuelegissi:
Amazing work! This is going to prevent a lot of headaches and requests for
assistance. Thanks

PR #8752, opened 2020-09-18
FDS Source: minor fix of GAS_PHASE_OUTPUT comment

rmcdermo:


rmcdermo:
@marcosvanella Please take notice of this.  In GAS_PHASE_OUPTUT, etc., these are not just comments, but reflect exactly what the user should input for QUANTITY

marcosvanella:
Noted, Thanks.

PR #8757, opened 2020-09-21
FDS Source: Change 'NET HEAT FLUX' to 'TOTAL HEAT FLUX'.

mcgratta:


mcgratta:
Note that 'NET HEAT FLUX' will still work.

PR #8784, opened 2020-09-30
FDS Validation Guide: Minor edits to Insulation Materials

mcgratta:


mcgratta:
Minor edits to the text, and I added the table of material properties from the paper.

PR #8786, opened 2020-09-30
Source: vege.f90 fix absence of spread in LS when there's no …

usfs-fds:
…wind

### DIFF
--- a/Source/vege.f90
+++ b/Source/vege.f90
@@ -389,7 +389,7 @@ DO JJG=1,JBAR
          PHI_W_Y = C_ROTH * ((3.281_EB * UMF_MAG)**B_ROTH) * (SF%VEG_LSET_BETA / BETA_OP_ROTH)**(-E_ROTH)
          PHI_W_Y = PHI_W_Y*UMF_Y/UMF_MAG
 
-         IF (UMF_MAG /= 0.0_EB) THEN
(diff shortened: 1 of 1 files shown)


Source/vege.f90
```diff
@@ -389,6 +389,14 @@ SUBROUTINE LEVEL_SET_FIRESPREAD(T,DT,NM)
          PHI_W_Y = C_ROTH * ((3.281_EB * UMF_MAG)**B_ROTH) * (SF%VEG_LSET_BETA / BETA_OP_ROTH)**(-E_ROTH)
          PHI_W_Y = PHI_W_Y*UMF_Y/UMF_MAG
 
+         IF (UMF_MAG /= 0.0_EB) THEN
```
rmcdermo: @ruddymell please use
```
ABS(UMF_MAG)<TWO_EPSILON_EB
```
Thanks!

rmcdermo: Rather, >

rmcdermo: You can just update your fork and your PR will be updated as long as we do not merge first.

ruddymell: Thanks, I just updated the fork to have ABS(UMF_MAG) > TWO_EPSILON_EB as suggested

PR #8813, opened 2020-10-09
Data structure for 1-D mass transfer

shostikk:
I am not sure if I need to manually change something for the ONE_D allocation counts. I think not, and this seems to work now.


rmcdermo:
@shostikk please don't forget to merge

rmcdermo:
Oh, never mind, I see your review request.

mcgratta:
Simo -- why can't we just use the current MATL_COMP structure to hold the gas species? In other words, can't we hold the gas species that are within the solid in the same arrays that we hold solid? We can just assign an attribute to the gas species allowing them to be transported. Adding this new machinery in this PR is going to break the code because there are array dimension counters that need to be updated when you add new stuff to ONE_D.

shostikk:
Well, that's what I tried to ask. Are there counters that would need to be updated? I couldn't figure that out right now.

Why not MATL? Of course we can do it, but as the treatment of gas and solid massess is going to be very different, I thought it would make a clearer code having separate data structure for the gas. This way, I would not need to touch the solid part at all (probably breaking it many times on the way) + code readability. All the loops etc would have special treatments for gas, making them confusing.

drjfloyd:
Before the ONE_D type in type.f90 there are a series of counters for scalar reals, integers, and logicals.  If you add one of those you need to up the counter. 

In func.f90 there are routines that set up global storage arrays for the ONE_D type.  Easiest way to figure it out from my experience is pick a current ONE_D array or a current ONE_D scalar (depending on what you add) and find it in func.f90 to see how the setup is handled.

mcgratta:
Search on RHO_DOT in func.f90. You will see that space must be added to the
array that holds the ONE_D parameters.

>

shostikk:
I found two places that had to be updated in order to add a new array to ONE_D. I hope I got them all. Also, I think I should not put the allocation lines inside an IF-statement. N_SPEC = 0 will take care of zero allocation when there is no gas.

mcgratta:
I will run the verification cases in debug mode to see if anything breaks.

mcgratta:
Everything seems OK when I run the verification cases a few time steps in debug mode. I will check again tomorrow when firebot runs completely, but so far things seem OK.

shostikk:
Firebot gave no warnings, except need to document the namelist keywords.
Next week, I will continue data structure initialization, and we will present you a draft document of equations to be solved.

PR #8815, opened 2020-10-09
FDS Source: add PARTICLE DRAG FORCE to PART output QUANTITIES

rmcdermo:


mcgratta:
ACCEL_X etc is an acceleration in m/s2. Would it be more useful to have

LP\%MASS * LP\%PWT * LP\%ACCEL_X

rmcdermo:
Yes, that's probably a good idea.  At the moment, I'm trying to hook up diagnostics to see where things are going off the rails.  I'll be more explicit about force versus acceleration.

PR #8818, opened 2020-10-09
FDS Source: Add SIGN to TWO_EPSILON_EB in DRAG_MAX expressions

drjfloyd:


Source/part.f90
```diff
@@ -2250,9 +2250,9 @@ SUBROUTINE MOVE_IN_GAS
          LP%ACCEL_X = -(MIN(DX(IIG_OLD),LP%DX)/DX(IIG_OLD))*(K_TERM(1)+Y_TERM(1))*UBAR*SFAC
          LP%ACCEL_Y = -(MIN(DY(JJG_OLD),LP%DY)/DY(JJG_OLD))*(K_TERM(2)+Y_TERM(2))*VBAR*SFAC
          LP%ACCEL_Z = -(MIN(DZ(KKG_OLD),LP%DZ)/DZ(KKG_OLD))*(K_TERM(3)+Y_TERM(3))*WBAR*SFAC
-         DRAG_MAX(1) = LP%ACCEL_X/(-UBAR+TWO_EPSILON_EB)
-         DRAG_MAX(2) = LP%ACCEL_Y/(-VBAR+TWO_EPSILON_EB)
-         DRAG_MAX(3) = LP%ACCEL_Z/(-WBAR+TWO_EPSILON_EB)
+         DRAG_MAX(1) = LP%ACCEL_X/(-UBAR+SIGN(1._EB,UBAR)*TWO_EPSILON_EB)
```
rmcdermo, line 2253: Don't you want the opposite sign?

drjfloyd: IF UBAR is negative then SIGN(1._EB,UBAR) will return -1._EB. So we substract TWO_EPSILON_EB and won't risk making the denominator 0.

drjfloyd: argh. Wait you are correct. I messed that up.

PR #8967, opened 2020-11-19
Changed the number in J to kJ to reduce reading errors for languages …

TristanHehnen:
…where the comma acts as decimal point, e.g. German.


drjfloyd:
Thanks for the suggestion.

TristanHehnen:
You're welcome and thank you very much!

PR #8969, opened 2020-11-19
FDS Source: Add VEL_PART to SURF

drjfloyd:


mcgratta:
Remember to add VEL_PART to the User's Guide. Thanks.

PR #9007, opened 2020-12-01
added missing '/' at end of SLCF line for CSIRO

amarcozzi:
Looks like a '/' was missed at the end of line 102 in the Case_C064_BFM_crude file.


rmcdermo:
Thanks for the catch!

PR #9012, opened 2020-12-02
FDS Source: Allow one to choose CONSMAG near wall model

mcgratta:


mcgratta:
Randy, FYI, this PR does not change anything, but allows me to use the old NEAR_WALL_TURB model. I'm running into some numerical instabilities in my cabinet arc fault cases and I want to see if it's related to the change. There was no way to override the selection of WALE, which is default unless the user specifies anything else for NWTM.

rmcdermo:
Thanks.  I meant to do this but had forgotten.

PR #9039, opened 2020-12-06
FDS Utilities: fixes in obj2fds

YudjinSud:
Deleted some hard-coded parameters in order to work with .obj files of all possible sizes

### DIFF
--- a/Utilities/CAD_Converters/obj2fds.m
+++ b/Utilities/CAD_Converters/obj2fds.m
@@ -11,24 +11,25 @@ close all
 clear all
 clc
 
-file='Death_Star.obj';
+file='Restaurant2.obj';
+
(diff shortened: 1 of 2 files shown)


Utilities/CAD_Converters/obj2fds.m
```diff
@@ -62,7 +63,7 @@
 [fid]=fopen([fileout],'w');
 
 label{1} = ['&GEOM ID=''%8s'', ']; 
-for i=2:10
+for i=2:nsurf+1
```
rmcdermo, line 66: Thanks for your contribution.

Where does `nsurf` get set?

rmcdermo: OK, I see now.

rmcdermo (approved):

PR #9052, opened 2020-12-11
1D mass transfer, implement layer porosity and explicit solver for gas diffusion

shostikk:


rmcdermo:
@MortezaHaghighi @shostikk How hard would it be to keep PYRO3D updated with the 1D mass transfer development?  In other words, could you implement a simple nominally 1D verification case, similar to `pyro3d_vs_pyro1d.fds`, for example, that would help develop the 3D solver?  I'm worried about these codes diverging again.  Thanks

rmcdermo:
@shostikk I suggest you either merge this or close it and send a new PR.  We are working on issues of mass flux consistency and I'd like to push up some changes that are likely to conflict with this.

shostikk:
Plan is to make sure consistency. I copied some concepts from Pyro3D over 1D, and will continue

PR #9056, opened 2020-12-16
FDS Source: Issue a WARNING rather than ERROR if MAXIMUM_TEMPERATURE …

mcgratta:
…exceeds 5000 K


mcgratta:
FYI, this is related to the arc fault calculations that Jason and I are doing. Note that `MAXIMUM_TEMPERATURE` does not affect the gas phase temperature, which is calculated from the density with no clipping.

PR #9067, opened 2020-12-22
FDS Source: Add mixing time to cfl.csv file

mcgratta:


mcgratta:
FYI, adding mixing time to _cfl.csv helps to diagnose small time steps.

PR #9068, opened 2020-12-23
FDS Source: Minor change to print out and fix NIST_NRC validation case

mcgratta:


mcgratta:
Jason, FYI, NIST_NRC_02 is a closed door test. You  had it as open.

PR #9099, opened 2021-01-09
FDS Source: Change MINT to MASS in various variable names

mcgratta:


mcgratta:
FYI, I changed all the "MINT" variables to "MASS", as in

MINT --> MASS_INTEGRAL
MINT_SUM --> MASS_DT
MINT_TOTAL --> MASS_TOTAL

I also cleaned up some MPI_GATHERV calls, reducing the number of "REAL_BUFFERs" and giving them better names. There should be no change in functionality with these changes and I ran the verification changes to check.

PR #9100, opened 2021-01-10
FDS Source: Change MYID to MY_RANK. No change in functionality

mcgratta:


mcgratta:
This is just a name change. MYID is a bit cryptic. I borrowed it from some sample MPI program, but MY_RANK is more descriptive. RANK is the proper name of an MPI process, but I worried that RANK might be a reserved word in Fortran or MPI.

Issue #8986, opened 2020-11-23
BURN_AWAY and RESTART conflict


MortezaHaghighi:
Hi,

Stoped simulation midway burning away, followed by restarting the calculations, will end up in the reproduction of the initial complete geometry.
Based on my experience, it's not just an SMV visualization problem, but rather a PYRO3D code bug (a Jump in MLR curve seen in another heavy simulation).
The attached input file was forcefully stopped at about  350s and started again.

![BURN_AWAY_RESTART_BUG_0 0s](https://user-images.githubusercontent.com/56713983/99987821-94962680-2db9-11eb-80c8-546fb45ccf73.png)
![BURN_AWAY_RESTART_BUG_349 2s](https://user-images.githubusercontent.com/56713983/99987836-99f37100-2db9-11eb-98ad-8af4eac3877c.png)
![BURN_AWAY_RESTART_BUG_349 8s](https://user-images.githubusercontent.com/56713983/99987854-9e1f8e80-2db9-11eb-8b1f-ca85073f80b8.png)
![BURN_AWAY_RESTART_BUG_451 2s](https://user-images.githubusercontent.com/56713983/99987861-a081e880-2db9-11eb-9fb1-04c2eafc7bbe.png)

[BURN_AWAY_RESTART_BUG.fds.txt](https://github.com/firemodels/fds/files/5584350/BURN_AWAY_RESTART_BUG.fds.txt)

Thanks in advance.

rmcdermo:
Thanks for the report.  I'll have a look.

rmcdermo:
This should be fixed with PR #8990 .  Give it a try and let me know.  Good catch and nice simple test case.  Thanks.  I'm nominating this for a Boris Stock!

MortezaHaghighi:
Overall looks fine to me. However, I still see a minor jump on the MLR curve while restarting. For example, the following figure is a result of restarting at 345 s. Other restarting times found to reproduce such a jump as well.

![MLR](https://user-images.githubusercontent.com/56713983/102762686-4c9afd00-4381-11eb-8e68-c2aad68d50b8.jpg)

rmcdermo:
I'm trying to track it down.  The only clue I have so far is that if I set DT_RESTART to coincide with with T_END, then we do not get the kink.  So, I suspect a time-related problem, averaging, or something.

rmcdermo:
This should be fixed with PR #9069.  Please confirm.

[...]

PR #9107, opened 2021-01-14
Verification case for PYRO3D restart

MortezaHaghighi:
Verification case for issue #8986


rmcdermo:
Thanks @MortezaHaghighi .  I'll compile things and make sure it works

PR #9120, opened 2021-01-18
FDS Source: Ensure compatibility with Fortran 2018

mcgratta:


mcgratta:
FYI, we now check for compatibility with Fortran 2018 standard. I removed:

1. EQUIVALENCE statements, meaning some outdated inputs are now gone permanently.
2. DSIN, MAX0, FLOAT, etc, are now replaced by their modern counterparts.
3. DO loops using numbers, like DO 570 J=1,10   ....  570 CONTINUE

Surprisingly, GOTOs did not make the cut. They are still tolerated, but probably not for long. There are old bits of code in turb.f90 that still use GOTOs.

I also removed some compiler options that no longer seem relevant or listed.

rmcdermo:
OK, I will work on turb.

mcgratta:
There are no warnings about the gotos. What I meant to say is that Fortran 2018 still does not consider them obsolescent. You don't have to do anything to code at this stage. I cleaned up all the warnings.

PR #9133, opened 2021-01-21
FDS Source: implement TEST_NEW_WAKE_REDUCTION

rmcdermo:


mcgratta:
KAPPA_E has dimensions of 1/m. Do you want to multiply it be the dimensionless drag coefficient?

rmcdermo:
I was going off the equations you have in the tech guide, 9.7, 9.8, 9.9.

rmcdermo:
Maybe I'm double counting an AREA?

mcgratta:
Yes, I think so. Follow C_DRAG through the part.f90 routines until you see
it used in Eq 9.1. Eq 9.7 is a rewriting of 9.1 in terms of WUI parameters,
but in FDS we still invoke Eq 9.1. That is, we apply the drag coefficient
to each particle.

>

PR #9139, opened 2021-01-22
FDS Source: fix drag output for screens and porous

rmcdermo:


rmcdermo:
@drjfloyd I pushed up the contracted version of C_DRAG.  Let me know if you want something different.

PR #9149, opened 2021-01-26
FDS Source: Decoupling fire and evacuation calculation: Just fire or …

tkorhon1:
…evacuation meshes present, evac version 2.6.0

Now there are only either fire of evacuation meshes present in a calculation. The evacuation calculation has three phases: 1) evacuation initialization => CHID_evac.xyz file 2) fire calculation that reads CHID_evac.xyz file and writes CHID_evac.fed file and does the fire calcualtion, 3) evacuation calculation that reads the CHID_evac.eff and CHID_evac.fed files. If just an evacuation drill is wanted, then this can be done in phase 1. And it can be done also in phase 3 (EVACUATION_MC_MODE=t and EVACUATION_DRILL=t).

What is not done (yet): Cleaning up the main.f90. Now there are just fire or evacuation meshes present, so one could simplify the logic somewhat. Update the evacuation manual (+readme-file) for these changes.


tkorhon1:
Oh! Now I must also edit the evacuation examples in the verification/evacuation folder. This takes a little time, so you (at NIST) could just "comment out" the evacuation verification tests for now (the every night run scripts).

rmcdermo:
I do not see an Evacuation cases in FDS_Cases.sh

PR #9153, opened 2021-01-27
FDS Source: 10 pressure iterations for SVLES

mcgratta:


mcgratta:
FYI, I am just making 10 pressure iterations the default for all modes. The low value for SVLES was causing trouble in some cases.

PR #9161, opened 2021-01-29
FDS Validation: Allow low density for charring cases

mcgratta:


Validation/SP_Wood_Cribs/FDS_Input_Files/SP_Test_9.fds
```diff
@@ -7,6 +7,8 @@
 
 &TIME T_END=720. /    
 
+&CLIP MINIMUM_DENSITY=0.05 /  Account for high temps due to charring
```
rmcdermo, line 10: @mcgratta If this is typical for charring, should we just make it default?  I worry about needing to add this to every WUI case.

mcgratta:
One out of the 12 SP_Wood_Crib cases had an instability due to hitting the lower limit. I suggest we look at more validation cases and then decide.

Issue #8230, opened 2020-04-09
2-D cylindrical application of PYRO3D


rmcdermo:
Hi folks,

I was trying to simplify a 3D pyrolysis case (set up by HT3D) to the 2D axisymmetrical model when I faced that I cannot pyrolyze the material in 2D!
It seems that 2D cylindrical codes are developed for the HT3D and PYRO3D routines in the source files (If I'm not misinterpreting the codes!). However, I couldn't apply it to my case. The material heats up but it does not pyrolyze.
I would appreciate it If someone can take a look at the attached input file to see whether I'm missing anything for an axisymmetric simulation (the same arrangement of input file works fine in a 3D cartesian setup).
TEFZEL density slices should show the pyrolyzate gas in case of something happening.

[2D_PYRO3D.fds.txt](https://github.com/firemodels/fds/files/4457257/2D_PYRO3D.fds.txt)

JLUrban:
I added ", PYRO3D_IOR=1" at the end of the "&OBST" line and see pyrolsate production at around 64 seconds. I also tried using PYRO3D_IOR=0 (which I believe is the default: pyrolysate sent to nearest wall surface) and I did not see any pyrolsate in the slice file.

rmcdermo:
I guess as a work around right now you can do this:
```
&OBST XB=0.0,0.1,-0.01,0.01,0.48,0.50, SURF_ID='COMBUSTIBLE_MATL', HT3D=T, PYRO3D_IOR=3/
&OBST XB=0.0,0.1,-0.01,0.01,0.00,0.48, SURF_ID='COMBUSTIBLE_MATL', HT3D=T, PYRO3D_IOR=1/
```

@mcgratta Can you have a look at the `ASSIGN_HT3D_WALL_INDICES` function and see if you can sort out why it would not assign wall cells for `PYRO3D_IOR=0`?  Thanks

Try with the attached input file below and print out `M%WALL_INDEX_HT3D(:,0)` at the end of the routine.  I see all zeros.

[2D_PYRO3D.fds.txt](https://github.com/firemodels/fds/files/4751847/2D_PYRO3D.fds.txt)

rmcdermo:
A similar case has just been posted to the discussion forum:

a case similar to "pyro3d_wood_char" is used to demonstrate my problem. The size of a single OBST is equal to the CFD cell-size.
Specifications of the attached input files are:

[...]

PR #9196, opened 2021-02-08
add VSRVC for 2D-PYRO3D

MortezaHaghighi:
Add missing case selection for the 2D case of PYRO3D

Issue #8230


marcosvanella:
Looks like you have some tabs in this change. _db is throwing a warning.

MortezaHaghighi:
Should I revert the PR, or it's editable?

rmcdermo:
I'll deal with it.  Please set your editor to not use table.

rmcdermo:
Should be fixed #9197

Issue #9198, opened 2021-02-08
EVACUATION_ONLY array is not allocated when there is an input line error


marcosvanella:
Grab any case in the Verification suite, add &MISC AAA=0 / (nonsense input), and when running you will see after the error write out:

MPI startup(): Warning: I_MPI_PMI_LIBRARY will be ignored since the hydra process manager was found

 Starting FDS ...

 MPI Process      0 started on blaze.el.nist.gov

 Reading FDS input file ...

ERROR: Problem with MISC line (CHID: cone_demo_2)
forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable EVACUATION_ONLY when it is not allocated

Image              PC                Routine            Line        Source             
fds_impi_intel_li  000000000C346CDF  Unknown               Unknown  Unknown
fds_impi_intel_li  0000000003F0A515  MAIN__                    122  main.f90
fds_impi_intel_li  000000000040A022  Unknown               Unknown  Unknown
libc-2.17.so       00002B8363B0D505  __libc_start_main     Unknown  Unknown
fds_impi_intel_li  0000000000409F29  Unknown               Unknown  Unknown

PR #9200, opened 2021-02-08
FDS Source: Issue #9198. Stop FDS if there is an ERROR in read.f90

mcgratta:


mcgratta:
When returning from READ, we must immediately make a call to STOP_CHECK to gracefully exit if any ERRORs were found.

PR #9209, opened 2021-02-10
FDS Validation: add CL DEVC for predicted Waterloo_Methanol cases

rmcdermo:


mcgratta:
The centerline species volume fractions for the methanol pool fires is
contained in the NIST_Pool_Fires folder. I understand you are combining
NIST and Waterloo experiments, but we don't need to duplicate the plots in
the FDS Validation Guide.

>

PR #9211, opened 2021-02-11
Integration of the McKeeney-Greengard-Mayo method in UScaRC and first corresponding test cases

SusanKilian:
This PR includes changes to 

- scrc.f90 : Include new McKeeney-Greengard-Mayo method (MGM) which solves the pressure equation in two passes: 1. global inhomogeneous structured Poisson solution, 2. local homogeneous unstructured Laplace solutions
- read.f90 : Include corresponding input parameters for the MGM method
- dump.f90: Print out achieved MGM velocity error along the interfaces in the usual ScaRC line in chid.out
 
MGM is available as a submethod of UScaRC. Along mesh interfaces different variants for the coupling of the local Laplace solutions can be used (and are currently being further developed). This usually requires a series of MGM iterations for the local Laplace solutions until a specified tolerance for the MGM velocity error has been reached (current default setting is 1E-4 with a maximum of 20 MGM iterations)

Some additional MGM testcases shall illustrate the necessary MGM input parameters related to:
- different solvers for the local Laplace problems (currently CG, LU, and permuted LU; the latter considers only the non-zero entries of the local right-hand sides).
- setting the interface boundary values for the local Laplace problems (currently based on simple mean values, others will follow soon)
- requiring higher accuracy for the MGM velocity error along the interfaces (using 1E-8 with 50 MGM iterations instead of the default)


rmcdermo:
This merge overwrote my changes from yesterday.  Everyone stop and let's get this fixed.

mcgratta:
My fault. I thought that this could not happen? I wasn't warned of a
possible overwrite.

>

rmcdermo:
@SusanKilian You need to be more careful about updating before you make commits.

@mcgratta Luckily this was minor, but let's be very careful with merging from contributors outside the normal group.  They usually need careful scrutiny.

rmcdermo:
Yes, it can happen.  You make a change.  I change it back, and so forth.

mcgratta:
If you make a change, and then I change it back, that assumes that I have
merged your change and deliberately changed it back. In this case, Susan
must have forgotten to merge in remote changes, and I would have assumed
that GitHub would have warned her about that. Whenever I try to push
something without first checking for upstream changes, I am stopped.

How can I tell if a pull request comes that the changes are OK to merge?

>

mcgratta:
Now I see it. I should have noticed that Susan would be unlikely to change
PDPA stuff. I see those lines of code in her commit.

>

rmcdermo:
In this case, yes.  But there is no ironclad guarantee.  I'd just say that we need to be more careful with outside PRs.  Susan promises to start making more frequent, and thus smaller commits.  Then these things are more obvious.  This is exactly why the "commit early and often" mantra exists.

SusanKilian:
Hello everyone, sorry for the troubles, but unfortunately I do not know what went wrong here. I did a 'reset --hard firemodels/master' directly before my PR and pushed that state to my repo, then I did the current changes to the fresh read and dump. Finally, I copied the rest from my own ScaRC repo to the FDS repo.
The PR indicated me that this can be done without conflicts. I would really like to avoid this confusion, but it's not clear to me how. Is there a better way to make sure that I really start from the current state of the FDS repo than the 'reset --hard'? And sorry, what is this PDPA stuff?

rmcdermo:
You have to be very, very careful with `reset --hard`.  In your case, I would copy your source files to a temporary directory (outside the repo, probably on your Desktop or something) and then clone a fresh copy of the repo.  Again, this is why we need small commits, more often.  You can't expect to be off line for months and then simply integrate into a file like `read.f90`.  We all touch that almost daily.

SusanKilian:
I am aware that all these files are changed very often and I definitely don't want to overwrite anything. So I've assumed so far that the cleanest way is to reset my repo to the very latest state before, to avoid any overlap here. But obviously this way has some risks. Of course, the smaller commits in the future should help to avoid such cases.

mcgratta:
I rarely use reset -- only when I'm in trouble, and it usually gets me into
more trouble. I always do this ritual everyday before starting:

git remote update
git merge firemodels/master
git status -uno

If I see that my local branch is ahead of my forked repository on GitHub
(origin), then I do

git push origin master

Now I do another

git status -uno

On branch master
Your branch is up to date with 'origin/master'.
nothing to commit (use -u to show untracked files)

I repeat this ritual when I am ready to commit, just in case more changes
have been made. Then, when making a pull request, I check the lines that
are changed and make sure that these changes are mine alone.

>

mcgratta:
I refer to Randy's notes

https://github.com/firemodels/fds/wiki/Git-Notes-Getting-Started

when i forget the ritual.

SusanKilian:
Thank you very much for the clarification!  I will avoid using 'reset --hard' in the future. The steps you described are actually also part of my daily procedure, based on just this wiki page. The problem was apparently that I didn't want to rely on it completely, but went the supposedly 'cleaner' route. I'll know in the future!

PR #9213, opened 2021-02-11
FDS User Guide: Remove SOOT_DENSITY

mcgratta:


mcgratta:
Simo, FYI, the variable SOOT_DENSITY was never used. I removed it from the code and documentation.

PR #9220, opened 2021-02-12
FDS Validation: Minor formatting for Matlab

mcgratta:


mcgratta:
Yearly battle with Matlab. We've fixed the weird spacing for superscripts in the linux version of Matlab, but the subscripts are still a problem. There is a unicode way to do subscripts, just like we do for superscripts, but besides the power of 2 and 3, things are not working well. The power of 2 and 3 are ASCII, I think, which is why they seem to work. 

Anyway, I noticed that writing {C_3H_8} instead of C_3H_8 with the TeX font interpreter holds the formula together, but allows the weird space to appear elsewhere. It is better this way, but not a permanent fix.

PR #9224, opened 2021-02-14
FDS Source: Add SOOT_DENSITY back to &MISC so there is a way to chang…

drjfloyd:
…e it


drjfloyd:
SOOT_DENSITY was first removed from RADI. Then a follow-on commit made it a global variable and used it in the species routines; however, there was no way to change the default value. It couldn't go back on RADI since that is read after SPEC so I put it on MISC.

PR #9226, opened 2021-02-14
FDS Source: Appears to be a typo in scrc.f90

drjfloyd:
I couldn't compile the current source which was surprising since firebot passed. It looks like there was a typo in scrc where a variable used PARIDSO rather than PARDISO. @SusanKilian  was this change correct?


gforney:
firebot builds fds with the MKL library.  The line of code with the problem was in the non-MKL portion of the code.  I'll look at adding a non-MKL fds build and test to firebot

SusanKilian:
Jason, this was perfect, thanks! I’ll also have a look to the non-MKL version again.

PR #9232, opened 2021-02-15
FDS Build: Remove static links for linux

mcgratta:


mcgratta:
FYI, I am removing compiler options that we don't need.

PR #9234, opened 2021-02-16
FDS Verification: Make DT_DUMP consistent with race_test_1

drjfloyd:


mcgratta:
The DT_DEVC lines of these cases are meant to be different so that the plots clearly show the cases to be equivalent. The dataplot routine only compares the last value. Were these cases causing trouble?

drjfloyd:
I did not look closely enough at the case. I see now the TIME_AVERAGED is false. When I ran these on the cluster they failed firebot and looking at the input I thought it was because DT_DUMP was different. I can undo the change.

PR #9248, opened 2021-02-18
FDS Verification: begin changing ribbed_channel cases from MEAN_FORCI…

rmcdermo:
…NG to FORCE_VECTOR approach

@mcgratta Note that I still have the 160 resolution case to do.  This will take a few runs to narrow in on the right force vector.  So, not quite ready to remove MEAN_FORCING yet.


mcgratta:
Something to think about -- we can keep MEAN_FORCING for this kind of
stuff, but limit it to uniform profiles. Much of the baggage associated
with MEAN_FORCING is when we have to integrate over each layer of cells in
the vertical direction. This is really where the basic idea falls apart.
But if we leave it as a counterpart of FORCE_VECTOR, where one wants to
specify the flow velocity rather than the pressure drop, that might be OK,
not for wind stuff, but for these channel and tunnel flow cases.

>

rmcdermo:
I'd rather purge it now.  I notice a dramatic difference in the amplitude of the oscillations around the mean with MEAN_FORCING compared with a constant FORCE_VECTOR.  See below; the highest resolution is still using MF.  The reality is, we have never had a good handle on DT_MEAN_FORCING... it's a really arbitrary parameter that affects the the qualitative behavior of the flow.  This case is a good example of how tricky MF is... the bulk flow is supposed to be 1.0 m/s.  But when you set that, you don't get exactly 1.  Why?  Because of all the other terms in the momentum equation interacting.  The folks who really do this nudging stuff for real, the weather codes, use a much more complicated scheme.  Anyway, at the moment I'm keen on purging.

ribbed_channel_Ubulk

PR #9251, opened 2021-02-19
FDS Verification: convert ribbed_channel_geom cases to FORCE_VECTOR

rmcdermo:


rmcdermo:
@mcgratta The only thing left in `fds/Verification` is `./Flowfields/mean_forcing_hole.fds`.  So, go ahead and purge when you are ready.

Issue #9235, opened 2021-02-16
HT3D OBST crosses MESH Boundary


jet5678:
Hello all,
I have noticed in a model that when HT3D OBSTs cross a mesh boundary the heat conduction in the direction of the surface normal of the boundary is stopped. Compare attached picture. If a minimal example is needed, I can provide it.
The picture shows a HT3D OBST with hot gas flowing around it. The black lines are the mesh boundaries running through the object.
Is it possible to implement 3D heat transport across mesh boundaries?

Thanks a lot
With best regards

![Screenshot 2021-02-16 173106](https://user-images.githubusercontent.com/50654162/108095806-8dad4600-7080-11eb-944d-bff60444d117.png)

drjfloyd:
Are you running with the most recent release of FDS? If you aren't try your case with the current release. If it is failing with the current release, please submit an input file.

rmcdermo:
Should work.  There are verification cases for this.  For example,

https://github.com/firemodels/fds/blob/master/Verification/Heat_Transfer/ht3d_2mesh_x.fds

jet5678:
I have created a minimal example in which the problem can be seen quite clearly. I have attached the code. I use FDS 6.7.5 with MPI 5.4.7 on CentOS version 8 Core.
[ht3d_obst_crossing_mesh_boundary.txt](https://github.com/firemodels/fds/files/5990960/ht3d_obst_crossing_mesh_boundary.txt)

rmcdermo:
Ok, thanks, I'll take a look.

rmcdermo:
I'm reproducing the problem.  Thanks for the report.

rmcdermo:
I think in this case the problem is with assigning OPEN boundaries to a VENT using MB.  This is dangerous.  Though we do not make a big deal about it, we do mention in `10.4.1 Basics` to use `DB` (domain boundary) for multi-mesh cases.  Try that (works for me) and let me know if that works for you.
```
! OPEN boundaries:
&VENT DB='XMIN', SURF_ID='OPEN' /
&VENT DB='XMAX', SURF_ID='OPEN' /
&VENT DB='YMIN', SURF_ID='OPEN' /
&VENT DB='YMAX', SURF_ID='OPEN' /
&VENT DB='ZMAX', SURF_ID='OPEN' /
```

jet5678:
[...]

PR #9253, opened 2021-02-19
FDS Source: Issue #9235. Fix bug in HT3D interpolation

mcgratta:


mcgratta:
FYI, when one declares an OPEN VENT, the ghost cells behind the vent are made non-solid, like SOLID(IC)=.FALSE.

If the OPEN vent is at the interface between two meshes with an HT3D object spanning, we must close the ghost cells (SOLID=T) so that the logic in the SOLID_HEAT_TRANSFER_3D routine works.

rmcdermo:
Thanks, nice detective work.

PR #9272, opened 2021-02-25
FDS Source: Add self-optimizing strategies for the local preconditioners and the local MGM-Laplace-solvers in UScaRC

SusanKilian:
This PR is intended to allow self-optimizing strategies when choosing the local Poisson preconditioner or the local MGM Laplace solver. In both cases, FFT is automatically selected if the current mesh is perfectly structured, i.e. contains no obstructions, or PARDISO if it contains obstructions.

It introduces the new option SCARC_PRECON = 'OPTIMIZED' for the global Krylov solver and the option SCARC_MGM_LAPLACE_SOLVER = 'OPTIMIZED' for the local MGM Laplace solvers.


rmcdermo:
@SusanKilian OK, this should be the last PR with 6000 lines of code change that I have no simple way to review. Please commit early and often.  Thanks

SusanKilian:
Yes, that was the last one I needed to get into this mode

PR #9282, opened 2021-03-01
FDS Source: Set correct scope and boundary conditions for local PARDISO preconditioners in MGM-Laplace solver

SusanKilian:
This PR fixes a bug in the boundary settings for the local MGM-Laplace problems


Source/scrc.f90
```diff
@@ -7733,9 +7732,11 @@ SUBROUTINE SCARC_SETUP_MGM_PARDISO(NM, NL)
 CROUTINE = 'SCARC_SETUP_PARDISO'
 
 CALL SCARC_POINT_TO_GRID (NM, NL)                                    
+
+MKL => L%MKL
 AS  => SCARC_POINT_TO_CMATRIX (NSCARC_MATRIX_LAPLACE_SYM)
 
-! Allocate workspace for parameters and pointers needed in MKL-routine
+! Allocate workspace for parameters nnd pointers eeded in MKL-routine
```
rmcdermo: introduced typos

SusanKilian: Oh, I thought I corrected them, thanks! I'll change them ...

PR #9293, opened 2021-03-03
FDS Source: Enable optional single-precision solution of the local Laplace problems by PARDISO

SusanKilian:
This PR adds the possibility to solve the local Laplace problems in the MGM method only with single precision in order to compare the accuracy and timings with the default double precision version. It also adds a corresponding test case for dancing_eddies.


Source/scrc.f90
```diff
@@ -7652,7 +7652,7 @@ SUBROUTINE SCARC_SETUP_PARDISO_MESH(NM, NL)
 MKL => L%MKL
 AS  => SCARC_POINT_TO_CMATRIX (NSCARC_MATRIX_POISSON_SYM)
 
-! Allocate workspace for parameters nnd pointers eeded in MKL-routine
+! Allocate workspace for parameters and pointers needed in MKL-routine
```
rmcdermo, line 7655: @SusanKilian Notice this line shows up as fixed yet again.  These are the kinds of things that make me nervous about you maintaining two repositories.  You are inevitably going to make mistakes in copying and rewriting files.

SusanKilian: @rmcdermo:  I am really with you on this point, this must be handled very carefully. In this particular case, however, it's actually because I had accidentally corrected only one of the two occurrences of the same typo error in the last PR, and I didn't notice it until after the PR. So it was actually a new correction today and should not happen again this way.

PR #9307, opened 2021-03-06
FDS Source: Allow a circular vent to live on a smaller surface

mcgratta:


mcgratta:
FYI, this change was necessary to put a circular vent on a slightly larger circular solid surface.

PR #9326, opened 2021-03-12
FDS Validation: Move Wu_Bakar_Tunnel_A-ScaRC case to separate folder

SusanKilian:
Allocate new FDS_Input_Files_ScaRC directory for Tunnel_Ventilation case and move ScaRC-case to it


rmcdermo:
Susan,
This is not how we are going to do this.  Let me reorganize.

PR #9327, opened 2021-03-12
FDS Source: Remove redundant IMPLICIT NONE statements

mcgratta:


mcgratta:
FYI, IMPLICIT NONE need only be invoked once in a MODULE, just below the USE statements. Fortran 2018 has additional functionality related to IMPLICIT NONE, so I am just paving the way for that. It's related to explicit vs implicit subroutine bindings and MPI.

PR #9335, opened 2021-03-12
FDS source: output geometry transparency to the .smv file

gforney:


gforney:
should have added a comment.  needed a fix in both fds and smokeview.  smokeview fix hadn't been made yet when I did this PR. my fds repo gets into trouble (with lots of merge commits) if I don't do a PR soon after I make a commit.  so I did this one first.

the smokeview your alias is pointing to should have the fix to show transparent geometry now.  there are a lot of combinations so let me know if you still see a problem

PR #9344, opened 2021-03-16
FDS Source: Use MPI 2008 standard Fortran bindings

mcgratta:


mcgratta:
This commit updates FDS to use the 2008 standard MPI-Fortran bindings. You will see
```
USE MPI_F08
```
instead of 
```
USE MPI
```
wherever we make MPI calls. The reason for this change is that the various parameters sent as arguments to MPI routines are typically typed as INTEGERs which doesn't conform to the latest Fortran standard (2008+). Now, these parameters have specific MPI datatypes which makes type checking easier for the compiler.

Marcos, when you can, delete the `-warn noexternals` from the makefile and compile the debug version of FDS. You will see a handful of externals that we need to resolve properly. One routine you use is SLEEP, which is non-standard Fortran. I'd like to get rid of it or replace it with a standard Fortran subroutines using DATE_AND_TIME. I can do this, let me know if you still need SLEEP.

PR #9349, opened 2021-03-17
FDS Source: Add MPI_F08 use statement and fix wrong matrix pointers in AMG

SusanKilian:
This PR adds a 'USE MPI_F08' in the SCARC_VARIABLES module and resets some matrix pointers in the Algebraic Multigrid. Besides, it includes some minor layout corrections.


rmcdermo:
@SusanKilian Looks like this **removes** MPI_F08 (??). Plus, why are you taking out timings?

SusanKilian:
@rmcdermo

It's right, that also a USE MPI_F08 is removed in the SCARC_METHODS module, but because it was doubled (was already included at the top of the module).

What should be new is the additional USE_F08 in the SCARC_VARIABLES module (line 1096) because of the changed REQ definition in line 1259.

As for the timings: I would like to rework the AMG timings because they didn't work properly as I have seen for the Wu_Bakar-cases. If you don't mind, I'd like to move this point back a bit (until after the work on the inseparable Poisson), because there are other related changes to be done here. The time measurements will certainly come in again in revised form.

Susan

rmcdermo:
I don't see anything added in line 1096.

SusanKilian:
I was also wondering that I couldn't find it in the compare list. I expanded it by hand and it looks like this, see attachment. It is the USE MPI_F08 in line 1096. Should I push again?
Screenshot_20210317_152218

rmcdermo:
This is no diff because that line is already on the master branch.  This is what concerns me about your workflow.  I am closing this PR.  Get your repos sorted and then send a new PR if something needs to be added.

singlewithguitar:
Susan want to go out on date

PR #9358, opened 2021-03-18
FDS Source: Add explicit interface for CLUSTER_SPARSE_SOLVER

mcgratta:


mcgratta:
Marcos, if this is OK, merge it. This eliminates the warning about explicit interface.

marcosvanella:
Should be fine.

mcgratta:
All that remains of implicitly interfaced externals are these calls in pres.f90
```
#ifdef __INTEL_COMPILER
   CALL KMP_SET_WARNINGS_OFF()
#endif
```
Do we even need these? The documentation on them is scarce.

marcosvanella:
These routines are omp5 routines from intel, added there to avoid warnings of the type:
```
 MPI Enabled;    Number of MPI Processes:       3
 OpenMP Enabled; Number of OpenMP Threads:      2

 MPI version: 3.1
 MPI library version: Open MPI v4.1.0, package: Open MPI mnv@PN102630.local Distribution, ident: 4.1.0, repo rev: v4.1.0, Dec 18, 2020

 Job TITLE        : Complex Geometry: Test helium-air through sphere, 3 meshes.
 Job ID string    : sphere_helium_3meshes

**OMP: Info #275: omp_get_nested routine deprecated, please use omp_get_max_active_levels instead.**
OMP: Info #275: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.
OMP: Info #275: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.
 Time Step:      1, Simulation Time:      0.01 s
 Time Step:      2, Simulation Time:      0.02 s
 Time Step:      3, Simulation Time:      0.03 s
 Time Step:      4, Simulation Time:      0.04 s
 Time Step:      5, Simulation Time:      0.05 s
```
When doing LU decomposition with the cluster solver when the case is threaded. It is a bug they said they would fix in next releases. Which flag are you using to pickup this externals issue?
The solution (need to test it) is to add:
```
#ifdef __INTEL_COMPILER
USE OPENMP, ONLY : KMP_SET_WARNINGS_OFF, KMP_SET_WARNINGS_ON
#endif
```
At the top of the GET_H_MATRIX_LUDCMP routine, below the USE MPI_F08.

marcosvanella:
Some reference on the previous at the end of this thread:
https://community.intel.com/t5/Intel-oneAPI-Math-Kernel-Library/Warnings-when-compiling-a-Fortran-program-with-MKL-and-ipo/m-p/1205507#M29991

PR #9361, opened 2021-03-19
FDS Source: Add IMPICIT NONE (TYPE,EXTERNAL)

mcgratta:


mcgratta:
IMPLICIT NONE (TYPE,EXTERNAL) checks for variable type and routine implicit/explicitness. This is a Fortran 2018.

PR #9374, opened 2021-03-22
FDS Source: Output all N_TRACKED_SPECIES in hrr.csv file

mcgratta:


mcgratta:
FYI -- I'm trying to improve output for cases with multiple reactions. The old _hrr.csv file had MLR_FUEL and MLR_TOTAL. Now we have the MLR of all the tracked species. This actually simplifies the code a bit, and gets rid of a lot of I_FUEL logic. I ran the verification cases and manually changed some validation output files, but watch out for other odd behavior.

PR #9386, opened 2021-03-25
FDS Source: Fix bug in ORIENTATION counter

mcgratta:


mcgratta:
Jason -- this does not fix the problem, but it did contribute to an excessive number of ORIENTATIONs

PR #9401, opened 2021-03-31
FDS Source: use molecular Schmidt in condensation model

rmcdermo:


rmcdermo:
@drjfloyd Please have a look at this and let me know if you see any mistake.  I'll run the cases

PR #9405, opened 2021-04-01
FDS Source : Remove NEW_SCALAR_TRANSPORT, don't CC exchange when soli…

marcosvanella:
…d phase only.


marcosvanella:
By NEW_SCALAR_TRANSPORT I mean the provisional logical flag and old Godunov treatment of interpolation in cut-cell region.

Issue #9400, opened 2021-03-31
Velocity flow at leak vent


mcgratta:
[test2d.txt](https://github.com/firemodels/fds/files/6238601/test2d.txt)
In this 2d case, there is a volumetric heat source in the small compartment, which has a leak to the larger compartment. When I look at the normal component of velocity on each VENT that constitute the leak, it is different. I would expect a small difference due to different densities, but this seems larger. Is this to be expected? You can see the difference visually in Smokeview by looking at the vectors of the cell-centered temperature. The devc file has the velocity components at every time step.

drjfloyd:
I'll take look. The face velocity is set using TMP_F of that face. Any delta in TMP_F is addressed with LEAK_ENTHALPY as a heat source/sink in the downstream gas cell.

mcgratta:
Yes, I assume rho*u should be the same, assuming the vent area is the same. But it appears to be much different. I'm wondering if perhaps I've hit a clip bound.

drjfloyd:
I modified your case to just have two chambers, constant cp for the gas, no radiative fraction, and a single vent.  The increase in mass in the one chamber does equal the loss in mass in the other chamber as well as the total mass flow in the leak duct. The energy though is off. Not sure why right now.

[test2da.txt](https://github.com/firemodels/fds/files/6238988/test2da.txt)

mcgratta:
The energy deficit might be related to temperature or density clipping.

drjfloyd:
In my simplified case I cut back the HRRPUV to avoid any clipping

drjfloyd:
Should be fixed.

mcgratta:
Thanks. I'll check how my original case behaves now.

mcgratta:
I modified my case to check the addition of internal energy for a fixed heat source of 40 kW for 1 s. I got 39.8 kJ. Very good, because I didn't go and tighten anything down and I left the 75% radiative fraction in place. These energy conservation cases are often complicated by radiation, but this one seems very good. Thanks.

PR #9406, opened 2021-04-01
FDS Source: Fix LEAK_ENTHALPY energy conservation Issue #9400.

drjfloyd:


drjfloyd:
Should be fixed now.  I made a small test case for this that I'll add to Verification Guide. I guess I never made a verification case for LEAK_ENTHALPY. The case is a 2D box divided in 2. Walls are adiabatic. One side has an HRRPUV and there is a LEAK between them.  The change in mass near zero (1E-6 kg out of 2 kg total) and the HRRPUV adds 360 kJ which equals the total change in internal energy. Before this fix the mass was OK but the internal energy was off a little.

[leak_enthalpy.txt](https://github.com/firemodels/fds/files/6245814/leak_enthalpy.txt)

PR #9409, opened 2021-04-02
FDS Source: Replace hard coded 5000 K with I_MAX_TEMP

drjfloyd:


mcgratta:
```
[mcgratta@burn Verification]$ grep forrtl */*err
Fires/fire_const_gamma.err:forrtl: error (72): floating overflow
Flowfields/helium_1d_const_gamma.err:forrtl: error (72): floating overflow
Heat_Transfer/adiabatic_con_flux.err:forrtl: error (72): floating overflow
Heat_Transfer/adiabatic_net_flux.err:forrtl: error (72): floating overflow
Pressure_Effects/zone_shape.err:forrtl: error (72): floating overflow
Pressure_Effects/zone_shape.err:forrtl: error (72): floating overflow
```

drjfloyd:
looks like something to do with CONSTANT_SPECIFIC_HEAT_RATIO. If I get rid of that I don't see an error.

mcgratta:
Something's going wrong in the JANAF table for H_V_H2O calculation. Look at zone_shape case.

mcgratta:
Looks like you're using CP_G_O before initializing
```
SUBROUTINE CALC_H2O_HV

USE PROPERTY_DATA, ONLY: JANAF_TABLE,JANAF_TABLE_LIQUID
CHARACTER(LABEL_LENGTH) :: WATER_VAPOR='WATER VAPOR'
INTEGER :: I
REAL(EB) :: CP_G,CP_G_O,CP_L,CP_L_O,H_G,H_L,H_G_0,H_L_0,G_F,RCON,H_V,T_R,T_M,T_B,DENSITY,MU_LIQUID,K_LIQUID,BETA_LIQUID
REAL(EB) :: H_L_3,H_L_4,H_G_3,H_G_4
LOGICAL :: FUEL

DO I=0,I_MAX_TEMP
   CALL JANAF_TABLE (I,CP_G,H_G_0,WATER_VAPOR,RCON,FUEL,G_F)
   CALL JANAF_TABLE_LIQUID (I,CP_L,H_V,H_L_0,T_R,T_M,T_B,WATER_VAPOR,FUEL,DENSITY,MU_LIQUID,K_LIQUID,BETA_LIQUID)
   IF (I==0) THEN
      H_G = H_G_0
      H_L = H_L_0
   ELSE
      H_G = H_G + 0.5_EB*(CP_G+CP_G_O)
      H_L = H_L + 0.5_EB*(CP_L+CP_L_O)
   ENDIF
write(0,*) I,H_G,H_L
   H_V_H2O(I) = H_G-H_L
   CP_G_O=CP_G
```

mcgratta:
RCON also seems uninitialized

drjfloyd:
Yes. That is the root issue.  I'll get that fixed and pushed up.

mcgratta:
I can't figure out how this worked before. Did we just not use H_V_H2O with CONSTANT_CP?

drjfloyd:
Had to make H_V_H2O allocatable in cons.f rather than just (0:5000). Then just got tripped up by the mess of logic that exists between REAC and SPEC in handling the definition of AIR and simple chemistry. Wound up with a path for CONSTANT_CP where things broke down. Part of the fix was to move setting up the arrays of atomic weight and definition of basic species MWs earlier in read. We were actually doing that fairly late, but it was always being done so could just move it earlier.  In the loop above For I=0, C_G_0 and C_L_0 are not used.  After the I=0 if block they are set and available for I>=1.

mcgratta:
Yes, I tried to hack in RCON=R0/MW_H2O but discovered that MW_H2O wasn't
defined yet.

>

mcgratta:
Should these be changed
```
[mcgratta@burn Source]$ grep "MIN(4999" *f90
fire.f90:      D_F = MAX(D_F,D_Z(MIN(4999,NINT(TMP_IN)),RN%FUEL_SMIX_INDEX))
func.f90:ITMP = MIN(4999,INT(TMPG))
func.f90:ITMP = MIN(4999,INT(TMPG))
func.f90:   ITMP = MIN(4999,INT(TMP_S))
func.f90:   ITMP = MIN(4999,INT(TMP_S))
wall.f90:                        ITMP = MIN(4999,INT(T_NODE))
wall.f90:      ITMP = MIN(4999,NINT(ONE_D%TMP_F))
wall.f90:      ITMP = MIN(4999,NINT(ONE_D%TMP_F))
wall.f90:               ITMP = MIN(4999,INT(ONE_D%TMP(I)))
wall.f90:                  ITMP = MIN(4999,INT(T_NODE))
wall.f90:ITMP = MIN(4999,NINT(TMP_FILM))
```

drjfloyd:
Good catch. Yes, I forgot to search for 4999. Those should be I_MAX_TEMP-1.  I guess should just make a variable I_MAX_TEMP_M1 or something so we aren't wasting cycles doing the subtraction all the time.

mcgratta:
I think I_MAX_TEMP-1 is OK.

>

drjfloyd:
OK.

mcgratta:
I think a more important question is whether we need the linear
interpolation in all the GET_ routines. I recall adding that to try to
match these various analytic solutions, but in the end did it matter?

>

drjfloyd:
It doesn't matter so much for things like K, CP, and MU. It does matter for enthalpy type quantities which should be only quantities where it is done.

mcgratta:
OK, let's leave that for another day.

>

drjfloyd:
I originally had it for all the GET. You removed it all. Then later there were some issues here and there with HRR and energy that were solved by adding it back to just the enthalpy quantities. It is not a terribly expensive thing.

mcgratta:
I remember going back and forth on it. OK, that's fine.

>

PR #9456, opened 2021-04-26
FDS User Guide: Add DX, DY, DZ for SPATIAL_STATs and POINTS>1

mcgratta:


mcgratta:
Jason, Randy -- have a look at this write up and let me know if it makes sense to you. This is the case where you want to use SPATIAL_STATs for each point in a linear array.

Issue #9460, opened 2021-04-28
Add TEMPORAL_STATISTIC='FAVRE AVERAGE'


rmcdermo:
Looks like we have SPATIAL_STATISTIC='MASS MEAN', which is equivalent to a "Favre filtered" quantity.  We are getting asked to understand the differences between the Reynolds averages and Favre averages in the centerline species measurements for the pool fires.

rmcdermo:
This should be working now.  But I want to add a verification case using the Shunn manufactured solution.

rmcdermo:
Here is the case with 'TIME AVERAGE' and 'FAVRE AVERAGE'.  The time average is reasonably behaved.  Still trying to figure out what is going on with FAVRE.
[shunn3_32.fds.txt](https://github.com/firemodels/fds/files/6401934/shunn3_32.fds.txt)

rmcdermo:
OK, this should be fixed now with PR #9467 .  Here is a test case:
[shunn3_32.fds.txt](https://github.com/firemodels/fds/files/6402087/shunn3_32.fds.txt)

Note that this test case uses the correct `'RUNNING AVERAGE'` for `'YBAR'`.  It is interesting that the FAVRE average is about 5 % lower.  Might be part of the issue with the Sandia Helium Plume case.

rmcdermo:
It is clear from looking at some preliminary results with the 1 m helium plume that we need to get `TEMPORAL_STATISTIC='FAVRE AVERAGE'` working in conjunction with `SPATIAL_STATISTIC='INTERPOLATION'` _and_ `POINTS>1`.  This will require some massaging because right now `FAVRE` makes use of the `VALUE_1` and `VALUE_2` from `SPATIAL_STATISTIC='MASS MEAN'`.

rmcdermo:
This is working and has a verification case.  See `fds/Verification/Species/favre_test.fds`

PR #9465, opened 2021-04-29
FDS Source: add TEMPORAL_STATISTIC='FAVRE AVERAGE'

rmcdermo:
This addresses Issue #9460


mcgratta:
It appears that your volume for 'MASS MEAN' is the nearest 8 cells. This means that you will be averaging over 8 cells rather than 1 for 'FAVRE AVERAGE'.

PR #9466, opened 2021-04-29
FDS Source: update FAVRE AVERAGE; still something wrong though

rmcdermo:


rmcdermo:
Is this what you mean?  Something is still wrong, though, because with this case, my YTILDE is greater than 1.
[shunn3_32.fds.txt](https://github.com/firemodels/fds/files/6401930/shunn3_32.fds.txt)

Issue #9468, opened 2021-04-29
Build fails with gfortran 11.1.0


marcfehling:
I've been trying to build FDS with the latest GNU Fortran compiler 11.1.0 and the latest release of OpenMPI 4.4.1.

However, if building FDS with this configuration either on the latest release 6.7.5 or the master branch, compilation fails with different errors as follows:
- `6.7.5`
```
Building mpi_gnu_linux_64
[...]
mpifort -c -m64 -O2 -std=f2008 -ffpe-summary=none -fall-intrinsics -cpp -DGITHASH_PP=\"-\" -DGITDATE_PP=\""\"" -DBUILDDATE_PP=\""Apr 29, 2021  17:08:57\"" -DCOMPVER_PP=\""Gnu gfortran 11.1.0"\"   -fopenmp ../../Source/main.f90
../../Source/main.f90:2726:61:

  359 |       CALL MESH_EXCHANGE(2) ! Exchange radiation intensity at interpolated boundaries
      |                                                                                     2
......
 2726 |             IF (ICYC>1) ANG_INC_COUNTER = M%ANGLE_INC_COUNTER
      |                                                             1
Error: Index variable ‘ang_inc_counter’ redefined at (1) in procedure ‘mesh_exchange’ called from within DO loop at (2)
../../Source/main.f90:3103:59:

  359 |       CALL MESH_EXCHANGE(2) ! Exchange radiation intensity at interpolated boundaries
      |                                                                                     2
......
 3103 |          IF (ICYC>1) ANG_INC_COUNTER = M4%ANGLE_INC_COUNTER
      |                                                           1
Error: Index variable ‘ang_inc_counter’ redefined at (1) in procedure ‘mesh_exchange’ called from within DO loop at (2)
../../Source/main.f90:2726:61:

  806 |             CALL MESH_EXCHANGE(2)
      |                                 2                            
......
 2726 |             IF (ICYC>1) ANG_INC_COUNTER = M%ANGLE_INC_COUNTER
      |                                                             1
Error: Index variable ‘ang_inc_counter’ redefined at (1) in procedure ‘mesh_exchange’ called from within DO loop at (2)
../../Source/main.f90:3103:59:

  806 |             CALL MESH_EXCHANGE(2)
      |                                 2                          
......
 3103 |          IF (ICYC>1) ANG_INC_COUNTER = M4%ANGLE_INC_COUNTER
      |                                                           1
[...]

PR #9500, opened 2021-05-11
FDS Source: Make scrc compile with new gfortran 11.1.0 version

SusanKilian:
Related to issue #9468 I changed the following things: 
1. deleting redundant MPI_F08-use statements where it was already included via GLOBAL_CONSTANTS before 
2. stricter application of some other USE statements 
3. removal of a supposed variable redefinition
4. moving a subroutine to a different location in the same module


SusanKilian:
I would like to add another lines which are based on the use of -Werror and Wextra

PR #9509, opened 2021-05-14
FDS Source: revert change to BAROCLINIC_CORRECTION that broke 2nd-ord…

rmcdermo:
…er temporal convergence

### DIFF
--- a/Source/velo.f90
+++ b/Source/velo.f90
@@ -3074,13 +3074,23 @@ IF (PREDICTOR) THEN
    VV => V
    WW => W
    RHOP=>RHO
-   HP => H
+   IF (PRESSURE_ITERATIONS>1) THEN
+      HP => H
+   ELSE
+      HP => HS
+   ENDIF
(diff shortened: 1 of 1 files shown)


mcgratta:
This will break the tunnel cases. In such cases H and HS are very
different, and using the wrong one on the first pressure iteration causes
trouble.

>

rmcdermo:
Which cases

rmcdermo:
Point me to the case that breaks.

mcgratta:
Cases with TUNNEL_PRECONDITIONER. Often the H and HS profiles will be
counter-acting, like an upward V vs a downward V. By changing H to HS or
vis versa on the first step, the rate of convergence gets thrown off.

>

rmcdermo:
I dialed back the repo to last night's source (passed firebot) and the `Pressure_Solver/tunnel_demo.fds` case fails if I use a single MPI process.  I can start an issue if you want.

With latest source (my change included) the `tunnel_demo` and `dancing_eddies` cases seem to be running fine.  I'll check the validation now.

PR #9521, opened 2021-05-18
FDS Source: Add column name COLOR_INDEX to EVAC

gjaeger:
Add column name `COLOR_INDEX` to _EVAC: Initial positions of the agents_ 

@rmcdermo I have only changed line 7082. I cannot explain the other changes. What can this be due to?


rmcdermo:
@gjaeger It's fine.  You have your editor set to remove white space on save.  I do the same, so I often see other lines changed,  but this is something we want (to get rid of extra white space).

PR #9540, opened 2021-05-26
Matlab: fix fonts in pyrolysis.m

rmcdermo:


mcgratta:
Isn't this bringing back the improper spacing in superscripts?

>

rmcdermo:
I don't know.  I'm too busy with carpentry stuff.  Test it on burn.  At least from here you can add your superscript magic unicode.  The main problem was just the fontname and interpreter on the right side.

PR #9567, opened 2021-06-15
FDS User Guide: change SLCF --> BNDF for TERRAIN_CASE section

rmcdermo:
@mcgratta Please double check me on this, but I think this should be BNDF. 

However, as discussed in Issue #9566 , I don't think we need `TERRAIN_CASE`.  Even here in the example of the user guide.  If `TERRAIN_IMAGE=...` then we have a flag to include the terrain image.  If we need a special BNDF for that conforms to the terrain, this should be a flag on the BNDF line, not a MISC parameter.


rmcdermo:
Kevin,
If I don't merge this, my fork is going to be out of sync.  If it is technically correct at this instant, then let's merge it.  And then we can purge the section if we get rid of terrain files.

mcgratta:
Sorry -- I didn't notice the need to review. Let me do that now.

PR #9570, opened 2021-06-16
FDS Validation: Added Askervein cases to dataplot_inputs

noellecrump2023:


rmcdermo:
@noellecrump2023 This is not quite ready yet.  If we were to run firebot tonight with this file, it would crash.  We need two more things:
1.  Output data committed to the `firemodels/out` repo.
2. Preprocessing script added to `firemodels/fds/Utilities/Matlab/scripts/` and also `.../Matlab/FDS_validation_script.m`

I'm closing for now.  But work on putting the appropriate files in the `out` repo.  Your processing script needs to grab the files from there, not your test directory.

noellecrump2023:
got it, good to know.

PR #9572, opened 2021-06-18
FDS Source: Overhaul liquid fuel model; allow multiple components

mcgratta:


mcgratta:
There is still work to be done for liquid fuels, but at least now the Pool Fires cases display surface temperatures just below the boiling point. To do -- look at the ramp up time of these pool fires. Still slow. The steady state portion seems to be working well, but it's slow to get to that point.

PR #9690, opened 2021-07-20
FDS: Updated Devices for new input file

noellecrump2023:
FDS recently had problems with an input file, and sections of it were removed while debugging. These sections are now restored.


rmcdermo:
So, spatial interpolation not working?

PR #9719, opened 2021-08-02
Zero thickeness changes

idpark0129:
Added default geometry thickness for when user defines a zero thickness wall.


marcosvanella:
Isaac, delete please the write statement:
WRITE(LU_ERR,*) 'XMIN,MAX in Z=',XB(5),XB(6)
in geom.f90 and push that change also. Thanks.

PR #9720, opened 2021-08-02
Complete Askervein 16m resolution data plot compatibility 

noellecrump2023:
Changed names of Askervein matlab script output filenames to include resolution.
Added Askervein Preprocessor to Matlab dataplot script
Added Askervein 16m resolution graphs to dataplot CSV file (compatible with the names changed above)
Changed Askervein 16m resolution graphs to be commented out. (because that's what you said you wanted before)


rmcdermo:
@noellecrump2023 I'm going to test this change before merging.

rmcdermo:
@noellecrump2023 When I try to run your script from a clean repository, I get this error.  This usually means you have files on your machine that have not been committed.  In this case, it looks like the files for the `out` repo have not been committed.  This is fine because we do not have the results finalized yet.  But it means that your scripts would break a firebot run.  Therefore, do not submit the change the `FDS_validation_script.m` in your pull request.  Everything else should be fine.  You will need to revert that commit in your local repo.  Thanks

Screen Shot 2021-08-02 at 2 03 31 PM

noellecrump2023:
Got it

noellecrump2023:
Will do

PR #9757, opened 2021-08-10
FDS Source: fix AREA INTEGRAL for GEOM gas phase quantities

rmcdermo:


rmcdermo:
@mcgratta
Thanks!  The XB for GEOM seems to work well now.

@marcosvanella 
Note I added AREA to the geom_channel case.  You'll see that the gas phase area comes out correctly whether XB spans the domain or just the gas phase region.  Let me know if you see any issues.

PR #9798, opened 2021-08-13
FDS Source : Initialize external mesh CFACEs.

marcosvanella:
- Also delete some unused vars/routine.


Source/fire.f90
```diff
@@ -717,122 +717,6 @@ SUBROUTINE REACTION_RATE(DZZ,ZZ_OLD,DT_SUB,RHO_0,TMP_0,KINETICS,Q_REAC_OUT,PRIOR
 END SUBROUTINE REACTION_RATE
 
 
-SUBROUTINE GET_FLAME_TEMPERATURE(TMP_FLAME,PHI_TILDE,ZZ_HAT,ZZ_0,ZZ_IN,TMP_IN)
-
-! Compute adiabatic flame temperature for reaction mixture
-! This routine was initially added for a TRI (turbulence radiation interaction) model, which has since been removed (unused)
```
rmcdermo, line 723: @marcosvanella I wanted to leave this

rmcdermo: Well, I suppose it's probably throwing an unused error with GNU?  Let's comment it.  This algorithm too me some time to get right and I suspect we will need this at some point.

marcosvanella: The whole routine should be commented. It's throwing an unused subroutine warning with gfortran.
I'll revert.

PR #9828, opened 2021-08-23
FDS Source: Fix bug in INTERPOLATION statistic

mcgratta:


mcgratta:
This change was needed because of commit 9fd87ed which changed the logic of how device volumes are chosen.

PR #9835, opened 2021-08-24
FDS Source: Do not let CF_DIVIDE divide by zero

mcgratta:


mcgratta:
Jason, could you double check this. I'm using the DIVIDE control function, and my initial denominator is zero. That changes the control value to infinity and latches the device such that it is never called again. I no longer allow a divide by zero, but I don't know what the alternative should be. Currently, it returns the numerator, but that is not right either.

PR #9877, opened 2021-09-02
FDS Source: Allow MAX or MIN control function to work with a CONSTANT

mcgratta:


mcgratta:
Jason -- FYI. Can you double check the ERROR messages for CONTROLs, They seem to be outdated.

drjfloyd:
Yep, these have gotten stale. I'll reorganize the list in PREC to group the functions based on restrictions (i.e. something like type <100 are logical type 100 to 200 take one input, etc) and update read.

drjfloyd:
Updated the error checking.

mcgratta:
Great, thanks.

PR #9897, opened 2021-09-09
Update PMMA_60_kW.fds

TristanHehnen:
Changed sample OBST thickness from 4 cm to 2 cm such that 1D heat conduction takes the back face temperature within the simulation domain into account.


mcgratta:
Thanks -- could you check the other two files too.

Issue #9856, opened 2021-08-26
prohibitively large memory use by particles for veg


ruddymell:
In simulating the Camp Swift research burn in grass we need to test different representations of the grass species using the particle model in FDS. This can result in up to 16 different types of particles (e.g., blade of grass versus the stem; lower moist versus upper dry layer). 

The attached plot shows the memory used (kB per grid cell) for FDS runs (on a MacPro) with increasing number of particle types. These numbers are from the top command in OSX or the private memory use for the process using the activity monitor. The input file in also attached. 

For a 100 m x 100 m plot with 1 m tall grass veg, a 10 cm grid resolution, and 16 types of particles (this is our most demanding case) the required memory would be 2.4 TB (which is a show stopper). With a 25 cm grid the memory need is 152 GB (which is workable). 

Assuming these results and tests are correct, is there something that can be done reduce the memory usage for this application?
![Plot_PartMem](https://user-images.githubusercontent.com/12721870/131038495-5c9d9c36-28a9-439d-a865-9f8ff31a7706.png)
[input_PartMemTest.txt](https://github.com/firemodels/fds/files/7062801/input_PartMemTest.txt)

mcgratta:
Probably. This might be a situation where we massively over-allocate something and never noticed until now. I'll take a look.

mcgratta:
Is the issue here the total number of particles? Or the total number of particle types?

mcgratta:
So far, the only thing I can suggest is to set `N_LAYER_CELLS_MAX(1)=1` on each `SURF` line. This limits the number of solid cells to 1, making these thermally thin. Currently, FDS uses 2 cells, but because of the potential for swelling, 5 cells are allocated. This reduced the memory by about 20%.

drjfloyd:
With 10x10x10 meshes, you have 1000 grid cells per mesh and 600 wall cells per mesh. Is it potentially an issue with having to allocate arrays for mesh-to-mesh particle exchanges; does memory usage improve if just one mesh is used?

ruddymell:
The issue is more with the total number of particle types.
The tests to date, and in the provided plot and in input file, were with one mesh

mcgratta:
I ran with the same number of particles, but with 1 particle type instead of 18. This saved some, but not alot. Why do you think it has to do with particles types?

ruddymell:
[...]

PR #9905, opened 2021-09-13
FDS Source: Issue #9856. Reorganize PARTICLE, WALL, and CFACE variables

mcgratta:


mcgratta:
Note -- this commit has many changes, but they should not affect code logic. It is just a re-org of boundary variables.

PR #9910, opened 2021-09-15
FDS Source: Introduce ScaRC no-flux call for inseparable Poisson problem in main

SusanKilian:
This PR makes the new no-flux routine for the inseparable Poisson solver available within PRESSURE_ITERATION_SCHEME


mcgratta:
Something seems to have gone very wrong with this last commit.

PR #9912, opened 2021-09-15
FDS Build: Fix dependency order

mcgratta:


mcgratta:
Susan -- check this carefully. I do not believe that velo.o depends on scrc.o, but double check.

PR #9942, opened 2021-09-29
FDS Source: optional disk drag model for cartesian particles

ericvmueller:


rmcdermo:
Eric,
As you work these new features, please make sure you keep the user guide updated.  It's much easier to do this as you go, rather than expecting you will remember to do it later.  Thanks!

ericvmueller:
Yup, will do!

PR #9968, opened 2021-10-08
Allocate meshes to MPI process automatically

JakeOShannessy:
This is more of an issue, but comes as a pull request as I've included a proof-of-concept. If this something worth doing I'll develop it further.

The default method for allocating meshes to MPI process is to sequentially assign one mesh per process until we reach the last process where we assign all remaining meshes. Where there are fewer process than meshes this is obviously not the most efficient approach.

Meshes can be manually assigned using MPI_PROCESS, but we've found that many modellers don't end up doing this, as it ties the model to a particular number of processes. It's far more efficient to do this so it would be nice to have at least a slightly better default mesh allocation to get some of this efficiency automatically.

I've included an performance example below. It's not perfectly real-world, but informative. "Auto/Patched" is with a very naive algorithm in the code of the pull request.

benchmark

Is this something worth pursing, or is there are particular reason/limitation that requires mesh allocation to be the way it is currently?

Even a fraction of the above performance boost would be hugely advantageous to us.

### DIFF
--- a/Source/read.f90
+++ b/Source/read.f90
@@ -537,12 +537,10 @@ USE EVAC, ONLY: N_DOORS, N_EXITS, N_CO_EXITS, EVAC_EMESH_EXITS_TYPE, EMESH_EXITS
                 EMESH_NM, N_DOOR_MESHES, EMESH_NFIELDS, HUMAN_SMOKE_HEIGHT, EVAC_DELTA_SEE, &
                 EMESH_STAIRS, EVAC_EMESH_STAIRS_TYPE, N_STRS, INPUT_EVAC_GRIDS, NO_EVAC_MESHES
 INTEGER, INTENT(IN) :: IMODE
-REAL :: NMESHES_R, N_MPI_PROCESSES_R
-INTEGER :: MESH_COUNTER 
-INTEGER :: MESHES_PER_PROCESS, MIN_PROCESS
 INTEGER :: IJK(3),NM,NM2,CURRENT_MPI_PROCESS,MPI_PROCESS,RGB(3),LEVEL,N_MESH_NEW,N,II,JJ,KK,NMESHES_READ,NNN,NEVAC_MESHES, &
            NMESHES_EVAC, NMESHES_FIRE, NM_EVAC, N_THREADS, JBAR_OLD_VALUE
 INTEGER, ALLOCATABLE, DIMENSION(:) :: NEIGHBOR_LIST
+INTEGER, ALLOCATABLE, DIMENSION(:) :: OMP_THREADS !< Stores the value of the N_THREADS parameter for each mesh for future use.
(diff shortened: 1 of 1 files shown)


rmcdermo:
Looks like a no brainer to me.  But I'll let Kevin review it.  Thanks, Jake!

JakeOShannessy:
Thanks Randy, in that case let me iterate on it a little. The proof-of-concept code likely doesn't cover OpenMP threads and evac meshes correctly.

mcgratta:
! TODO: some processes won't be touched this way, need to determine if
this
 ! is an issue.

Yes, this is an issue. Also, this will probably break EVACuation stuff,
which has its own additional logic for mesh assignment. IMO, this seems to
be a good task for the pre-processor like PyroSim. What you really want to
do is assign processes based on mesh sizes. Currently in READ_MESH, we sort
out the mesh/process mapping before even reading in the mesh size. I'd
rather do all this in the pre-processor.

>

mcgratta:
I'm going to add Timo to the conversation. He has to ensure that this will not break FDS+EVAC.

JakeOShannessy:
Thanks for taking a look Kevin. I had originally looked at doing a pre-processor as well. In the end I thought I'd try this route first in order to keep everything as mainstream as possible. Having CHID.fds be completely agnostic to the number of processes is a really useful feature for us, so that's what I'm trying to aim for.

I think the issue I highlighted in my TODO item is resolved. The way I've approached evac is to try and leave the state of mesh allocation as if it were specified using the `MPI_PROCESS` parameter before the evac code kicks in. Ideally it shouldn't be able to tell the difference, but that may not be true in practice.

Ideally I'd like to be able to use cell counts in the algorithm, but that might mean re-ordering meshes, which touches a lot more code. A pre-processor would have an advantage here, but it means CHID.fds files are no longer agnostic to the number of processes.

JakeOShannessy:
Actually a quick note on using mesh cell counts. It can still be done with the setup I've pushed above (mesh/process allocation is slightly deferred), but it doesn't allow for mesh reordering, so it would still be a "best effort" with the given mesh ordering.

mcgratta:
I've decided not to pull this request. I appreciate the work you've done, but I fear that this functionality might have unintended consequences. I just added a quick check in read.f90 that stops the code in error if someone creates more meshes than MPI processes without specifically using MPI_PROCESS to assign them. I think that in most cases the user should be told to fix what might just be a silly mistake rather than have FDS fix it. For example, someone might create 17 MESH lines and then ask for 16 MPI processes. No amount of re-finagling by FDS is going to save this job from being half as efficient. As a user, I would rather be told to fix it myself. I would also want this kind of logic to be built into a program like PyroSim or Blender, which would have much more latitude in reordering meshes than FDS does.

JakeOShannessy:
This question has come up a few times, so for posterity I'll record my solution here. We went with a separate pre-processor https://github.com/Smoke-Cloud/fdspp.

gforney:
jake, could you check the link.  Im getting a 404 error

JakeOShannessy:
Apologies, should be fixed now.

PR #9969, opened 2021-10-09
FDS Source: fix divide by 0 for PRYO3D with mass transport

rmcdermo:


mcgratta:
Setting ZZ=0 can also be problematic if ZZ is to be sent, for example, into a `GET_SOMETHING` subroutine or function. Would it be better to just reset RHO and ZZ to ambient values?

rmcdermo:
This code is dead.  I'd rather just purge it than change it.  The only reason it is still here is to possibly help Simo and Christian.  If they don't need it, we can get rid of it.

PR #9976, opened 2021-10-13
Increased domain to cover the complete reaction zone

TristanHehnen:
As discussed via email, here is the pull request with the increased domain in an effort to cover the complete reaction zone. All three FDS input files and the `Run_All.sh` were adjusted.

Commit message: "Increased domain to cover the complete reaction zone (based on PMMA case). Note: number of meshes was increased to 420, from before 96."

### DIFF
--- a/Validation/NIST_NRC_Parallel_Panels/FDS_Input_Files/PBT_60_kW.fds
+++ b/Validation/NIST_NRC_Parallel_Panels/FDS_Input_Files/PBT_60_kW.fds
@@ -1,13 +1,16 @@
 &HEAD CHID='PBT_60_kW', TITLE='Parallel Panel Test, PBT, Propane, 60 kW' /
 
-&MESH IJK=15,20,20, XB=-0.60,-0.30,-0.40,0.00,-0.30,0.10, MULT_ID='mesh' /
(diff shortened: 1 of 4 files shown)


rmcdermo:
@TristanHehnen 420 gives me pause.  Are we getting mass loss rate (MLR) correct with the smaller domain?  If so, I suggest that we make MLR*HOC the value that we compare with experiment.

TristanHehnen:
I haven't checked MLR yet. If the number of meshes is too large I could readjuste/rearrange them back to something closer to 96 again.

rmcdermo:
Let's check MLR for original configuration.  If it is correct, then I don't see a need to change the domain.

mcgratta:
Yes, the number of meshes is too large. We routinely run these validation cases to check new releases, and we try to keep the cases to a manageable size.

TristanHehnen:
Okay, the MLR plots are all relatively close to each other.

rmcdermo:
Then I suggest we just make this an Issue and change the output to a DEVC with MLR * HOC.

mcgratta:
If that is the case, let's use an output line like that is similar to this
```
&DEVC ID='HRR', QUANTITY='MASS FLUX', SPEC_ID='PMMA FUEL', XB=..., SURF_ID='PMMA panel', SPATIAL_STATISTIC='SURFACE INTEGRAL', CONVERSION_FACTOR=43000, UNITS='kW' /
```

PR #10002, opened 2021-11-03
FDS Source: Add option to neglect surface oxidation in total HRR output

ericvmueller:
Kevin, I will leave this unmerged so you can take a look and see if you're fine with that, or you want to go about it another way.


mcgratta:
Remember to add this new parameter to the FDS User Guide.

PR #10016, opened 2021-11-08
FDS Source: improve particle ring insertion

### LINKED ISSUE #10015: Particle ring insertion
There are currently a couple of issues with particle ring insertion when the ring spans multiple meshes:

1) A ring with random spacing will have `N_PARTICLES` in _each_ mesh, rather than in total. You can see this in the output file for the attached example, depending if you use 1 or 2 meshes. This is not ideal, since you can end up with meshing-dependent results for things like the Douglas fir case.

2) A ring with uniform spacing will get stuck in initialization if it tries to insert a particle in a different mesh. This leads to an infinite do loop in the `UNIFORM_RING` function of func.f90. This can also be seen in the attached example if 2 meshes are used and `UNIFORM=T`

I will send a pull request with some potential fixes. These can probably be improved for efficiency and to make sure the total number of particles is strictly enforced, but it is a start.

[ring_test.txt](https://github.com/firemodels/fds/files/7498247/ring_test.txt)

ericvmueller:
resolves #10015


mcgratta:
Create a simple verification case to ensure that if the ring spans multiple meshes that you conserve total number of particles and HRR.

PR #10043, opened 2021-11-16
FDS Validation Guide: Fix some labels

mcgratta:


mcgratta:
FYI, for Matlab rendering of chemical formulae, it sometimes helps to enclose in brackets, like {CO_2}. This fixes bad spacing within the chemical formula, even though the bad spacing might show up somewhere else in the label.

Also, use m³ instead of m^3 in Matlab labels that are not using the LaTeX font interpreter. The superscript 3 is ALT+0179. Superscript 2 is ALT+0178.

PR #10104, opened 2021-12-07
FDS Source: Additional XYZ coordinate precision in text output

gtabaka:
Change &DEVC location printout to include extra precision comparable to prior format. Currently: XYZ coordinates have only three significant figures of output, but this may be insufficient for domains large in any coordinate or far from the origin. The text line still fits within 80 columns as amended and the output remains in scientific format.


mcgratta:
Great, thanks.

PR #10105, opened 2021-12-07
FDS Source: add RAMP_SPEED_Z and RAMP_DIRECTION_Z for wind profiles

ericvmueller:
Introduces speed and direction profiles for the `WIND` line, which can be combined with temporal ramps. U0 and V0 should now only be used for special cases of flow initialization, not wind-driven flows. 

For now, the synthetic eddy method still uses secondary spatial and temporal ramps for U0 and V0, computed from speed and direction, but this will be addressed in a subsequent pull request. Then RAMP_U0_T and RAMP_V0_T can be deprecated.


rmcdermo:
Eric,
This looks fine to me.  But there are lots of changes, so just make sure you have a test case that is utilizing these calls.  Thanks.  Go ahead merge yourself when you are ready.

PR #10122, opened 2021-12-13
FDS Source: Implement and remove an experimental POISSON_ITERATION_ACCELERATOR

mcgratta:


mcgratta:
The idea of the POISSON_ITERATION_ACCERATOR was to solve a small version of the Poisson equation across mesh interfaces in order to better approximate the pressure at the boundary for the bigger Poisson solve. The scheme reduced the number of Poisson iterations by about a factor of 2, but the communication cost cancelled out the savings. I decided to commit the scheme and then remove it so that we have a record of it should we want to explore it again.

PR #10128, opened 2021-12-14
FDS Source: Remove OpenMP flags from rcal.f90

mcgratta:


mcgratta:
I tried the new version of the Intel Fortran compiler (called the beta version for now, or ifx). This new version will replace the "classic" version in the next year or so. Its advantage is that it works with the new LLVM framework, and most importantly can run on GPUs and CPUs. 

Anyway, I noticed that the new compiler did not like the single OpenMP loop in rcal.f90. I could not figure out what was wrong with the loop, who added it, or whether it made any difference. I then noticed that when one compiles radcal in the radcal GitHub repo, rcal.f90 was not being built with the OpenMP flag. So this OpenMP loop was being used in fds, but not radcal. I decided to remove it, and then cleaned up some inconsistencies in the radcal repo, bringing it up to date with fds and the 2018 Fortran standard.

drjfloyd:
That directive was present in the source we got from Vivien. Removing it just means setup might take a little longer when OpenMP is used for FDS.

mcgratta:
There's a subroutine call within the loop, and I don't know enough about OpenMP to debug it. The new compiler did not indicate what was wrong. It only produced a seg fault at that line. Since the new compiler has GPU compatibility, my guess is that something that maybe we got away with is no longer kosher. I'm not too worried about set up time, but let's keep our eyes on it.

PR #10131, opened 2021-12-14
FDS Source: Apply MAX_REACTION_RATE to r, not rho*r

mcgratta:


shostikk:
Thanks Kevin. This is probably clearer this way.

PR #10144, opened 2021-12-16
FDS Source: Add SET_H to SURF.

drjfloyd:
Adds SET_H to &SURF. Allows one to specify both CONVECTIVE_HEAT_FLUX and TMP_FRONT and have FDS compute h based on the gas temperature. Added to support development of wall heat transfer.


mcgratta:
Does one need to explicitly set SET_H? Can't it be set to T if CON_HEAT_FLUX and TMP_FRONT are both specified?

drjfloyd:
I think it is good to have this be a conscious decision on the part of the User. I think in many cases setting both would an unintended error on the part of a user.

mcgratta:
I think it is going to lead to confusion. Why not just use SET_H internally. Having to set it seems redundant.
```
   IF (ABS(SF%CONVECTIVE_HEAT_FLUX)>TWO_EPSILON_EB) THEN
      IF (.NOT. SET_H .AND. TMP_FRONT >= -TMPM) THEN
         WRITE(MESSAGE,'(A)') 'ERROR: SURF '//TRIM(SF%ID)// ' cannot have TMP_FRONT with CONVECTIVE_HEAT_FLUX'
         CALL SHUTDOWN(MESSAGE) ; RETURN
      ENDIF
      IF (SET_H .AND. TMP_FRONT < -TMPM) THEN
         WRITE(MESSAGE,'(A)') 'ERROR: SURF '//TRIM(SF%ID)// ' need to set TMP_FRONT with SET_H=T and CONVECTIVE_HEAT_FLUX'
         CALL SHUTDOWN(MESSAGE) ; RETURN
      ENDIF
   ENDIF
```

drjfloyd:
I think for our typical user setting both would probably not really be what they need and having the extra keyword prevents the trap.

mcgratta:
What's the harm? The front temperature gets set right and the convective heat
flux gets set right. Is there something more to this, like some
unanticipated outcome?

drjfloyd:
I'll make the change.

drjfloyd:
MY latest PR made this change.

PR #10147, opened 2021-12-16
FDS Source: Remove split particles, N_ORIENTATION>1

mcgratta:


mcgratta:
This commit removes the feature whereby one could "split" a solid Lagrangian particle into halves, quarters, etc, and use discrete values of radiative flux for each piece. This feature was always a bit awkward to apply and of questionable validity. I don't think it is being used, and I'd prefer that it not be used in future. There are too many issues of mass and energy conservation.

PR #10167, opened 2021-12-21
FDS Source: Remove unused inputs

mcgratta:


drjfloyd:
The control function PERCENTILE is in the FDS source. It looks like you added it in August.

mcgratta:
Yup, I was just adding it to the CTRL list of inputs. I documented the
feature in August, but forgot to add the parameter to the input list.

PR #10172, opened 2021-12-22
FDS Source: revert changes to HVAC_TANGENTIAL and put back CONSMAG as…

rmcdermo:
… HVAC near wall turb model


rmcdermo:
@drjfloyd This PR tries to put things back as they were prior to the commit 9316d5b, which you cite in Issue #10075 as being where things changed.  The results for the user's case now seem to be what they were looking for, see below.  However, we should note that even the older results do not give exactly 45 degrees.  Further, if you look at the verification case we have in the user guide under "Tangential Velocity", we discuss that very coarse grid resolution may lead to inaccuracies with such cases.

Please review this PR and if you think we are headed in the right direction, please merge.  I was not sure about some of the changes in the HVAC_TANGENTIAL section in VELOCITY_BC in velo.f90.  But this puts things back as they were.

Screen Shot 2021-12-22 at 1 30 01 PM

drjfloyd:
This looks like the old version.  I see your point that it is a little bit off but not as badly off I think as it was in the simplified case in the issue prior to this PR.

PR #10181, opened 2021-12-28
FDS Validation: Simplify NIST_NRC gridding

mcgratta:


drjfloyd (commented):
These changes put us back to violating our recommended practice in the User's Guide in 10.3.2 "However, keep in mind that the pressure boundary condition on such an opening is imperfect, and it is recommended that if the ﬂow through the doorway or window is important, you should extend the domain a few meters rather than use an OPEN boundary."

For cases with an open door, then open door is now an OPEN VENT rather than a HOLE in a wall with some small portion of the domain on the other side. I feel that our validation inputs should represent best practices as defined in the User's Guide.

mcgratta:
In this case, I don't think the flow through the door has a significant effect on the fire dynamics within the compartment. My bigger concern with these cases was the very complicated gridding and the inclusion of extra pressure iterations that did not seem to have much benefit.

drjfloyd:
People use our validation inputs to learn how to use FDS. I think our inputs should reflect best practice which in this case is not having the door for the room with the fire right on an OPEN boundary.

mcgratta:
I don't consider the input lines below to reflect best practice. Nor do I think it necessary to add little meshes outside of every open door or window of a simulation that considers the fire behavior within a building. It is good to have examples in the Validation Guide of both approaches. For example, in Steckler and Bryant, we include the volume inside and outside of an open door because we are interested in the detailed flow dynamics. In NIST/NRC, we show that one can use an OPEN boundary to represent an open door because mass and energy are still conserved and the detailed flow dynamics at the doorway are probably not all that important to the overall simulation.

I don't like the idea of saying that such and such a technique is wrong or violates some standard of practice. This often leads to overly complicated and long simulations when they are not necessary. I'd be happy to amend the guidance in the User's Guide, but I'd like to keep these old compartment cases relatively simple.
```
&MESH XB=8.908421053,12.79157895,2.112,4.928,0,3.82, IJK=34,24,32/
&MESH XB=0,8.908421053,0,7.04,0,3.82, IJK=39,30,32/
&MESH XB=12.79157895,21.7,0,7.04,0,3.82, IJK=39,30,32/
&MESH XB=8.908421053,12.79157895,0,2.112,0,3.82, IJK=17,9,32/
&MESH XB=8.908421053,12.79157895,4.928,7.04,0,3.82, IJK=17,9,32/

&MESH XB=-1.827368421,0,2.112,4.928,0,2.62625, IJK =8,12,22/
&VENT XB=-1.827368421,0,2.112,4.928,2.62625,2.62625, SURF_ID='OPEN'/
&VENT XB=-1.827368421,0,2.112,2.112,0,2.62625, SURF_ID='OPEN'/
&VENT XB=-1.827368421,0,4.928,4.928,2.62625,2.62625, SURF_ID='OPEN'/
&VENT XB=-1.827368421,-1.827368421,2.112,4.928,0,2.62625, SURF_ID='OPEN'/

&OBST XB=0,0,2.112,2.52,0,2.62625/
&OBST XB=0,0,4.52,4.928,0,2.62625/
&OBST XB=0,0,2.52,4.52,2,2.62625/

&PRES VELOCITY_TOLERANCE=0.001,MAX_PRESSURE_ITERATIONS=50/
```

drjfloyd:
If you were doing a smoke control simulation with specified mechanical exhaust and are using open doors / windows for make-up air, I would agree that extending the domain is not needed. In this circumstance, even if the door is near the fire, the flow rate is being forced by the specified exhaust. 

In this case the opening is for the fire compartment and we don't know (as you say "probably not all that important") the extent to which the OPEN boundary at the edge of the room impacts the conditions in the compartment for each of these fires. In practical usage the FDS predicted layer depth and temperature are generally very important quantities. One can conserve mass and energy and still make really bad predictions.

rmcdermo:
If the simulation results rely on accurate representation of the ventilation rate---usually for an under-ventilated compartment---then we need to have the added volume.  I think that we could add "under-ventilated" to the guide to focus the discussion there.  If the flow is one-way through the opening and well-ventilated, simpler is better.

PR #10189, opened 2021-12-31
FDS Source: special treatment of corner edges

rmcdermo:
This addresses Issue #10075.  Test with the attached input file.  Note that use of CONSMAG for the near wall model is what suppresses the jets that were the original complaint on the issue.  This PR addresses the backward velocity on top of the OBST.

[angle_45a.fds.txt](https://github.com/firemodels/fds/files/7796497/angle_45a.fds.txt)


rmcdermo:
I've been testing the ribbed channel cases and this treatment seems to allow for a simpler version of the SLIP_COEF to be implemented.  But I'm going to run through firebot before merging.

mcgratta:
I've been playing around with this case. I can't really see a downside in what you propose. To help, add some particles:
```
&PART ID='tracers', MASSLESS=T, SAMPLING_FACTOR=1 /
&SURF ID='out',COLOR='BLUE',VEL=-3,VEL_T=0,-3,NO_SLIP=T, NEAR_WALL_TURBULENCE_MODEL='CONSMAG', PART_ID='tracers', DT_INSERT=0.01 /
```

rmcdermo:
I ran the verification suite.  I get "Out of Tolerance" warnings on:

* hole
* HVAC\_flow\_loss (1 m/s)
* wind\_example\_5

Other than this, everything looks fine.  I'll see if there is anything obvious with these cases before merging.

drjfloyd:
HVAC_flow_loss should be fairly dead on. It is three vents tied to a tee with a fixed volume flow. Possible this somehow related to my recent HVAC changes more so than this chage?

drjfloyd:
What did you get for HVAC_flow_loss? I ran with your changes and see 1.002 and 1.997 both of which should pass the 0.005 relative error.

rmcdermo:
It's possible something just did not run.  Or that I made a mistake with my manual firebot run.  I'm inclined to merge this, and deal with whatever issues we see.

drjfloyd:
I agree. I merged.

PR #10204, opened 2022-01-06
FDS Source: Fix minor error in wake reduction

mcgratta:


mcgratta:
Simo, can you check this, and PR #10203.

shostikk:
I cannot understand why L/D-1/2 was changed to L/D-1.

According to Fig 1 of Ramirez paper, s/d = s_0/d - 1/2. Our L is s_0 (center-to-center distance). s is the distance from the rear-surface of the leading sphere to the center of the second sphere.

mcgratta:
If the two spheres touch, isn't L=D? I was also confused by this, but the denominator of those expressions must go to zero as the spheres get closer and closer.

shostikk:
Why should the denominator vanish at L=D? Yes, at L=D the two spheres touch, and the drag has its smallest value. It is not zero, though. The authors of the paper compare their model against another one, and discuss their decision to use the surface-to-center distance instead of center-to-center.

mcgratta:
OK -- I just assumed that the denominator should go to zero. I'll fix.

PR #10242, opened 2022-01-15
Makefile: remove -ipo from Intel macOS; no longer supported

rmcdermo:


mcgratta:
Interesting. Is there an equivalent? -ipo is what makes the FDS release compilation long, but it does speed things up.

rmcdermo:
There does not seem to be for the new macOS.  If anyone finds one, let me know.  We are going to be in a transition period with macOS.  As soon as M1 comes along, we will not even be able to use Intel.  The only other option is to have different release targets for different macOS version... thumbs down on that.

PR #10254, opened 2022-01-21
FDS Source: Change name of compiler directive

mcgratta:
Marcos -- please check this and if it's OK, accept the PR. I am changing the name of a compiler directive. I would prefer not to use __INTEL_COMPILER, but rather something that we explicitly define. Someone had trouble with this.


marcosvanella:
These KMP_ routines are from the intel openmp library. I put them there to get rid of a warning when calling the MKL cluster solver, i.e.:
https://community.intel.com/t5/Intel-oneAPI-Math-Kernel-Library/PARDISO-compatibility-problem-with-new-version-OPENMP/td-p/1169208
We can probably get rid of these KMP_ routine calls altogether.

marcosvanella:
Looks like this issue has been resolved from the intel side. On the pres.f90 file I would just get rid of all lines related to the __INTEL_COMPILER and KMP_* calls. Note that when compiling with the intel compiler this _INTEL_COMPILER preprocessor variable is always set. I find it strange that there would be an error not finding these routines in such case, unless the openmp library being used is not the intel distribution.

Issue #10252, opened 2022-01-21
heated_channel case results are off


rmcdermo:
This happened somewhere between 6.7.7 and now.  I just happened to see this in last night's run.  Need to add quantitative checks.

[heated_channel_Tplus.pdf](https://github.com/firemodels/fds/files/7915969/heated_channel_Tplus.pdf)

rmcdermo:
Here is an image of how these cases looked with version 6.7.5

<img width="350" alt="Screen Shot 2022-02-03 at 2 22 35 PM" src="https://user-images.githubusercontent.com/4418497/152414052-eeee3b94-34e2-4268-b0b6-4f1d6fd516ae.png">

After the fix to the INIT lines #10298 we now have this

<img width="706" alt="Screen Shot 2022-02-03 at 2 24 19 PM" src="https://user-images.githubusercontent.com/4418497/152414339-f7f20977-714c-4f9d-8187-9dc31eb78f1a.png">

rmcdermo:
PR #10371 addresses this issue.  Latest results are shown below.

<img width="842" alt="Screen Shot 2022-03-07 at 11 31 26 AM" src="https://user-images.githubusercontent.com/4418497/157076308-7f5b7b64-b584-4323-9e3e-00f24dd4c17e.png">
<img width="842" alt="Screen Shot 2022-03-07 at 11 31 23 AM" src="https://user-images.githubusercontent.com/4418497/157076309-88f73f8b-1f1b-43d5-a5ba-bb614f047efb.png">

PR #10257, opened 2022-01-22
FDS Validation: change AIR to LJ AIR for heated_channel cases

rmcdermo:


rmcdermo:
I suspect that this is not the only issue with this case.  The problem with the profiles is likely due to the change in the treatment of the WALL_MODEL and the prescribed FORCE_VECTOR.  See Issue #10252

PR #10263, opened 2022-01-24
FDS Source: Clarify use of AIR as a defined SPEC

mcgratta:
Jason, use this little test case to review this PR:
```
&HEAD CHID='junk'/
&MESH IJK=5,5,5, XB=0,1,0,1,0,1/
&TIME T_END=1./
&REAC FUEL='METHANE'/
&SPEC ID='AIR', BACKGROUND=F/
&SURF ID='burner', HRRPUA=500, COLOR='RED'/
&OBST XB=.25,.75,.25,.75,0,.05, SURF_IDS='burner','INERT','INERT'/
&TAIL/
```
Also, is there a reason why unlisted species take on properties of NITROGEN vs some variant of air?


mcgratta:
If these changes are OK, merge the PR. Thanks

drjfloyd:
Changes look OK. 

Using N2 for the default has been there for a very long time. At that time we didn't have the LJ AIR species so there wasn't a simple way to get air properties and using N2 as the default was probably chosen just because it was easy and since N2 is most of air not an unreasonable species to pick if we were to pick just a single species.

PR #10269, opened 2022-01-25
Only run github-actions if certain paths were modified.

marcfehling:
Requested in https://github.com/firemodels/fds/pull/10207#issuecomment-1020548895.

Based on [this example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths).


rmcdermo:
@marcfehling Thanks!

PR #10276, opened 2022-01-25
FDS Source: set CFL_VELOCITY_NORM=1 if MAX_CELL_ASPECT_RATIO > 4

rmcdermo:
In this PR:
* compute MAX_CELL_ASPECT_RATIO per mesh; MPI_ALLGATHERV in main
* set CFL_VELOCITY_NORM = 1 by default, if MAX_CELL_ASPECT_RATIO > 4
* unless user has specified, which will override the default


rmcdermo:
@mcgratta Please merge if you are ok with this

PR #10303, opened 2022-02-04
Shouldn't it be 'Direct'?

TristanHehnen:
Hi!

As the title says, I believe there is a typo in the Users Guide.

Best,
Tristan


drjfloyd:
Thanks!

PR #10335, opened 2022-02-18
FDS Source: Remove ramps for U0, V0, W0

mcgratta:


mcgratta:
Note that this is just a general clean-up of WIND parameters. RAMPs for U0, V0, and W0 are not being used in any V&V cases, and the functionality is replicated with SPEED and DIRECTION RAMPs

PR #10339, opened 2022-02-18
github actions: added `apt-get update`

### LINKED ISSUE #10338: gnu linux workflow action no longer compiler
@marcfehling Looks like there is some problem with the [latest gnu linux compile](https://github.com/firemodels/fds/actions/runs/1867010127).  Could you please take a look at this.  Perhaps we do not need to grab the latest open-mpi.

marcfehling:
Fixes #10338.


rmcdermo:
Thanks!

PR #10342, opened 2022-02-22
FDS Source: Add ability to distribute trees via a SURF line

mcgratta:


emanuelegissi:
May I use this to start working on qgis2fds tree import from FastFuels
website?

Il giorno mar 22 feb 2022 alle ore 23:37 Kevin McGrattan <
***@***.***> ha scritto:

ericvmueller:
@emanuelegissi FYI in addition to this I'm working on a simple python script to convert the 3D data from FastFuels into an input file, which may be of interest and I'm happy to share with you. However, it currently requires the user to manually download the data for a region. I'm not sure the FastFuels server can be directly accessed without login credentials, but maybe you have a way or we can request access.

emanuelegissi:
@Eric Great, I will integrate that into qgis2fds, thank you.
They evidently store their data on a geodb. I would ask them to share that
data as any kind of shapefile. Or just attach our script to their API.
Anyway, not too complicated.

mcgratta:
Ema -- the ability to add random trees to a SURF line is for cases where we do not have specific tree-by-tree information. My understanding of FastFuels is that it would create a very specific distribution of trees/vegetation for a particular place. My addition of INIT_IDS to the SURF line is for a hypothetical case.  For the current qgis2fds script, one could just added INIT_IDS and INIT_PER_AREA to the SURF line that qgis2fds creates, assuming that the numerical grid is fine enough to distinguish individual trees. However, for a large scale simulation involving level sets, the "trees" are implied by the fuel type and we would not add trees to the input file.

ruddymell:
Yes I think that's correct, in FastFuels information defining the location of cells (voxels) that encompass individual tree crowns can be output. This information would be used to define properties of particles in FDS. Surface vegetation (grass, pine needles, shrubs) in FastFuels is currently obtained from LANDFIRE which could be used by the level set model or the boundary fuel model in FDS.

emanuelegissi:
Ok, so if I understand correctly the wildfire simulation approaches could
be:
- pure levelset
- levelset + trees
- boundary fuels + trees
- all particles (trees included)
This is great, because it lets us choose the best tool for the specific
job, available calculation resources, and available input data detail (in
terms of vegetation and wind).

Do I understand well?

@Kevin Yes, you are right, that's already implemented:
https://github.com/firetools/qgis2fds/wiki/Quickstart#step-4-get-and-customize-the-landuse-type-csv-file

@Ruddy @Kevin Ok. I took a look at FastFuels, I downloaded the tree
position and properties csv.
I presume that implementing the automatic generation of single trees would
not take a lot of time.
What I currently miss is how you wish this information to be exported to
fds.
Anyway I think that @Eric's script is going to answer this question.



Il giorno mer 23 feb 2022 alle ore 17:18 ruddymell ***@***.***>
ha scritto:

ericvmueller:
Yup, that sounds like an accurate description of the different simulation approaches.

I focused first on importing the voxel data from FastFuels (if you chose 'export to fire model' from the website), rather than the csv list. That way we can leave the details of generating the tree structure/geometry to the experts (FastFuels). But it should also be relatively easy to replace the trees in the csv file with cones of equivalent mass, if you prefer that way. I will send you some scripts this afternoon.
I should also mention that this all assumes you want to resolve the trees with particles. Collapsing the fuel mass to 2D for pure level set or boundary fuels is not something I have considered yet.

PR #10345, opened 2022-02-23
FDS Source: swap sign for PRESSURE_GRADIENT_FORCE

ericvmueller:
@mcgratta I think there is a sign that needs to be swapped on the pressure gradient force read to match the meteorological convention for direction. But it is fine if you specify a direction ramp.


mcgratta:
Yup, good catch. I had been using FORCE_VECTOR directly in my verification cases, so missed that.

PR #10378, opened 2022-03-08
FDS Source : Move CV linking to SET_CUTCELLS_3D, add velocity computa…

marcosvanella:
…tion from divg in shared faces for linked pressure.


marcosvanella:
Also commented, WIP on logic to block small and unlinked cut-cells.

PR #10440, opened 2022-04-08
FDS Validation: Adjust NIST_Douglas_Fir scripts

mcgratta:


mcgratta:
Eric -- remember to modify the validation case run scripts if you change input file meshes or output file types.

ericvmueller:
Ah, yup - thanks for catching that

PR #10488, opened 2022-04-20
FDS Source: Use actual droplet diameter for HTC calc

mcgratta:


Verification/Sprinklers_and_Sprays/Ranz_Marshall_thermally_thick.fds
```diff
@@ -27,8 +27,7 @@ from Ranz and Marshall "Evaporation from drops - Part II" Chemical Engineering P
       THICKNESS    = 521.5E-6
       TMP_INNER    = 9.1
       CELL_SIZE_FACTOR = 0.01
-      MINIMUM_LAYER_THICKNESS = 5.E-6
-      BLOWING      = T /
```
rmcdermo, line 31: @mcgratta Why are you removing BLOWING?  I believe this is the only case that uses it.  I do not think this is something we want to purge, it's the opposite, we want to move toward this being default.

PR #10501, opened 2022-04-25
FDS Source: Overhaul of convective heat transfer models

mcgratta:


rmcdermo:
Could the `Validation/Vegetation/` directory be `Frankman_Vegetation`?  I worry about having another "Pool_Fires" directory that should really be "VTT_Pool_Fires"

drjfloyd:
Why go back to hard coding the various heat transfer coefficients? It doesn't substantially impact code complexity.

ericvmueller:
I agree with Jason - the custom coefficients were very helpful for the work we did on convection in vegetation beds, and I guess will be helpful for any continued convection studies in the small tunnel at NIST.

mcgratta:
I can add back the coefficients.

PR #10553, opened 2022-05-05
FDS Source: modify ember generation criteria

ericvmueller:
@rmcdermo some possible modifications to the ember generation:
- the density is based on current rather than initial volume (in case things shrink)
- the velocity threshold is replaced with a drag force threshold

Let me know if that works for you or if you want to go another way.


rmcdermo:
@ericvmueller 
* I think the way I had the ember density is ok.  LP%MASS changes with time and R_D changes as the ember shrinks.  The surface density did not work correctly for me.  I'm more comfortable with how I had it, unless you have a case that shows it was wrong.
* I personally think that a velocity threshold is more intuitive.  It could be translated to a drag force internally.  I have a better sense of what 5 m/s means compared to 50 Newtons.

ericvmueller:
Fair enough. I like that force connects directly to the mechanism that leads to generation, but that's just me. I missed the bit with R_D, so I guess the only argument for SURFACE_DENSITY is that we already have the mechanism in place to calculate it ...but I'm not bothered either way.

rmcdermo:
Did you test that surface density works?  If so, it is cleaner and I don't have a problem with it.  I tried it first and it did not work for me, but I could have been doing something wrong.

Also, I am not wed to velocity.  If there ends up being a universal drag threshold then I'm fine using it.  I'm just saying that, as of right now, I do not know what number to put there that would make sense for crude testing.

ericvmueller:
I have a case where SURFACE_DENSITY appears to work for the Douglas fir ...but there's a lot going on there so I can try something simpler to verify. And yeah, I certainly get the case for velocity for now ..all good with me.

Edit: I see there is the ember_1 verification case... so I'll check the results are the same

PR #10570, opened 2022-05-10
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,6 +19,9 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 
+permissions:
+  contents: read
+
 jobs:
   linux-intel-intelmpi:
     # debug build on ubuntu using ifort and intelmpi


marcfehling (approved):
It doesn't hurt to have this check. I think we are not very likely to be prone to an attack through github actions, as the workflow won't be started automatically for first time contributors. However, attackers possibly always find a way if they are determined enough :-)

Didn't know about this feature. I will also add this to the other repositories I'm working on.

Let's see if the actions script succeeds, and then merge.

marcfehling:
@rmcdermo or anyone else: Would you start the workflow manually? Only maintainers can start it.

marcosvanella:
Hi Marc, you want us to merge it?

PR #10609, opened 2022-05-16
FDS Source: Trigger Porterie vegetation model using BETA_CHAR

mcgratta:


mcgratta:
This change makes the parameter `BETA_CHAR` the trigger for using the special Porterie vegetation model. The parameter `NU_O2_CHAR`, which was used previously, has a role to play regardless of any particular kind of char model. It represents the mass of O2 consumed per unit mass of char oxidized. It is needed to get the mass and energy balances right.

PR #10630, opened 2022-05-20
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+      interval: "weekly"


marcfehling (approved):
Another feature I wasn't aware of. Looks good to me.

The question for the FDS maintainers is: do you want `dependabot` to automatically create pull requests in your repository?

rmcdermo:
@marcfehling No, I do not want automatic PRs.

marcfehling:
`dependabot` only creates PRs. You still need to merge them manually.

If you don't want the PRs, then I would suggest to reject this patch.

As an alternative, you can turn on `dependabot alerts`, which only notifies you if `github-actions` is out-of-date - without automatically creating PRs. For this, follow these instructions:
- https://docs.github.com/en/code-security/dependabot/dependabot-alerts/configuring-dependabot-alerts#managing-dependabot-alerts-for-your-repository

rmcdermo:
In reading about dependabot, it sounds like its purpose it to keep code dependancies up to date.  But in your compile tests you do updates for oneapi, gnu, and openmpi.  So, what else could need to be updated?  How else could `github-actions` become out of date?

marcfehling:
It keeps the scripts updated that github uses to checkout code for the workflow, manage caching, etc. These tools are called `actions`.

For example: https://github.com/firemodels/fds/blob/57e3ae0ba24897747ff1cc6c911bc14dfae93d77/.github/workflows/linux.yml#L34 https://github.com/firemodels/fds/blob/57e3ae0ba24897747ff1cc6c911bc14dfae93d77/.github/workflows/linux.yml#L50

`dependabot` - in the configuration suggested in this patch - would check if new versions are available every week, and then open PRs to update these lines. `dependabot alerts` would only notify you about this in the 'security' tab.

rmcdermo:
OK, thanks Marc and Naveen.

PR #10676, opened 2022-06-14
FDS Validation: added Deep Fuel Beds input file generation code

noellecrump2023:
Added a folder for the deep fuel beds case and code that uses swaps.py to generate input files. This request adds the python conversion code, swaps.py template, and the data used to make those fds files to the repository. The files this code creates are far from prefect. Next steps are:

Change excelsior particles form a single rectangle per row to a closer approximation of the experiment shown in the video (preferably, columns of excelsior)
Assess the grid layout such that a higher resolution mesh encases all flames
Achieve a burn that is able to spread past the second bed the way it was able to in earlier versions.
Revisit/recode the ignition VENT such that it is not obstructed at steeper inclines.
Noelle Crump


rmcdermo:
Noelle, let's work on the organization and naming of the directories offline before merging.  Thanks

PR #10679, opened 2022-06-14
Adding USFS Deep Fuel Beds to Validation cases

noellecrump2023:
adds code to generate Deep Fuel Beds fds input files.
the code build_paramfile.py uses exp_params.csv and Template.fds to generate paramfile.csv, wich used with ../../../Utilities/Input_File_Tools/swaps.py  generates fds input files for the burn parameters recorded in exp_params.csv

the input files generated are not finished versions


rmcdermo:
Noelle, let's add a README.md file to the FDS_Input_Files directory to describe the workflow for generating the input files.  Thanks

PR #10698, opened 2022-06-22
FDS Source: fix INIT HRRPUV for Cut Cells

rmcdermo:


rmcdermo:
@drjfloyd @mcgratta 

Looking into this brought to my attention that a _major_ item on the to-do list is to get the Q from cut cells to be properly accounted for CHI_R, both in terms of the correction and the optically thin version.  If GEOM is to be used with coarse grids, then we may have a significant part of the heat release near the surface, and so we can't just hand wave the radiant fraction away as not important.

I am going to try to set up a verification case that makes this obvious.  But at the moment I am very worried about the connection between GEOM and radi.  Just dealing with simple things like the INIT HRRPUV is complicated.   For example, I don't really know how to handle things like the "MODE"s in the INIT HRRPUV function (`ADD_VOLUMETRIC_HEAT_SOURCE(MODE)`).  By the way, adding a comment about what these MODEs are would be helpful.  Thanks

mcgratta:
I added some comments

PR #10700, opened 2022-06-22
FDS Verification: add test for intersection of GEOM and INIT HRRPUV a…

rmcdermo:
…nd particles with heat release


rmcdermo:
The `geom_part_init.fds` case is commented out until we get the last line of the hrr file sorted out.

PR #10709, opened 2022-06-24
FDS Validation: Modified FM_Burner input files

mcgratta:


mcgratta:
These cases used hot particles for ignition, but I noticed that (1) the temperature and HRR of the particles had to be adjusted for each grid resolution and even each fuel, and (2) the particles disrupted the flow field at the burner edge. For the moment, I decided to just use `AIT_EXCLUSION_ZONE` like we do in the UMD_Line_Burner cases. We can revisit the particles later.

PR #10718, opened 2022-06-29
FDS Validation: USFS Deep Fuel Beds add vegetation model

noellecrump2023:


rmcdermo:
@mcgratta Whenever you are ready to put this in a central location, let me know.  Or, if you want us to keep making copies for the vegetation cases, we can do that too.

PR #10722, opened 2022-06-30
FDS Source: update EFFECTIVE HEAT TRANSFER COEFFICIENT output

ericvmueller:
@rmcdermo I was trying the `EFFECTIVE HEAT TRANSFER COEFFICIENT` quantity you added recently but got strange values ...for example, if you add this output to the `particle_heating_convection.fds` verification case.

Here is a suggested modification. It looks like `TMP_F` and `TMP_F_OLD` had the same values (and also there was a flipped sign). But maybe I am just missing something, so let me know what you think.


rmcdermo:
Yep.  Thanks for the catch.

Can you add the output to the verification case so that it gets exercised?

ericvmueller:
Will do

PR #10739, opened 2022-07-06
FDS Source: add PILOT_FUEL_MODEL, allows REAC specific AIT

rmcdermo:


rmcdermo:
In this PR, the PILOT_FUEL_MODEL is an alternative idea to using the AIT_EXCLUSION_ZONE in cases where you have a pilot but still want the AIT of the main fuel to be high to avoid spurious ignition.  Attached is a working input file from the FM_Burner series.  In that test series the burner was ignited with a premixed ethylene pilot ring.  The input file has 36 particles that emit ethylene with an AIT=0 K.  The methane uses AIT=600 C.  There is no AIT_EXCLUSION_ZONE.  The burner seems to light and burn with the appropriate HRR.

[FM_15cm_Burner_CH4_2cm.fds.txt](https://github.com/firemodels/fds/files/9057151/FM_15cm_Burner_CH4_2cm.fds.txt)

Screen Shot 2022-07-06 at 2 17 32 PM

PR #10751, opened 2022-07-11
FDS Source: add HVAC_QFAN routines and clean up H_M_T

drjfloyd:
Updates HVAC_QFAN to be non-beta
Removes HVAC_MASS_TRANSPORT and adds HVAC_MASS_TRANSPORT_CELL_L
Removes DUCT_INTERP_TYPE and add NODE_ID to INIT
Adds a GET_TEMPERATURE function for the various places where we extract a gas temperature from species mass and enthalpy
Cleans up various HVAC verification cases and updates the HVAC_MASS_TRANSPORT cases


bmralph:
A thing of beauty! Thanks for kicking `QFAN` out of beta and licking this into shape Jason!

drjfloyd:
I still need to make some verification cases. I was going to do the ASHRAE 7 quadratic and table cases with qfan to show it gets essentially the same solution. Then I have made a case of three ducts with fans and interconnecting dampers. So as dampers cycle different combinations of fans see each other.

PR #10764, opened 2022-07-13
FDS Source: Fix errors checking for zero area leak path.

drjfloyd:
Also add RADIATION=F to a couple of the HVAC verification cases that were missing it to reduce firebot time.


rmcdermo:
@drjfloyd You committed two new source files hvac_qfan_jacobian.f90 and wall_temp_init.f90.

Is this a mistake?

drjfloyd:
Yes that was an error.  I removed them.

PR #10790, opened 2022-07-15
FDS Validation: Theobald Hose Stream, added breakup length parameter, fixed os move files.

noellecrump2023:
Fixed the `os.system('mv ./burn*.fds ../.')` in `USFS_Deep_Fuel_Beds` as well. Neither of them worked (at least for me) The new solution is set up so that it moves only the files it generated, using the filenames column of the `dfout` dataframe. It's also set up to remove the previous file in the destination directory if it's there, because the move will not work otherwise.


rmcdermo:
Noelle,

The scripts work fine on our cluster.  So something is wrong on your end.  You'll need to rework these cases and explain what is changing.

PR #10795, opened 2022-07-18
FDS Validation: USFS Deep Fuel Beds added make_Run_All.py

noellecrump2023:
`make_Run_All.py` builds a `Run_All.sh` file from all the .fds files in the `FDS_Input_Files` directory by copying `Run_All_base.txt` and writing each files to a new line. It then moves the newly made `Run_All.sh` file to the case directory (in this case, `USFS_Deep_Fuel_Beds`) and then gives that file executable permissions. This script does not work if all the input files are not in the `FDS_Input_Files` directory, and may create a bad Run_All script if any .fds file that is not an input file to be run is in the `FDS_Input_Files` directory.


rmcdermo:
Noelle,

This looks good.  But we probably don't need the base.txt file.  Just write those strings from your python script.

PR #10840, opened 2022-08-02
Added MacOSX workers to 'Github Actions'.

marcfehling:
This patch adds two more workers to 'Github Actions' which compile FDS with `gfortran` and `ifort` on MacOSX systems.

As Intel does not provide a repository for Mac, their compilers need to be installed with their web installers. The script also compiles OpenMPI with `ifort`. Whenever a new version of the Intel compiler or OpenMPI gets released and you want to check FDS against them, then you will need to update the URLS in the `env` section of the script manually. (To be clear, the linux workers do _not_ require this maintenance.)

Feel free to reject this patch if you are not okay with these maintenance tasks. If you'd like to add different features let me know.

@marcosvanella -- I had to use some flags from your script.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -1,0 +1,170 @@
+name: github-osx
+
+
+on:
+  push:
+    paths:
+    - .github/**
+    - Build/**
+    - Source/**
+  pull_request:
+    paths:
+    - .github/**
+    - Build/**
+    - Source/**
+
+
+concurrency:
+  group: ${{ github.event_name }}-${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


marcosvanella:
Marc, very nice. I note the osx intel openmpi target seems to be installing openmpi. This will happen only once for these library versions correct?

marcosvanella:
By installing I meant configuring/compiling/installing.

marcfehling:
The macos workers finish successfully on my fork of FDS.

Building openmpi with the Intel compiler takes about 4 hours. After it has been compiled successfully, the cache routine will write and restore openmpi with every new run -- this just takes a couple of seconds.

We also added MKL to the workers. I will also add MKL to the linux workers in a follow-up pull request.

PR #10867, opened 2022-08-08
Added Windows worker (no cache).

marcfehling:
This patch adds another worker to 'Github Actions' which compiles FDS with ifort, impi, and MKL on Windows systems.

The worker takes about half an hour.

I couldn't get the cache to work yet because, for some reason, Intel doesn't initialize its modules correctly after the cache is restored. I suspect that the Intel installers write something into the Windows registry that I need to add manually.

I will get in touch with the Intel support about the issue.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,71 @@
+name: github-windows
+
+
+on:
+  push:
+    paths:
+    - .github/**
+    - Build/**
+    - Source/**
+  pull_request:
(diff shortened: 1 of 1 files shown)


marcfehling:
I opened a thread in the [Intel forum](https://community.intel.com/t5/oneAPI-Registration-Download/Cache-restore-fails-on-Windows-worker-on-Github-Actions/m-p/1406814). I hope this is the right spot to ask this question.

PR #10941, opened 2022-09-16
copy input file to CHID.inp for use by smokeview

gforney:


drjfloyd:
I suggest keeping the copy routines in dump or putting them in func. If these routines are only being used in dump no need to add a dependency to read that doesn't already exist.

rmcdermo:
@gforney Why do we need a copy of the input file?

rmcdermo:
Also, that commit did not pass the CI checks.  Please revert it and let's discuss.  Thanks

https://github.com/rmcdermo/fds/actions/runs/3068132143/jobs/4955216811#step:8:57

PR #10947, opened 2022-09-19
FDS Source: improve initialization of UII field

rmcdermo:
This PR improves the initialization of UII, UIID, and Q_RAD_OUT to be consistent with any prescribed INIT regions for TEMPERATURE.

Note: If we want to get tight with the radiation initialization for INIT of TEMPERATURE, then we would have to do a radiation solve for Q_RAD_IN *prior* to calling WALL_BC.  Because in WALL_BC we recomputed TMP_F based on Q_RAD_IN and this and leads to the need for 100s of radiation iterations to correct this initial imbalance.


drjfloyd:
In ALLOCATE_BOUNDARY_ONE_D_ARRAYS  Q_RAD_OUT is being set before we have set TMP_F.  Also here we set TMP_F to SF%TMP_FRONT. The actual temperature of the wall cell at T_BEGIN will only be TMP_F if TMP_FRONT = TMPA or TAU_T=0.  I think we should just keep Q_RAD_OUT as it in ALLOCATE_BOUNDARY_ONE_D_ARRAYS, and in init.f90 at the end of INIT_WALL_CELL after we set TMP_F based on TAU_Q or RAMP_Q then set Q_RAD_OUT.

rmcdermo:
OK, I'll fix.

rmcdermo:
Jason,
Please me know if you are ok with the changes.  Feel free to merge, or I can, just let me know.  Thanks

PR #10998, opened 2022-10-11
Initial addition of JH FRA cases to validation guide

johodges:
Primary focus on non-combustible lining cases heat flux measurements.


mcgratta:
You need to commit the output files for the cases run to `firemodels/out` and the experimental data to `firemodels/exp`. Currently our nightly build is broken because of missing files.

PR #11005, opened 2022-10-13
Remove currently unused JH_FRA compartment_23 from matlab vnv input.

johodges:
I think firebot is failing due to figures not being generated. Rather than uploading the out file for the case (not ready for addition yet), I removed it from the matlab input file.


rmcdermo:
Please reset that commit and just change the "d" to an "s".  Thanks

johodges:
Updated commit retains the in-progress lines with an "s" instead of a "d".

PR #11024, opened 2022-10-18
Scaling pyrolysis

johodges:


johodges:
I added time averaging to the scaling-based pyrolysis model. There is a new input parameter on the surf line which controls the window size (HF_SMOOTHING_WINDOW, default = 10 seconds). I also added a new boundary output "SCALING HEAT FLUX" to output the smoothed heat flux used in scaling with time.

Small example case showing the functionality below.
[small.txt](https://github.com/firemodels/fds/files/9814284/small.txt)

mcgratta:
This commit would break the code. The reason is that when you add a variable name to ONE_D, you have to increment the appropriate counter and add the variable name to the list of variables that get packed into a simple data structure for transferring via MPI calls. For example, droplets use the ONE_D derived type, and all the ONE_D variables must be packed into a 1-D array when a droplet crosses mesh boundaries. Check with Jason about this, since he originally wrote that routine.

johodges:
Thanks for the information. I will follow-up with Jason.

PR #11036, opened 2022-10-21
Add time-averaging to cone heat flux scaling

johodges:
Sending this over again. A few notes:

- Jason reviewed the addition of the variable to ONE_D grouping and said it looks right. 
- Simplified the statement in the code to remove unnecessary conditional statements and give the SURF entry a more intuitive name "CONE_FLUX_SMOOTHING_WINDOW".
- The time-averaging is done slightly different than the way it is handled in the particle code. Rather than using a relaxation parameter every time step, we are using a fixed time window which can be set by the user. This should be similar to how the DEVC averaging is done with a user specified time window.


mcgratta:
Did you remember to add the new input parameter to the User's Guide?

johodges:
Thanks for the reminder, I did not add it to the user guide. I'll get the
new input and output added to the guide tomorrow.

PR #11044, opened 2022-10-22
Add new input and output parameters for cone HF scaling model.

johodges:
Added:
- CONE_FLUX_SCALING_WINDOW to SURF parameter input table
- SCALING_HEAT_FLUX to infrequently used outputs table
- Text describing the time averaging, input, and output parameters to Section 11.4.4 (Solid Fuels that Burn at a Specified Rate)


mcgratta:
I merged the pull request, but there seems to be some mismatch in names. Is it `SCALING_HEAT_FLUX` or `CONE_HEAT_FLUX`? In any event, I think it should be `REFERENCE_HEAT_FLUX` to be consistent with other input parameters like `REFERENCE_ENTHALPY` and `REFERENCE_TEMPERATURE`. Then I think it would be better to rename `CONE_FLUX_SCALING_WINDOW` to `REFERENCE_HEAT_FLUX_TIME_INTERVAL`. The common root of these two parameters helps the user better understand that they are related to the same functionality.

johodges:
Thanks for the feedback, I can make this change. Do you foresee any issues
having the input parameter REFERENCE_HEAT_FLUX (representative of flux used
in experiment) and the output REFERENCE_HEAT_FLUX (FDS calculated
INCIDENT_RADIATIVE_HEAT_FLUX + CONVECTIVE_HEAT_FLUX) having the same name?

mcgratta:
I don't think there is an output or variable named `REFERENCE_HEAT_FLUX`.

drjfloyd:
I had named it CONE_HEAT_FLUX when I added it a year or so ago.  For this particular method of pyrolysis, the source of data is likely a cone calorimeter test.

If these names change make sure to also update the related verification cases.

And we should note it in the release notes.

johodges:
I added the time averaged heat flux used in scaling as a boundary output,
SCALING_HEAT_FLUX. From the first message I was thinking you wanted this to
be renamed to REFERENCE_HEAT_FLUX as well.

Issue #10968, opened 2022-09-29
Suggestion: Global vent/leak parameter for pressure zones


FireResearch-BK:
This is an enhancement suggestion stemming from issues with pressure zones / enclosed compartments.

**Situation:**
We get many CAD models from clients, which are not built manifold/solid. These are imported into Pyrosim and ultimately exported as FDS input files. This may generate lots of thin-walled, enclosed compartments, resulting in a lot of pressure zones (in some cases more than 700). These enclosed compartments/zones cause all kinds of pressure-related issues during simulation runs when close to the fire, causing frequent crashes due to numerical instability.

**Attempted workarounds:**
After months of reading the user guide, technical reference guide, user group and bug tracker posts, as well as countless hours of experimentation (not to mention the frustration at seeing a simulation crash after two weeks of computation), we are aware of the following workarounds to deal with pressure zones and enclosed compartments:
- Rebuild geometry in simpler fashion purely with obstacles -> Unfeasible because of the necessary extra time.
- Manually close resulting pressure zones by obstruction -> Zones may be identified by FDS and viewed in Smokeview, but there is no way to dump a human-readable list of zones (with cell locations) from FDS, which makes this approach a time intensive trial & error guessing game.
- Manually assign vents and leakage paths to zones to simulate venting or bursting -> Unfeasible due potentially very high to number of zones and the aforementioned lack of human-readable data related to zone locations.
- Force thickening of geometry when converting into blocks (in Pyrosim) -> Sometimes helps, sometimes doesn't and won't prevent zones from being created in the first place as well.
- Disable pressure zones altogether or use an alternate pressure solver -> Won't prevent numerical instability due to excessive pressure in enclosed compartments as well.
- Material burn-away -> The most logical solution to open enclosed compartments, but very hard to implement, increases computation time by an unfeasible amount, is not required for most of our simulations, and won't prevent crashes from these pressure issues either.

**Suggestion:**
Implement an optional global flag in FDS that automatically deals with excessive pressure in zones or enclosed compartments. Here's some ideas:
- "&PRES ZONE_LEAKAGE=0.00001/": Assign a global leakage rate to the ambient zone (zone 0) to all zones.
- "&PRES ZONE_EQUALIZE=T/": Keep zone pressure equal to the pressure of any surrounding ambient cell at all times.
[...]

PR #11085, opened 2022-11-06
FDS Source: Issue #10968. Change mass flux term at thin obstruction

mcgratta:


mcgratta:
Randy, Jason -- this PR is a partial solution to the problem of numerical instabilities arising within small pressure zones with thin boundaries. It is a trade-off of error, not a perfect solution. In the test cases I've looked at, this change prevents the density in the pressure zone from increasing to intolerable levels. 

I did try to add a pressure relief valve to small pressure zones, but that did not solve the problem of high density and low temperature. 

I did run this change through the verification cases and everything passed.

rmcdermo:
@drjfloyd This looks fine to me.  If you are good with it, please merge.

PR #11087, opened 2022-11-07
Update FDS_Verification_Guide.tex

ykim0aumd:
Additional text about numerical errors for Radiation inside a Box case.


rmcdermo:
@ykim0aumd Thank you for the contribution!

PR #11116, opened 2022-11-15
FDS Source: Add enthalpy of formation for bound test

drjfloyd:


rmcdermo:
@drjfloyd Wouldn't it make sense to just make these H_F default to 0?  When we are using fictitious tracers, I don't think we want to have to set thermodynamic properties.  We can throw a warning, just like we do for other properties (assigned to N2 value).

drjfloyd:
There is a REAC specified. With a REAC I think we need the user to say
something about enthalpy as FDS doesn't know what is fake and what is real.

rmcdermo:
I think the most common case is is going to be where users specify a custom fuel species using a FORMULA.  The AIR and PRODUCT compositions will likely be made up of know species from the database and so FDS will then need to compute the H_F for the custom fuel to make it consistent with a specified HEAT_OF_COMBUSTION.

I think that in cases like `bound_test`, having a default of H_F=0 for a custom SPEC is fine (it's what we've been doing and there has yet to be any problem; plus we can put a warning and it is spelled out in the .out file what the heat of formation is).  I'd rather not force changes to input files that are not really necessary.  I think users will bristle at needing to locate a H_F for custom SPEC.

drjfloyd:
The logic is there now to deal with the fuel not being known. In
these cases none of the six species are known and there are two reactions.

PR #11130, opened 2022-11-16
FDS User Guide: update convection section

rmcdermo:


rmcdermo:
@mcgratta Should we think about cross-referencing the convection discussion between the user guide and the tech guide?  We have almost identical information in both.  We could break it out into its own convection.tex file and use and include for both documents if you  really want the same information in both.  Or we could reference the user guide from the tech guide.

Also, now that we are not using C_2, do you want to keep this in the table and available for customization?  Fine with me, just checking.

mcgratta:
Good idea to have one common tex file or we'll get out of sync. We do need this discussion in both guides so that user's know what these input constants fit into. 

Yes, let's keep the NUSSELT_C2.  I'm sure someone will want it sometime.

PR #11132, opened 2022-11-16
Bibliography: remove hrefs causing errors

rmcdermo:


rmcdermo:
@ericvmueller Not sure what was going on with these.  The `make_guide.sh` was passing, but my latex builds were failing from the changes you made.  I don't think we need links to documents in the bibliography.  You can put them up on the directory of validation reports.

PR #11168, opened 2022-11-28
github actions: added another openmpi package to linux

marcfehling:
This should fix the warnings on the `linux-gnu-openmpi` worker:
```
Nonexistent include directory ‘/usr/lib/x86_64-linux-gnu/openmpi/lib/../../fortran/gfortran-mod-15/openmpi’
```


marcosvanella:
Thank you Marc!

Issue #11173, opened 2022-11-29
FDS installed incorrectly on computer cluster


V-bode:
(I asked this question in the discussion group, and was told to post it in the issue tracker. Sorry if this is not the right place for this)

I am running FDS on my university's computer cluster (running CentOS7 Linux), but the submodel relating to solid pyrolysis seems to install wrongly even after multiple attempts

I got the Linux version of FDS from the NIST website (https://pages.nist.gov/fds-smv/downloads.html), which is FDS version 6.7.8 (smokeview 6.7.19).

The input file (attached below as a txt) contains a material A that reacts at a constant rate of 0.2 /s, meaning it should fully react within 5 seconds, but as the figure shows, FDS on the cluster vastly underestimates how quickly it reacts. FDS on my laptop does give the expected result, so it is definitely a problem with the installation, not the model itself.

[enthalpy.txt](https://github.com/firemodels/fds/files/10114057/enthalpy.txt)
![Enthalpy Temp](https://user-images.githubusercontent.com/119428891/204558717-c5449481-d58b-47ca-9585-e7539192529d.png)
![Figure 2022-11-29 143241](https://user-images.githubusercontent.com/119428891/204558734-15fe01f8-60e9-4cd4-a14e-c801a88894e2.png)

mcgratta:
You are in the right place. I'll take a look.

mcgratta:
Are you running these cases with the same version of FDS?

mcgratta:
I think I know what the problem could be. Several versions ago, we changed slightly the expression for reaction rate. I think in the test case called enthalpy.fds, we did not account for this. I'll check.

V-bode:
Now that you mention it
My laptop uses FDS version 6.7.6, whereas the cluster uses 6.7.8 (the version on my laptop was installed a while ago)

I'll rerun the simulation on the cluster in FDS version 6.7.6, and see if that does works

mcgratta:
That's a clue. I should be able to bisect and catch the problem.

mcgratta:
Change A in the input file from 0.2 to 6. There was a change in the reaction rate expression between the two versions you downloaded.

V-bode:
Ah okay, that solves the issue. Thank you very much
I will probably continue using version 6.7.6, even on the cluster, for compatability with my previous simulations

PR #11177, opened 2022-11-29
FDS Source: Issue #11173. Adjust A in reaction rate

mcgratta:


mcgratta:
Change A in the input file from 0.2 to 6. There was a change in the reaction rate expression between the two versions you downloaded.

PR #11226, opened 2022-12-13
FDS Source: WAYPOINTS and NETWORK_ID to HVAC plus write strings

drjfloyd:


mcgratta:
```HVAC/HVAC_filter.err:ERROR: DUCT nas no LENGTH and one NODE lacks an XYZ. Duct: DUCT1 (CHID: HVAC_filter)```

Two problems -- spelling error and failed latest firebot check

drjfloyd:
I need to scrub the verification cases. We had been setting a default LENGTH of 1 m. In general we really shouldn't be setting a LENGTH for non-leak path ducts. I've changed HVAC that if a LENGTH isn't given it will use the node XYZ values if specified (including any WAYPOINTS).

mcgratta:
What if nothing is given? Shall we just throw an ERROR. That might help people better specify networks.

drjfloyd:
The changes are that if no length is given, FDS will use the duct path to get the length if it is explicltly defined. If there are no NODE XYZs and no LENGTH then the error is given.  Getting the length automatically was something  asked for in one of our recent FDS 7 calls.

If there is a LENGTH it will take precedence. This will be reflected in the guide updates I am working on.

PR #11325, opened 2023-01-11
FDS Source: Fill pressure ZONEs

mcgratta:


rmcdermo:
Kevin,
Looks great!  Can we think of a sanity check verification case?  Maybe create a HOLE in an OBST of a certain volume and put a DEVC to check if a PRESSURE ZONE is created?

mcgratta:
Yes, I was going to try that today.

PR #11373, opened 2023-01-24
github-actions: update intel oneapi kits to 2023.0.0

marcfehling:
Follow-up to #10979.

The CI will take a while for the next pull request as caches need to be rebuild. But you will be back to the 5 minute long checks after that is done.


rmcdermo:
Great.  Thanks, Marc!

PR #11376, opened 2023-01-24
FDS Source: Create CELL and EDGE derived types

mcgratta:


mcgratta:
Note that now variables like SOLID, EXTERIOR, EDGE_INDEX are now bundled in a derived type called CELL and TAU_E, OME, IJKE, etc are in EDGE. The definition of these derived types is at the end of type.f90

PR #11379, opened 2023-01-24
FDS Source: Remove IF SOLID from 3D loops

mcgratta:


rmcdermo:
While we are at it... should `DELTA = LES_FILTER_WIDTH_FUNCTION(DX(I),DY(J),DZ(K))` simply be stored in an array instead of recomputed all the time?  I think the ship has sailed on pinching pennies with memory.

mcgratta:
Right. I'll make it a variable called `M%LES_FILTER_WIDTH(I,J,K)` which need only be set once.

PR #11456, opened 2023-02-20
FDS Source: Split BOUNDARY_ONE_D derived type into two

mcgratta:


mcgratta:
This PR splits BOUNDARY_ONE_D so that now only THERMALLY_THICK solids use ONE_D. The other parameters are in BOUNDARY_PROP1. Less widely used properties (formerly BOUNDARY_PROPS) is not BOUNDARY_PROP2. 

Be careful when setting boundary pointers at the start of a boundary loop. The point of this exercise is to lessen the number of parameters associated with various types of boundary cells and particles. For example, a MASSLESS particle now only requires BOUNDARY_COORD.

PR #11487, opened 2023-02-28
Fix error in check for SPECIFIC_HEAT_LIQUID definition.

johodges:
The current code base has two checks for whether a user defines SPECIFIC_HEAT_LIQUID. The second check works as expected (either SPECIFIC_HEAT_LIQUID or RAMP_CP_L must be defined). However, the first check fails if SPECIFIC_HEAT_LIQUID is not found regardless of whether RAMP_CP_L is defined. This pull request removes the first check.


johodges:
See referenced issue with an example input file demonstrating the issue which is addressed in this pull request. https://github.com/firemodels/fds/issues/11488

drjfloyd:
These checks on defintions are often the result of some case that failed at some point. I am reluctant to simply delete the check. That fixes this specific case but I would worry that we break something else.  I think a better option would be to make the check look for the RAMP:

      IF (CP_TEMP < 0._EB .AND. SS%SPECIFIC_HEAT_LIQUID < 0._EB .AND. SS%RAMP_CP_L_INDEX <=0) THEN
         WRITE(MESSAGE,'(A,A,A,A)') 'ERROR: SPEC ',TRIM(SS%ID),' is not predefined and does not have a SPECIFIC_HEAT_LIQUID ',&
            'or a RAMP_CP_L.'
         CALL SHUTDOWN(MESSAGE) ; RETURN
      ENDIF

drjfloyd:
As an additional comment we can pass the LIQUID_IF check by having a TMP_V but no data on specific heat.  So it isn't the case that passing LIQUID_IF means we are good for CP.

johodges:
I believe the check removed in this commit was a duplicate. The modification you proposed is already in line 3783-3785 which were left in the code.

drjfloyd:
3783 is a set of OR statement; it is is checking if any liquid inputs are given. The input file could have no CP information but have a TMP_V and 3783 will pass.  The error checks within the LIQUID_IF block then look to see that a complete set are given.

drjfloyd:
The fix I suggested seems to be working in my fork. I'll nix this PR and do my fix.

johodges:
Thanks for closing

PR #11501, opened 2023-03-03
Scaling pyrolysis

johodges:
Add limits to the scaling heat flux in source code and corresponding description in the user guide.


drjfloyd:
I don't agree with this change. Sure it is unlikely that there is cone data less than 25 or greater than 75 but we shouldn't be limiting Q_IN_SMOOTH to that range. If conditions in the simulation are such that the burning surface sees 85, we should march along the curve at 85 and not cap it at 75. In reality if we had cone data at 85, would we not expect a higher burning rate than at 75? Similarly on the low end this would imply that is something caused the incident flux to drop well below 25 in the simulation that we would still pyrolyze as if it saw 25. Which is also a behavior we would not want.

johodges:
This was a recommendation that came up during the SFPE conference last year in discussion with Kevin and Arnaud. It was initially in the context of the large scale variability in the exposure flux over time which we previously addressed by adding the time averaging to Q_IN_SMOOTH. I generally agree with your comments, especially on the higher end of the exposures where the scaling is a high quality match to experiments. 

The main reason for this change is for the lower end of the exposures. I agree in principle with your comment that the desired behavior without a lower limit would be a natural reduction in the burning rate. However, what prompted this change was a case I am currently analyzing which uses a small igniter which results in low initial exposures to a highly combustible material. Because the initial reference flux is so low (~10 kW/m2), the combustible curve is significantly dampened at the start. This results in lower flame heat fluxes than would be expected in reality leading to slow initial growth.

One compromise which could work while we are testing this feature is to set the min/max to extreme values (say 0/200 kW/m2) so that this feature is available but not working by default.

drjfloyd:
IF (SF%REFERENCE_HEAT_FLUX > 0._EB) THEN
         TSI = MIN(T-T_BEGIN+DT, SF%REFERENCE_HEAT_FLUX_TIME_INTERVAL+DT)
         B1%Q_IN_SMOOTH = (B1%Q_IN_SMOOTH *(TSI-DT) + DT*(B1%Q_CON_F+B1%Q_RAD_IN))/TSI
      ENDIF

Is this the underlying issue? The idea is that the cone curve starts from when a sample ignites. We should only start marching when we hit the ignition temperature. Rather than T-T_BEGIN we should be looking at T-B1%T_IGN and not invoking this at all if B1%T_IGN > T

johodges:
That may be a contributing factor but I don't think it is the root cause. However, I think the long default REFERENCE_HEAT_FLUX_TIME_INTERVAL (10 s) could be a contributor to the behavior I am seeing on this case. I'm re-running my test case with a shorter averaging window to evaluate that impact.

drjfloyd:
I just fixed it so that it doesn't start moving along the cone ramp unless it has reached ignition. For your proposed change we would want the lower to be 0 as a default. One of the verfication tests would fail at 25.  Also, I think that if we are going to have an upper bound disabled as a default that it should be large enough to never kick in. Unintentionally hitting arbirtrary upper limits is often a bane. I suggest setting it to HUGE(EB). But first make sure this is something that is needed and has a good basis.

johodges:
Going to convert this to a draft while I continue to review over this week.

rmcdermo:
Guys, let's get this merged in or closed.  I don't like to have these PRs hanging out too long.  It gets very hard for me to keep track of all the merge commits.  Thanks

johodges:
Jason and I connected on this earlier today and agreed to merge it in with min defaulting to 0 and max to HUGE. The commits added to this pull request make this change and fix a bug on the previous MIN(MAX(... command. I realized I forgot to update the user guide defaults, I will add those now.

johodges:
Added the update to the user guide. This should be ready for merging.

PR #11511, opened 2023-03-07
FDS Source: Fix overflow error in max scaling flux.

drjfloyd:


johodges:
Thanks for the hotfix. Does the same change need to be made in type.f90 line 796?

drjfloyd:
Since we always set SURF%Min/Max_S_H_F using the values in read, we don't need to set default values in type.f90.  They are always overwritten. Probably best to delete the values from type.f90 that way in the future someone isn't confused about which lines of code actually control the defaults.

drjfloyd:
Another option is keep the default as is in type and in read set the values to -1 and then only replace the SF% value if the user has specified a positive value. Now that I am thinking about this more we should probably also add an error check that any values specified are MAX > MIN >= 0.

johodges:
I agree with your change to have -1 in read, apply the fix for HUGE in type, and add the error check in read. Do you want to make that change on your side in this pull request or do you want me to make it on my side and do a new pull request?

mcgratta:
I  just set the value to 1 billion.

PR #11630, opened 2023-04-03
Fix for particle INIT overwriting other init lines.

johodges:
Currently, &INIT lines for particles do not require the user to specify a location and assume the whole domain for XB. As a result, the default values for other gas-phase quantities are being applied by particle INIT lines. In the attached example, the gas temperature is initialized appropriately if the particle line comes before the other INIT line, but gets reset to TMPA if the particle line comes after.
[init.txt](https://github.com/firemodels/fds/files/11140961/init.txt)

This pull request has two fixes for this issue:
1. Added a check for &INIT lines which requires XYZ, XB, DB, or PATH_RAMP to be specified if a PART_ID is defined
2. Omitted INIT lines with a PART_ID from the loops initializing the gas phase properties in the domain.


drjfloyd:
Right now we could have an INIT line where you give a temperature plus put one particle per cell. It seems like this PR would break that.  I think it is good to make sure that if all an INIT has is a PART_ID, that we don't attempt to set any gas phase, but I don't think we should prohibit an INIT from having both particles and gas settings. It may be better to add a check for overlapping XB ranges and write a WARNING that the last INIT will take precedence.

johodges:
That's a good point. I'll make that change to the PR. Ok with adding the check in read (if PART_ID defined, must specify domain info)?

johodges:
Most recent commit removes the check I had added on particles and replaces it with a warning for overlapping INITs. I made it so that the first INIT for an individual cell is the one that is used by the model. This is the same behavior as the logic used in vents with "first come, first serve". However, this is opposite the behavior listed in the user guide. If you are ok with this change I'll modify the user guide to reflect this change.

mcgratta:
I'll leave the merge to Jason, but fix statements like this
```
IF (ALL(ABS(XB)== 1e6) .AND. ALL(XYZ == -1e6)
```
It should be something like this
```
IF (ALL(ABS(XB)>1.E5_EB) .AND. ALL(XYZ<-1.E5) ...
```
In an IF statement, both numbers should be of the same type, and it's not a good idea to use == with real numbers, even though we sometimes make the mistake ourselves. It is better to have an obvious inequality rather than an equality that might not work on all processors.

johodges:
Most recent commit:
1. Changed the IF statement for reals to match Kevin's guidance.
2. Updated user guide to reflect update to INIT to use first come first serve approach rather than last defined takes precedence. I was thinking about this change a bit earlier and I want to verify that it does not impact any of the v&v cases. Taking a look at that now.

johodges:
Most recent commit changes the warning to be printed only once for each combination of INITs in each mesh. The warning also includes identifying information for the two overlapping INITs. Since it is not common to specify an ID for non-particle INITs, I configured it to print the ID if not null, otherwise print the number of the INIT.

This change should not change particle behavior since it is limited to the two loops initializing gas phase temperature, species, and radiation intensity. However, if an INIT line only specifying a particle comes before a line specifying changes to the ambient flow field, the changes to the ambient flow field will be rejected.

Is there a particular reason we want the INIT lines to re-write ambient conditions that are not defined in the line? I could modify the function to only re-write conditions that are defined in the line which avoid these issues.

drjfloyd:
We need particle only lines to not interfere with other INIT. There is the ability in FDS to read in a set inputs using CATF. There are tools under development to predefine INIT for vegation that then get read in to an master input file which could have its own INIT lines for setting properties.

I agree with your idea that we shouldn't overwrite conditions that aren't specified. A MASS_FRACTION only init shouldn't set temperature or density and an INIT with TEMPERATURE or DENSITY only (setting one effects the other) shouldn't reset the gas mass fractions.

johodges:
I'll go ahead and work on that change to INIT to limit changes to what is specified. Could you email me an example input file using CATF that I can use for verification of the change?

If there are no other concerns with this pull request, can we get this one merged in and I'll send another pull request once I have made the broader change to INIT?

drjfloyd:
CATF gets processed first. It creates a new input file that concatenates the original file with content referenced in the CATF inputs. Then it uses that new input file in the rest of input processing as if it was the file you started with. 

The attached case has one INIT for mass fraction, one overlapping INIT for either temperature or density (should work if either is selected but both should result in the preference rule), and two overlapping INIT for different particles. Where the particle INIT lines are in the file should not have an effect on the outcome (i.e., the two INIT for particles before, after, or one on either side of the MF and T/RHO INIT lines).

[part .txt](https://github.com/firemodels/fds/files/11161402/part.txt)

drjfloyd:
This PR still doesn't seem to address alllowing mutliple overlapping particle regions. Can you check make sure this doesn't break the Validation\NIST_Douglas_Firs before I merge?

johodges:
Particle initialization is handled in SUBROUTINE INSERT_VOLUMETRIC_PARTICLES (line 913 of part.f90) which is not changed in this PR. As far as I can tell SUBROUTINE INITIALIZE_MESH_VARIABLES_1 does not interact with that routine. They are both called separately in main.f90. 

Running the douglas fir validation case before and after the PR to verify.

drjfloyd:
Thanks for double checking. You are right I don't think there is any risk this breaks those cases. I'll merge.

johodges:
Could you take a look at the recent commit on my fork? I updated the attached part.txt file to have additional devices and inits for testing. There are comments in the file on the expected behavior. The output device values look right except for the integrated intensity on which I am unclear. FDS is outputting 6.14 for all regions even though the temperature is set correctly and when I write the values to screen they are changing.

https://github.com/johodges/fds/
[part.txt](https://github.com/firemodels/fds/files/11163197/part.txt)

PR #11633, opened 2023-04-04
github-actions: update intel oneapi kits to 2023.1.0, update openmpi to 4.1.5

marcfehling:
Follow up to #11373.

The CI will take a while for the next pull request as caches need to be rebuild. But you will be back to the 5 minute long checks after that is done.


marcosvanella:
Thanks Marc!

PR #11634, opened 2023-04-05
FDS Utilities: Remove module loading and unloading from qfds.sh

mcgratta:


mcgratta:
Glenn -- note that `qfds.sh` still reports which modules are loaded in the `.qlog` file. That might help us diagnose issues. But modules are no longer purged or loaded by `qfds.sh`.

PR #11636, opened 2023-04-06
Update INIT to only update gas quantities when specified in the INIT line rather than overwrite with defaults

johodges:
Update to INIT per Jason's last comment in the linked pull request https://github.com/firemodels/fds/pull/11630. Updates in this PR:

- The updated INIT functions allow for temperature and species to be specified in overlapping regions on separate INIT lines without overwriting with default. The attached input file "part_no_overlap" and "part_overlap" are intended to have the same behavior (except the particles which are commented out in no_overlap). In master they do not as some of the INIT regions overlap and do not take effect. The two input files have the same behavior after the PR. The device names correspond to the intended values for ease of comparison. 
- Fixed a minor bug on the new print warning for overlapping INITs (was printing the wrong INIT number before).
- INIT with many particles should be faster now. Previously, the INIT routine looped through every grid cell for each particle INIT to update temperatures, densities, and species. Now the need to update temperatures, densities, and species is checked outside the 3-nested loop.
[part_no_overlap.txt](https://github.com/firemodels/fds/files/11165899/part_no_overlap.txt)
[part_overlap.txt](https://github.com/firemodels/fds/files/11165901/part_overlap.txt)


mcgratta:
I will test this.

mcgratta:
Problem number 1
```
-rwxr-xr-x. 1 ... Mar  2 16:53 FDS_Cases.sh
```
becomes
```
-rw-r--r--. 1 ... Apr  6 14:38 FDS_Cases.sh
```
How are you editing these files?

mcgratta:
I'll pretend I didn't read that. Avoid all MS and Windows apps as much as possible. If you need to use, say Excel, check the line endings and/or convert to Unix format.

mcgratta:
I ran the verification cases for a few time steps in debug mode. I did not see any errors or Fortran run time errors. I'll let Jason look over the actual changes before committing.

drjfloyd:
I have had good luck with Sublime and SciTE on Windows for editting the sh other text files.  Make sure your git config has:

core.autocrlf=true

johodges:
Fixed the permissions on FDS_Cases.sh, no other bash files were updated. Thanks for the suggestion Jason. I'll modify my workflow to avoid this in the future.

johodges:
Both the user guide and verification guide build now. Needed to fix a missing } in the verification guide so the firebot run started earlier this afternoon will fail there.

PR #11651, opened 2023-04-08
Repeat metric quantity for all metric in dataplot for verification cases. Do not include moving FED and init overlap cases in scatplot for now.

johodges:
The only verification case using the all metric is the new moving FED one, so this will not affect other cases. I changed the FED moving and init overlap cases to "g" for now to mark as still in development and not include in scatplot. Scatplot runs without crashing now, although it warns me of mismatch size error on several cases. I believe these are due to me not having result files for all verification cases locally.


mcgratta:
"all" is used extensively for the validation cases? Have you checked all of those to see if you broke anything?

johodges:
I did not see Save_Measured_Quantity or Save_Predicted_Quantity referenced in the validation section of scatplot. I will verify there is no impact on the validation run today.

mcgratta:
No, in the file FDS_validation_dataplot_inputs.csv, we use "all" for a number of cases. If you changed the functionality of "all", you might have affected the validation plots and stats.

johodges:
The change I made was repeating the stored quantity across the third axis in the struct. I will compare the uncertainty statistics with the current out data with and without this change to identify any differences.

johodges:
I ran through the Matlab validation script at commit level 0f14ba6ae with the update from this PR and with the update commented out. The scatplot .tex and .csv files were generated in both cases. No differences between the .csv and .tex files generated.

PR #11661, opened 2023-04-11
Fix warning for overlapping INITs in radiation loop.

johodges:
Fix for issue https://github.com/firemodels/fds/issues/11659

Excessive warnings were being printed out in the radiation initialization loop. Results should not be affected by this PR. Init_overlap verification case passes and validation case Heil_pre_2ms does not print any warnings.

### DIFF
--- a/Source/init.f90
+++ b/Source/init.f90
@@ -529,10 +529,12 @@ DO N=1,N_INIT
                 M%ZC(K) > IN%Z1 .AND. M%ZC(K) < IN%Z2) THEN
                N2 = M%IWORK1(I,J,K)
                IF (N2 > 0) THEN
-                  IF (.NOT. INIT_OVERLAP1(N,N2)) THEN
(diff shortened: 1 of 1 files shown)


ericvmueller:
Just curious (maybe a dumb question), any reason to put the conditional inside of the mesh loop rather than just after pointing to each INIT, like is done in the previous section (saving some looping):
https://github.com/firemodels/fds/blob/0fe35fea8bbd8f956f22a0364577dea5b51db0c4/Source/init.f90#L464-L467

johodges:
The desired functionality is for each grid cell if any of temperature/density/species is defined to be not default and this is the first time this cell has been checked, update the radiation calc. I think we could have the conditional outside the loop if we also check for IN%ADJUST_SPECIES. We can probably remove the warning print out from that loop entirely. In the case that an overlapping temperature/density INIT region is defined it will be flagged by the prior loop.

Let me take another look at it to confirm.

PR #11685, opened 2023-04-13
First cut at WRITE_STL functionality.

johodges:
This PR adds the functionality to dump OBST lines to an STL file. The STL file can be viewed in Paraview or other CAD packages. Currently no colors are applied to the STL since the standard format does not include colors. The WRITE_STL function is added during the setup only so it can be generated with T_END=0.


Manuals/FDS_User_Guide/FDS_User_Guide.tex
```diff
@@ -11259,6 +11263,7 @@ \section{\texorpdfstring{{\tt DUMP}}{DUMP} (Output Parameters)}
 {\ct STATUS\_FILES}                 & Logical      & Section~\ref{info:DUMP}                &           & {\ct F}                        \\ \hline
 {\ct SUPPRESS\_DIAGNOSTICS}         & Logical      & Section~\ref{info:monitoring_progress} &           & {\ct F}                        \\ \hline
 {\ct VELOCITY\_ERROR\_FILE}         & Logical      & Section~\ref{info:TIMING}              &           & {\ct F}                        \\ \hline
+{\ct WRITE\_STL}                    & Logical      & Section~\ref{info:Paraview}            &           & {\ct F}                        \\ \hline
```
rmcdermo, line 11266: @johodges Where is the "info:Paraview" section?

johodges: good catch, that was intended to be info:STL. Updated in the latest commit.

johodges:
No problem, I'll keep the branch separate for now.

PR #11747, opened 2023-04-17
FDS Verification: Reduce error tolerance for water_pool

drjfloyd:
reducing the error tolerance for water_pool. it was 0.2 which meant we didn't catch a source issue in firebot. Current code is 3 % error so 6 % should pass and provide a little buffer for any minor shifts in the current output.


drjfloyd:
Remainder of this fixes handling of H_F. Missed a case in the logic when updating species and reactions and the Guide needed clarification.

PR #11774, opened 2023-04-26
FDS Source: Add device and control functionality to AIT_EXCLUSION_ZONEs

mcgratta:


mcgratta:
This PR expands slightly the existing control logic for auto-ignition exclusion zones.

PR #11793, opened 2023-05-04
FDS Source: Remove AIR from BOUNDARY_FUEL_MODEL

mcgratta:


ericvmueller:
Checking [ground_vegetation_load.fds](https://github.com/firemodels/fds/blob/master/Verification/WUI/ground_vegetation_load.fds), the in-depth density profile is now based on MATL density rather than MASS_PER_VOLUME on SURF:
image

Which causes issues elsewhere (the boundary fuel no longer burns in that case)

mcgratta:
OK, let me see what's up.

mcgratta:
I think we need to invoke `DENSITY_ADJUST_FACTOR`

ericvmueller:
I had tried to use the calculated `PACKING_RATIO` to adjust the density but I think my issue was I had applied it directly to `LAYER_DENSITY` rather than `DENSITY_ADJUST_FACTOR`

mcgratta:
I propose adding the last line of this chunk of code:
```
         ! Define BOUNDARY_FUEL_MODEL

         IF (SF%PACKING_RATIO(NL)>TWO_EPSILON_EB) THEN
            SF%BOUNDARY_FUEL_MODEL = .TRUE.
            SF%KAPPA_S(NL) = SHAPE_FACTOR*SF%PACKING_RATIO(NL)*SURFACE_VOLUME_RATIO(NL)
         ENDIF

         IF (MOISTURE_FRACTION(NL)>0._EB) THEN  ! Add MOISTURE
            MATL_ID(NL,NN) = 'MOISTURE'
            MATL_MASS_FRACTION(NL,NN) = MOISTURE_FRACTION(NL)/(1._EB+MOISTURE_FRACTION(NL))
            MATL_MASS_FRACTION(NL,1:NN-1) = MATL_MASS_FRACTION(NL,1:NN-1)*(1._EB-MATL_MASS_FRACTION(NL,NN))
         ENDIF

         IF (SF%BOUNDARY_FUEL_MODEL) SF%DENSITY_ADJUST_FACTOR(NL,:) = SF%DENSITY_ADJUST_FACTOR(NL,:)*SF%PACKING_RATIO(NL)
```

ericvmueller:
I think it just needs to also check that PACKING_RATIO(NL)>0. Otherwise if you have something like a soil layer below it affects that too and causes issues.

mcgratta:
This is implied by  `IF (SF%BOUNDARY_FUEL_MODEL)`

ericvmueller:
But that is a whole surface property, not layer by layer

mcgratta:
```
         IF (SF%BOUNDARY_FUEL_MODEL .AND. SF%PACKING_RATIO(NL)>TWO_EPSILON_EB) THEN
            SF%DENSITY_ADJUST_FACTOR(NL,:) = SF%DENSITY_ADJUST_FACTOR(NL,:)*SF%PACKING_RATIO(NL)
         ENDIF
```

ericvmueller:
Let's give it a try and see how the verification cases go

PR #11795, opened 2023-05-04
FDS Source: Adjust density in BOUNDARY_FUEL_MODEL

mcgratta:


ericvmueller:
Looking at that verification case again, the vegetation density now starts off at the correct bulk value, but something happens right away where it re-adjusts (and the layer shrinks). This is 0.5s in, so before any significant heating or reaction should occur.

image

mcgratta:
OK, should be easy to track down where that happens

PR #11818, opened 2023-05-11
FDS User Guide: Fix math error

mcgratta:


mcgratta:
Eric -- this fixes an obvious typo, but not the case overall.

PR #11819, opened 2023-05-11
Allow devices with DB defined to be used if center point is not in a …

johodges:
mesh

### DIFF
--- a/Source/read.f90
+++ b/Source/read.f90
@@ -12657,6 +12657,10 @@ READ_DEVC_LOOP: DO NN=1,N_DEVC_READ
       BAD = .TRUE.
       MESH_LOOP: DO NM=1,NMESHES
          M=>MESHES(NM)
+         IF (DB/='null') THEN
+            BAD = .FALSE.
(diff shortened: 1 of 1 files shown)


mcgratta:
This looks dangerous. Can you submit a simple case where this is needed.

johodges:
This came up in an analysis of a large atrium space. The atrium is fully connected from a lower level up through several stories on one side (see image). We use a statistics device to calculate the free air volume for the whole domain neglecting fully enclosed areas by calculating a volume integral on pressure zone 1. This approach is convenient for larger geometries with fully enclosed rooms (which may not be meshed) or highly congested areas where calculating the free air volume can be tedious and prone to errors. The attached input file demonstrates the approach with all identifying information removed.
atrium

As far as I can tell this check is only used to identify if a device is contained in the domain and is not referred to elsewhere in the code. Devices will still go through other error checking for quantity, etc.

[atrium_mesh_only.txt](https://github.com/firemodels/fds/files/11456021/atrium_mesh_only.txt)

PR #11832, opened 2023-05-15
FDS Source: Use same forced convection model throughout code

mcgratta:


mcgratta:
All invocations of the Nusselt number now come from one subroutine. 

No longer use 300 W/m2/K for HTC between liquid film and solid surface. Use correlation instead.

Issue #11897, opened 2023-06-13
Particles not being culled ax MAXIMUM_PARTICLES


drjfloyd:
In this deluge case MAXIMUM_PARTICLES is reached near 40 s. By 70 s almost no partices exist in the domain. It appears that particles are being culled to make room for new particles but that no new particles are being added.

[Theater_test7.fds.txt](https://github.com/firemodels/fds/files/11733974/Theater_test7.fds.txt)

mcgratta:
OK, I'll fix it.

drjfloyd:
I was going to add this:

LP  => MESHES(NM)%LAGRANGIAN_PARTICLE(LP_INDEX)
LP2 => MESHES(NM)%LAGRANGIAN_PARTICLE(LP_INDEX)
LPC => LAGRANGIAN_PARTICLE_CLASS(LPC_INDEX)

BC_INDEX = LP%BC_INDEX
OD_INDEX = LP%OD_INDEX
B1_INDEX = LP%B1_INDEX
B2_INDEX = LP%B2_INDEX
BR_INDEX = LP%BR_INDEX

LP = LP2

IF (LPC%INCLUDE_BOUNDARY_COORD_TYPE) MESHES(NM)%BOUNDARY_COORD(BC_INDEX)    = MESHES(NM)%BOUNDARY_COORD(LP%BC_INDEX)
IF (LPC%INCLUDE_BOUNDARY_ONE_D_TYPE) MESHES(NM)%BOUNDARY_ONE_D(LP%OD_INDEX) = MESHES(NM)%BOUNDARY_ONE_D(LP%OD_INDEX)
IF (LPC%INCLUDE_BOUNDARY_PROP1_TYPE) MESHES(NM)%BOUNDARY_PROP1(B1_INDEX)    = MESHES(NM)%BOUNDARY_PROP1(LP%B1_INDEX)
IF (LPC%INCLUDE_BOUNDARY_PROP2_TYPE) MESHES(NM)%BOUNDARY_PROP2(B2_INDEX)    = MESHES(NM)%BOUNDARY_PROP2(LP%B2_INDEX)
IF (LPC%INCLUDE_BOUNDARY_RADIA_TYPE) MESHES(NM)%BOUNDARY_RADIA(BR_INDEX)    = MESHES(NM)%BOUNDARY_RADIA(LP%BR_INDEX)
LP%BC_INDEX = BC_INDEX
LP%OD_INDEX = OD_INDEX
[...]

PR #11900, opened 2023-06-13
FDS Source: Issue #11897. Fix bug in REMOVE_OLD_PARTICLES

mcgratta:


drjfloyd:
by resetting to zero won't that cause ALLOCATE_STORAGE to create new storage locatoins for the data rather than just overwriting the existing storage?

PR #11903, opened 2023-06-13
Add more materials to scaling pyrolysis validation, add to dataplot

johodges:


mcgratta:
This is starting to get messy. Send me the compiled pages from the Validation Guide. I'm wondering if we're going too far down the road to reproducing entire databases within the FDS docs.

johodges:
The number of lines is due to defining the reference curves. I am downsampling the experimental curves, but the current approach is pretty simplified. I am selecting the interval to preserve 99% of the peak HRRPUA and 99.9% of the total energy released in the experimental curve. I can make the algorithm more intelligent or manually make adjustments by inspection which will reduce the number of lines significantly.

We are actually seeing very low scatter in the peak HRRPUA across materials (see attached). I think there is some advantage in the breadth of materials being benchmarked with the model to provide confidence in the uncertainty statistics. However, we don't necessarily need to render or show figures for all materials in the guide.

image

drjfloyd:
We shouldn't need anywhere near that fine a timestep in the RAMP to get a decent representation of the cone date. This plot is the FPL waferboard 13 mm data (which in the current input is 0.1 s intervals for 9000+ lines) sampled every 5 s (188 lines). The peak HRR is 236.70 vs. 236.69 sampled and the integration is 107670 vs 107440 sampled (0.2 %  difference). 

image

johodges:
Thanks. I updated the algorithm I was using to down sample and the ramps are much smaller now. The biggest one is around 180 lines with a more typical length of 50-100. I am running these through dataplot now to verify that stats are not affected by the change.

mcgratta:
I cannot imagine why we would need more than 50 points. How crucial is it to the end result?

drjfloyd:
Though with a variable spacing 50 points is probably more than enough. That would mean spacing points such that the change in burning rate between points is no more than 4 % of the peak (assuming the curve is 0 to peak to 0)

johodges:
In the updated algorithm I am hard coding that the peak must be one of the points. I am then calculating the L1 normal between a very coarse representation and the experimental curve. The point with the highest L1 norm is then added to the set to keep and the L1 norm re-computed. The attached figure shows the impact of N on how well the curve is reconstructed. The version with N=96 is the one I have implemented at the moment. To Kevin's point, I think N=48 is good enough, and for most practical purposes N=20 would be fine as well.

image

johodges:
For most materials 50 points yields a max deviation of ~1 kW/m2 between the representative curve and the experiment. There are some materials which have a higher max (FAA polymers PC ~4.1 kW/m2 with 50 points) but the difference is negligible when you compare it to the values in the curve (1% of the value at the time that occurs in the PC case). I propose we set the rule as the fewest points necessary to have the max deviation < 1 kW/m2 up to a maximum limit of 50 points. Thoughts?

mcgratta:
How much does this affect the actual run time? In any case, I wouldn't over-think it. 50 seems more than enough.

mcgratta:
OK, so I think this is just a matter of keeping input files reasonable and not telegraphing the idea that one has to save every single point of the curve.

PR #11924, opened 2023-06-22
FDS Source : Drop zero length segments in surface linearization.

marcosvanella:


Source/geom.f90
```diff
@@ -19681,6 +19681,7 @@ SUBROUTINE GET_CARTCELL_CUTFACES(NM,ISTR,IEND,JSTR,JEND,KSTR,KEND,BNDINT_FLAG)
          SEG_CELL_AUX = SEG_CELL
          COUNT = 0
          DO IDUM = 1,NSEG
+         IF ( SEG_CELL_AUX(NOD1,IDUM)==SEG_CELL_AUX(NOD2,IDUM) ) CYCLE
```
rmcdermo, line 19684: @marcosvanella Shouldn't we do something like:
```
IF ( ABS(SEG_CELL_AUX(NOD1,IDUM)-SEG_CELL_AUX(NOD2,IDUM)) < TWO_EPSILON_EB ) CYCLE
```

marcosvanella: No it's fine. It is checking that the vertex indexes (integers) for a segment entry are not the same.

PR #11957, opened 2023-07-11
FDS Source: Fix array out of bounds error

mcgratta:


mcgratta:
Marcos, take a look at this code change. I ran a terrain case and the variable ICF was -11, which caused a seg fault.

mcgratta:
```
/home4/mcgratta/Test/Terrain
```

PR #11971, opened 2023-07-17
FDS Source: Add special time ramps like a sine function

mcgratta:


drjfloyd:
At this point with the math capabilties of CTRL and the STATISTIC inputs for DEVC we can do most anything in terms of a special ramp without hard coding a new special ramp. You could make a RAMP of sin(a*t+b) with CTRL, DEVC, and RAMP.  sin() is between -1 and 1 so you could have CTRL that evaluates the sin that is then used as the input to a linear RAMP from -1 to 1 with the CTRL as the alternate input. CTRL.  

&DEVC QUANTITY='TIME',ID='TIME'/
&CTRL ID='a*t',FUNCTION_TPYE='MULTIPLY',INPUT_ID='CONSTANT','TIME',CONSTANT=a/
&CTRL ID='a*t+b', FUNCTION_TYPE='SUM',INPUT_ID='a*t','CONSTANT',CONSTANT=b/
&CTRL ID='sin',FUNCTION_TYPE='SIN',INPUT_ID='a*t+b'/
&RAMP ID='ramp sin',T=-1,F=-1,CTRL_ID='sin(at+b)'/
&RAMP ID='ramp sin',T=1,F=1/

mcgratta:
I figured. Is that second RAMP line needed to satisfy the requirement of 2 RAMP lines? In any case, I'll rework my example to use this. As long as we have a template, I don't  mind if it's a little more complicated. I don't want to overload the RAMP line.

drjfloyd:
I had added DEVC_ID and CTRL_ID to be alternate independent inputs to a RAMP. The only way now to get the output of a DEVC or CTRL to act like the dependent value output of a RAMP is to have a linear RAMP that spans the range of values you expect to get out of the CTRL or DEVC (in this case -1 to 1 for sin). I keep finding myself doing this.  It may be useful to have two sets of DEVC/CTRL inputs for RAMP.  One which sets the indepedent value and one that sets the dependent value (I do think it would be useful to keep the alternate input version). With the second then you would just need one RAMP line and no T or F.  Maybe something like CTRL_DEP_ID and DEVC_DEP_ID?

&RAMP ID='ramp sin',CTRL_DEP_ID='sin'/

Then in EVALUATE_RAMP instead of getting a value by table lookup, just set the value to the output of the CTRL or DEVC.

drjfloyd:
I added CTRL_DEP_ID and DEVC_DEP_ID.  I think it will be usefull to have a CTRL or DEVC be either the indepedent or dependent value for a RAMP.  A case where independent is useful is where you have an exhaust that has multiple flow rates based on various ranges. Say no exahust for T < 40 C, 1 m^3/s for 40 C <= T < 60 C and 2 m^3/s above 60.  In your case it would be nice to just have the DEVC or CTRL value be used as the RAMP output without having to make sure you have the correct linear range and the correct NUMBER_INTERPOLATION_POINTS depending on how big the range is.

In doing this noted we were writing RAMP data to the smv file, but Glenn says he isn't doing anything with it so I deleted those lines.

I'll update your example and the guide.

PR #11985, opened 2023-07-21
FDS Source: Add CTRL_ID_DEP and DEVC_ID_DEP to RAMP.

drjfloyd:


mcgratta:
Does the example that I added to the User's Guide need to be changed, or the test case?

Issue #12019, opened 2023-08-03
Flow rate measurements at two different cross sections


Er9y714:
There are a few flow measurement devices in the model. One group is on the orifice, other is across the mesh. I see a slight difference in between for both volume flow rate and the mass flow rate. Is that accepted as normal?
 
Thank you

![image](https://github.com/firemodels/fds/assets/26166281/d2b67de2-5287-4a56-82da-b4fea2ac86d9)
![image](https://github.com/firemodels/fds/assets/26166281/3ed9827e-ecae-4168-b29b-43b19a12c697)
[stairs3.zip](https://github.com/firemodels/fds/files/12252563/stairs3.zip)

rmcdermo:
@Er9y714 can you distill this a bit more, I appreciate it.  Otherwise, I have to dive in an take time to study the case.  Is there an obvious issue?  You know well whether the flows should match.  Not all flows will be identical, but the overall mass balance should be.  You need to do the mass balance.  It may not be locally perfectly steady state.  There are examples in the verification guide to follow.  You may need to tighten up penetration errors at solid boundaries or velocity errors at mesh boundaries, etc.  Please check all this... it is what I would have to do.

Er9y714:
Yes of course. Let me spend some time on it. Good to see I did not miss anything obvious.

mcgratta:
I'll run the case.

mcgratta:
This might be a bug related to the integration of the velocity component. Try this. Instead of 
```
&DEVC ID='AOV_VolFlow_01', QUANTITY='W-VELOCITY', SPATIAL_STATISTIC='AREA INTEGRAL', XB=143.6,144.6,32.8,33.8,54.8,54.8/
```
write this instead
```
&DEVC ID='AOV_VolFlow_01', QUANTITY='W-VELOCITY', SPATIAL_STATISTIC='AREA INTEGRAL', XB=143.61,144.59,32.81,33.79,54.8,54.8/
```
FDS is integrating the value over one too many or one too few cells.

Er9y714:
@rmcdermo Since @mcgratta picked up the issue and found a bug, I guess I do not need to distil the issue anymore. 

@mcgratta 
Thank you. It works well now. It appears that the issue is present both at and away (next to OBST) the mesh boundaries. Limiting the bounding box of DEVC prevents the issue. 

This reminds me a similar issue with INIT regions. I will check and report if I can demonstrate. Thank you very much for your help. I will close the issue.

PR #12022, opened 2023-08-03
FDS Source: Issue #12019. Fix bug related to double precision reals

mcgratta:


mcgratta:
This PR does the following:

1. Change the name `ALMOST_ONE` to `ONE_M_EPS`
2. Add `ONE_P_EPS=1+EPSILON`
3. Fix bug in `read.f90` related to finding bounds for spatial integration.

There are a few places in the code where we want a number to be a little bigger or smaller to avoid computer round-off. Previously, we would typically write `Y - TWO_EPSILON_EB`  to be a number just less than `Y` if it were 12.00000000 and we wanted to round down to 11.999999.  Like when converting `XB` on the `DEVC` line to find the proper cells over which to integrate a spatial stat. Anyway, `EPSILON(X)` is the smallest number for which `1-EPSILON(X)<1` . But if we add or subtract one or two epsilons from a larger number, the result may not be larger or smaller than the number. Instead, it is better to multiply our number by `1-EPS` or `1+EPS` to ensure that the result will be significant. 

Note, this does not affect the vast majority of uses for `TWO_EPSILON_EB` where we just want to know if a number is close to zero.

PR #12023, opened 2023-08-03
FDS Validation: add 2cm GEOM NIST_Pool_Fires cases

rmcdermo:


rmcdermo:
@marcosvanella Please have a look at these cases.  Just pick one.  NIST_Acetone_GEOM_Prescribed_2cm.fds, for example.  When I try to run with multi-mesh, and a GEOM that ran fine with a coarse single mesh, I get a numerical instability right away.  Please have a look.   Note that the cases where I prescribe the GEOM via CYLINDER seem to work.  The ones using the cad bingeom seem to fail.  There are WARNINGs about

```
WARNING: BACK CFACE search, other mesh NOM not equal to working mesh NM. NM=           1 , NOM and other cell IIO,JJO,KKO=           2           1          24          15 , intersection pt=  7.345232278457023E-005 -0.152496389156279       6.249999888232369E-003
```

rmcdermo:
There is a cell going negative in temperature.

Screenshot 2023-08-03 at 5 31 56 PM

marcosvanella:
I reproduced this in optimized mode with and without noise. In dv mode the case runs.
Might have to do with one of my latest changes. I'll look further.

PR #12043, opened 2023-08-10
FDS Source: Use radiation-corrected TEMPERATURE for Smoke3D

mcgratta:


mcgratta:
This PR automatically adjusts the TEMPERATURE field used by Smoke3D to account for radiative fraction. That is, the TEMPERATURE field has been adjusted to account for a user-specified radiative fraction.

Issue #12055, opened 2023-08-15
particle-GEOM interaction


ericvmueller:
This example has a solid particle dropped on a solid. In the first mesh it lands on an OBST and is stationary, but in the second mesh it lands on a GEOM and begins to travel. 
![image](https://github.com/firemodels/fds/assets/13366803/9f2b599d-c0b3-4c81-a772-83ff84df0b58)

This is because of a difference in the use of `LPC%HORIZONTAL_VELOCITY`. In the OBST case this value is ignored if `LPC%ADHERE_TO_SOLID` is false, which is the default for solid particles, but the parameter is not used in the GEOM logic. 

The changes in https://github.com/firemodels/fds/commit/7f55f5b361104d2f4874e3cb7187366c7fc558f9 correct this, but I wanted to give anyone a chance to chime in before I merge ...in case that doesn't agree with the overall logic we want for particle-solid interaction.

[falling_particle_GEOM.txt](https://github.com/firemodels/fds/files/12344883/falling_particle_GEOM.txt)

mcgratta:
Sounds good to  me.

PR #12056, opened 2023-08-15
FDS Source: mimic OBST logic for ADHERE_TO_SOLID for particle on CFAC…

ericvmueller:
…E (dont move if FALSE)


ericvmueller:
resolves #12055

PR #12096, opened 2023-09-04
Natural convection for downward facing hot plate

mcgratta:


mcgratta:
Randy -- take a look at the Verification Guide when you have a chance. I combined all the natural convection cases into a single section with subsections for each configuration. I'm trying to distinguish between the flat, finite plate case and the "enclosure" cases that you have worked on. The HT text books distinguish the flat plate scenario from the enclosure scenario. Also, the cases I added are truly verification cases; that is, they just test what we have coded. Your cases are more like validation cases in the sense that you are comparing results against correlations that are not coded in FDS. I tried to make that point, but double check.

rmcdermo:
Looks good.  I assume these run under validation, hence the reason the git label contains your development branch.  If these run fast, then we could put them under verification.

PR #12115, opened 2023-09-15
FDS Source : Fix UGLMAT issues in multimesh+geometry cases.

marcosvanella:


johodges:
Does this mean we are ready for another fireX test?

marcosvanella:
Yes, I pushed the changes to FireX too.

PR #12220, opened 2023-11-09
update checkerboard case to high emissivity on back and add partition…

johodges:
…ed HT outputs.


mcgratta:
This case is just an example. You can set up your validation cases differently if you like.

PR #12221, opened 2023-11-09
FDS Source: Make NVEC (normal vector) a component of BOUNDARY_COORD

mcgratta:


mcgratta:
This PR moves NVEC from the derived type CFACE to the derived type BOUNDARY_COORD. This means that both WALL and CFACEs can use NVEC in a consistent way.

Marcos, please check geom.f90, ccib.f90, etc.

Issue #12223, opened 2023-11-11
Suggested changes to FDS User Guide


gomyhr:
I have been reading the first part of the user guide, and thought I would submit a pull request for some corrections. But I realized, that in addition to some obvious corrections, I had quite a few "QA-comments", where a problem is identified, but I risk introducing more problems by trying to rewrite those parts myself. So, in case it's useful, I attach a copy of the PDF user guide with some comments from the first few chapters. A few of them are local rewordings that I could probably submit a pull request for. Some other comments I could probably also implement in a PR if I get feedback that the suggestion is valid.

[FDS_User_Guide_commented.pdf](https://github.com/firemodels/fds/files/13326521/FDS_User_Guide_commented.pdf)

mcgratta:
I suggest you make a PR for some of the simple corrections that you have made. Because FDS is evolving, sometimes the User's Guide is missing some things that have been added to the code. Questions like "why do you format this parameter this way?" are difficult to answer because these things have evolved over 25 years. Clarifications about Wordpad or ASCII are helpful. I often do Internet searches on these various terms and definitions, but these things change too. I try to avoid being too explicit about editors and spreadsheets because these discussions may become outdated.

So for things that are obviously wrong, make a PR, and use the Discussions for questions like "why do you do it this way?" Thanks.

gomyhr:
Working on it... I have not been able to build the documentation yet, since I have not found the figures that need to be in the SCRIPT_FIGURES directory. Therefore I have not been able to verify that the changed tex file builds. I have requested access to https://drive.google.com/folderview?id=0B-W-dkXwdHWNOGVsZXNzTjdLek0&usp=sharing which is indicated in Manuals/scripts/copy_fds_figs.sh, in case they can be copied from there.

drjfloyd:
That link is old. You can now get figures at the test release github page:  https://github.com/firemodels/test_bundles/releases.  Go to the FDS release and expand the assets dropdown.

gomyhr:
Thank you drjfloyd, I found all except 3 figures in https://github.com/firemodels/test_bundles/releases (checkerboard_1000, checkerboard and box_burn_away9) in FDS_UG_figures.tar.gz. I got it to build for change verification purposes by copying another png file to those file names. I will move forward with preparing the PR.

PR #12228, opened 2023-11-14
FDS User Guide: Minor fixes and reformulations

gomyhr:
Implementation of the most straightforward changes in the FDS User Guide of those described in the (attachment to) issue #12223.


mcgratta:
Great, thanks.

PR #12254, opened 2023-11-29
FDS Source: Do not allow thick layers on thin obstruction

mcgratta:


drjfloyd:
Is this just for when an OBST XB is less than a single grid cell?

mcgratta:
This is for HT3D, where one specifies THICKNESS and HT3D on a SURF line. That works unless the THICKNESS exceeds the width of the underlying OBST.

PR #12266, opened 2023-12-04
Add CONVECTIVE HEAT FLUX GAUGE output

johodges:
Similar output to radiometer for radiation. For use with validation studies which have split gauge measured heat fluxes into convective and radiative components.


Source/dump.f90
```diff
@@ -9242,7 +9242,14 @@ REAL(EB) FUNCTION SOLID_PHASE_OUTPUT(NM,INDX,Y_INDEX,Z_INDEX,PART_INDEX,OPT_WALL
          ENDDO SH_MATL_LOOP
          SOLID_PHASE_OUTPUT = SOLID_PHASE_OUTPUT * 0.001_EB
       ENDIF
-
+   CASE(76) ! CONVECTIVE HEAT FLUX GAUGE
+         IF (PY%HEAT_TRANSFER_COEFFICIENT>=0._EB) THEN
```
rmcdermo, line 9246: More tab issues?  Please set you editor to use spaces only.

johodges: My editor is set up for spaces only, but when I committed the updates there were some pre-existing tabs which my editor replaced with spaces. I reverted those back to the tabs which were already there thinking they were intentional. Do you want me to commit with those converted to spaces?

rmcdermo: Yes, please go ahead and correct the tabs to spaces issue (per developer guidelines, only spaces should be used).  Thanks

mcgratta:
What file had the pre-existing tabs?

rmcdermo:
> What file had the pre-existing tabs?

Apparently, dump.f90

mcgratta:
I cannot find any tabs in dump.f90

rmcdermo:
@johodges I was only worried about dump.  That is where my comment was.  Not in the user guide.

mcgratta:
There is only one place where we actually need tabs, and this is the makefile. Elsewhere, tabs are a menace.

johodges:
The tabs were only in the user guide .tex file. I do not see any in dump either.

PR #12269, opened 2023-12-05
Add VT Hybrid Heat Flux Experiments validation input files and start section in guide

johodges:
I added one configuration of Wasson's tests with six modeling configurations (default and rayleigh convection each at three grid resolutions). Currently running the 6 cases on our cluster.


mcgratta:
I'd rather not commit to the central repo results that were run with a branch version of FDS. Can you move all of this into a branch of your own?

johodges:
I can do that.

PR #12270, opened 2023-12-05
FDS Source : Initialize variable in read.f90.

marcosvanella:


Source/read.f90
```diff
@@ -13404,6 +13404,7 @@ SUBROUTINE SET_CTRL_DEFAULTS
    ID            = 'null'
    LATCH         = .TRUE.
    INITIAL_STATE = .FALSE.
+   CONTROL_FORCE = .FALSE.
```
rmcdermo, line 13407: @marcosvanella Sorry, this was my fault.

marcosvanella: Ok, thanks.

PR #12384, opened 2024-01-19
Add ability to save binary results to user specified directory

johodges:
Tested on ubuntu and windows.


rmcdermo:
OK, thanks.  I'll start testing it.

PR #12386, opened 2024-01-20
move when trailing slash is added to RESULT_DIR

johodges:


rmcdermo:
Thanks!  I'll test it.

PR #12422, opened 2024-01-31
FDS Source: Fix PROF output and ht3d_sphere cases

mcgratta:


mcgratta:
I added an error message to the ht3d_sphere script.

PR #12431, opened 2024-02-02
FDS vv: change boundary value ini setting for sphere_radiate case to …

gforney:
…set_val (0 instead of 3).  I need to take a further look at how smokeview handles bound settings in ini files.  for now this change to the sphere_radiate.ini file should fix the firebot problem from last night


gforney:
ps:  later when I have time, I will also look at other ini files to see if they have the same issue.

PR #12485, opened 2024-02-15
FDS Validation: Leave plate present but off when spray turns on.

drjfloyd:


mcgratta:
Did anything change except for removing the timer and plate removal?

drjfloyd:
Just timer and plate.  My editor stripped trailing spaces which caused a few lines to show as changed.

mcgratta:
Right -- I didn't realize you had changed the code too. Thanks.

PR #12495, opened 2024-02-16
FDS Source: increase character string length for filenames, needed wh…

rmcdermo:
…en including RESULTS_DIR

I will test this with debug on burn...


rmcdermo:
FYI, running debug cases on burn all seems fine.

drjfloyd:
I just ran all the debug on pele and they passed. Running in release now.

PR #12513, opened 2024-02-21
update spyro write up in tech guide

johodges:


johodges:
Jason has some changes to merge with this in the main body section and in the new appendix.

PR #12530, opened 2024-02-24
FDS Verification: rename expected cases for spyro_cone_demo

rmcdermo:


rmcdermo:
@johodges next time please run Matlab to check that you have captured all the necessary changes.  When you test your guide, do a clean build, else you just pick up old images and the guide thinks it's happy.

johodges:
Thanks for catching.

PR #12564, opened 2024-02-26
FDS Tech Guide: clean up units and subscripts

rmcdermo:


johodges:
I see you changed the \si{} to \rm{} for fixing the subscripts and replaced \si with \unit throughout the appendix. Is the intent to deprecate the \si command from the guides going forward? I am not sure what the difference is under the hood, but I see \si is still used a lot throughout the guides.

rmcdermo:
`\si` seems to be deprecated per the [latest siunitx documentation](https://texdoc.org/serve/siunitx/0).  At least the examples that they give use `\unit` (`\SI` may also be deprecated in favor of `\qty`).

But the main change I made was to use `{\rm text}` for a subscript instead of \si or \unit.  They may give something that looks the same right now, but it is not proper in my mind.  `\rm` may also be deprecated, and we may need to move all these to \textrm{}, technically.

johodges:
Sounds good. I will use that approach going forward.

PR #12635, opened 2024-03-16
Manuals: fix website links in config management plan

rmcdermo:


rmcdermo:
We should rework this document to avoid hyper links and if we have to have one, it should exist in one place.

rmcdermo:
@gforney Whatever software (name and version) you are using to scan the bundle for viruses, let's add a sentence to this document that mentioned that as part of the distribution process.

PR #12638, opened 2024-03-16
FDS Validation Guide: fix fontsize issue intotroduced by SPyro section

rmcdermo:


rmcdermo:
@johodges note that when you want to reduce font size in latex, you do {\scriptsize whatever}, not \scriptsize{whatever}.  It is not a function.

johodges:
Thanks for the info, will do for the next time.

PR #12645, opened 2024-03-18
FDS User Guide: update table of default radiative fractions to match …

rmcdermo:
…source code


rmcdermo:
These were caught by @johodges .  Thanks

PR #12650, opened 2024-03-19
func.f90 - add this close file (LU_EXTERNAL) to avoid problems when c…

RIO4FDS:
add this close file (LU_EXTERNAL) to avoid problems when changing the file by another process


drjfloyd:
You are correct that we need to close the file but that should be done after then ENDDO and before the RETURN. You can't close the file in the middle of a DO loop that is reading the file line by line.  This will result in an error when it goes to read the next line.   I'll make that change and push it up.

drjfloyd:
Take a look at https://github.com/firemodels/fds/commit/5c6299f21347fdad19de061774b14bbec590e66b  
I think I covered the various paths throuh the routine with closing the file when it is open and also saw that I didn't have EXTERNAL_CTRL defined correctly.

PR #12837, opened 2024-04-21
Modifications to V&V scripts to allow them to be run in other contexts

JakeOShannessy:
This PR includes 4 commits to help run the V&V cases in different contexts.

- The first two modify two scripts (`qfds.sh` and `Run_FDS_Cases.sh`) to return an error code on incorrect arguments.
- The third allows `Run_FDS_Cases.sh` to also take a `-e` argument to specify a specific build of FDS.
- The fourth modifies `Run_FDS_cases.sh` to call `qfds.sh` via relative path (in the same way it calls other scripts). This script already errors if not run from the correct directory.

This should not change any behaviour unless the `-e` options is used.


gforney:
Ill take a look

gforney:
jake,
can you send the PR to branch test in my fds fork instead.  it will be easier for me to test.  it might be a few days before I can do this.  its a busy week.
thanks
glenn
https://github.com/gforney/fds/tree/test

gforney:
ignore my previous comment.  I can do what I want to do without you creating a new PR

gforney:
jake,
kevin made some simplifications to qfds.sh which resulted in some conflicts in your PR.  They were just echo statements in the usage routine so shouldn't change what you did.  pretty interesting that I was able to do this with a web editor that github provided
glenn

JakeOShannessy:
Thanks Glenn. The web editor is pretty useful when it's just a single file.

PR #12935, opened 2024-05-14
Reduce fdsplotlib pdf size 

johodges:


rmcdermo:
Cool.  I'll give it a try.  What size were you able to get to?

johodges:
It's 4kb now on my machine compared to the 3kb in Matlab.

johodges:
I found this setting in the "core fonts" section of the matplotlib documentation. https://matplotlib.org/stable/users/explain/text/fonts.html

rmcdermo:
@johodges I reran all the cases this morning and the `1_step_2_step_compare_MLRPUA.pdf` plot is showing 3.2 KB, which is great.  I think we are in business.

PR #12942, opened 2024-05-15
FDS Validation: start folder for Wasson Impinging Plumes

rmcdermo:
@johodges Let's use this to organize the input files for these cases.  Keep inputs similar to this series.  Let's work toward getting all the cases in Table 2.1 into the suite.

There is a different ignition process here, using a ramp with hrrpuv.  These are all stable, but we need to tune the burner temp and AIT to get the hrr distribution right.


johodges:
Thanks, I will take a look at the ignition approach in these files. I'll update the input files from the other tests to match the other changes we have made for test 1 and get them added.

rmcdermo:
Just note that the ignition method is in development.  The 12 mm case did not stay lit---this is the trick, how to make this robust without living on the edge.

PR #13029, opened 2024-06-05
Fix for MASS_FLUX_VAR and HRRPUA output with 2-step chemistry

johodges:
Current behavior on master for these two bugs are:

1. MASS_FLUX_VAR scales the total mass flux variable (MFT) in wall, but does not apply this scaling to the individual component arrays in B1. This bugfix applies the scaling back to the B1 arrays for internal consistency.
2. The HRRPUA output is calculated based on the REACTION(NR)%HEAT_OF_COMBUSTION; however, this does not work for the 2-step chemistry reaction since that is only the HOC of that component. Although we loop through the second step of the reaction, the mass flux of intermediate products is zero at the surface since that reaction occurs in the gas phase. The fix swaps to HOC_COMPLETE to avoid this issue and adds logic to skip the intermediate product reaction in the 2-step chemistry case.


rmcdermo:
@johodges Sorry, I'd like to understand this better before merging.  Let's chat tomorrow.

johodges:
No problem. My schedule is flexible today between 8-930, 1030-130. The two "hrrpua_simple" cases demonstrate the issue with HRRPUA boundary output and the 2-step reaction. The "hrrpua_var" case demonstrates the issue with MASS_FLUX_VAR.

[hrrpua_simple2.txt](https://github.com/user-attachments/files/15613959/hrrpua_simple2.txt)
[hrrpua_simple.txt](https://github.com/user-attachments/files/15613960/hrrpua_simple.txt)

[hrrpua_var.txt](https://github.com/user-attachments/files/15614025/hrrpua_var.txt)

PR #13031, opened 2024-06-06
FDS Source: fix 2 reaction HRRPUA and mass flux variation for individ…

rmcdermo:
…ual species


rmcdermo:
@johodges Please have a look and see if this works with your cases.

johodges:
I can confirm this works for the cases. Thanks!

drjfloyd:
Good catch on the HRRPUA issue. I missed that one when we reworked REAC to allow for multiple simple chemistry reactions.

PR #13035, opened 2024-06-07
scripts: add test so legacy pmi2 parameter is only used on blaze - ie…

gforney:
… not spark


gforney:
you'll need to logout and log back into blaze for these changes to take effect

PR #13047, opened 2024-06-11
FDS Source: Add equivalence ratio check for CVODE call; Add ramp to avoid warning in blaze.

cxp484:


drjfloyd:
I think EQUIV_RATIO_CHECK and the two limits should be inputs available on COMB.

rmcdermo:
@drjfloyd We agree.  This is just the initial push.  We have lots of parameters that need to be documented, and I really want to see the ignition_delay cases working and documented asap.

PR #13070, opened 2024-06-17
add python version of extinction verification plotting

johodges:
I wanted to compare the smoothed version of extinction 2 with existing functionality. Converted the matlab script to python as a start.


johodges:
Sorry about that, I had the check on if the file exists swapped. The updated commit should fix it. Plots should be generated with the same name as the ones from matlab. Slightly larger at ~5kb on my machine than the existing 4kb.

PR #13094, opened 2024-06-25
Chemistry: update workflow for generation of FDS chemistry files

rmcdermo:


drjfloyd:
@rmcdermo  the py file already existed in the python script folder. This commit makes two copies of this in different locations in the repo. I suggest removing form the mechanisms directory and making sure the python folder is correct.

rmcdermo:
Yes, I know.  I think it is better to have this script local to the chemistry mechanisms libraries.  I did not remove the original yet because I did not know if you had a workflow that needed it there.  If not, I'll remove it.

drjfloyd:
I'd vote for keeping python scripts in a common folder. It is easier to find things that way.

PR #13150, opened 2024-07-10
Add BNDF output for 'COMBUSTIBLE MASS'

johodges:
I added this output to help with troubleshooting an issue related to BURN_AWAY, BULK_DENSITY, and HOLEs. The attached case has the output included and demonstrates the issue I am troubleshooting. Planning to submit an issue related to this in a bit but I wanted to be able to refer to this output in the issue tracker.

[burnaway_test2.txt](https://github.com/user-attachments/files/16166966/burnaway_test2.txt)


mcgratta:
You cannot just add another parameter to `B1`. This derived type is used by particles, walls and cfaces and gets packed into bundles for MPI exchange. Not only that, but this variable is going to be included with every particle and wall cells, costing lots of memory. Do we really need this?

rmcdermo:
I don't think so.  Can't you use SURFACE DENSITY?

johodges:
I will try SURFACE_DENSITY and follow up. Closing the PR for now while we discuss.

johodges:
I gave SURFACE DENSITY a try in the case, but the values for it do not update over time. It might be easier to discuss the issue I am seeing and this output at the same time. Do you have a few minutes to hop on a call? My schedule is flexible the rest of the day.

mcgratta:
I'm here now

Issue #8422, opened 2020-06-11
RAMP for DT control


shostikk:
**Motivation**
Every now and then, we have a simulation with a quick transient event that we want to capture with good time resolution (sprinkler activation, blower activation, aircraft impact, covid sneeze,..). If the event is preceeded or followed by long period of relatively low-velocity flow, e.g. wind field generation, controlling DT will increase the simulation time of the whole run. We want to set time dependent control of maximum time step.

**Idea:**
&TIME T_END = 100, DT = 0.1 RAMP_DT = 'MY_RAMP' /
&RAMP ID='MY_RAMP' T = 0.00, F = 1.0 /
&RAMP ID='MY_RAMP' T = 9.90, F = 1.0 /
&RAMP ID='MY_RAMP' T = 10.0, F = 0.1 /
&RAMP ID='MY_RAMP' T = 11.0, F = 0.1 /
&RAMP ID='MY_RAMP' T = 11.1, F = 1.0 /

rmcdermo:
Simo,
Have you tried simply setting `RESTRICT_TIME_STEP=F`?  Also, I think you may be arguing more for control over CFL_MAX, not DT.

shostikk:
Let's say the intial period is proceeding with a time step of 0.1 s. If I want to introduce a transient effect that lasts only 0.1 s, the event would be over before CFL kicks in to reduce DT. (A bit extreme example but you get the point.)

Instead, I as a user want to make sure that DT is a fraction of 0.1 s at the moment when that event starts.

rmcdermo:
OK, I see your point.

shostikk:
Currently, it is possible to deal with this using the flow field intialization or restarts. Doable, but troublesome sometimes.

drjfloyd:
This seems like a good idea that won't have any impact on a typical user.

In the write up I think we need to caution users to make sure that the ramp span and number of interpolation points don't wipe out the time period with the smaller DT.  As an extreme case take Simo's RAMP. If we had N_INTERPOLATION_POINTS=5, then the interpolated values would sit at 2.22,4.44,8.88,11.1 s and we would never see the DT when doing a lookup. 

Also when determining the next time step I think we would want the query to be something like MINVAL(RP%INTERPOLATED_DATA(X:Y)) where X and Y are the interpolated independent points at the beginning and end of the timestep. That way we don't skip over a brief drop if the current timestep is large enough.

mcgratta:
[...]

PR #13205, opened 2024-07-22
FDS Source: Add RAMP_DT to TIME Issue #8422

drjfloyd:


drjfloyd:
Also added in opening and closing bndf, s3d, and part files to keep us from running out of open file units. slcf were already opened and closed. I've run through almost all of the verification suite in release and nothing has thrown a runtime error.  Now the only files left open are the err, out, isof, and csv files.  For production cases where it is generally seconds between file dumps the open and close time is negligible.  If this passes firebot, I can also add in devc and ctrl csv files and try for the isof (that has it's own special write routine in smvv and it was a little less transpartent than the s3d routine so skipped it for now).

rmcdermo:
All looks fine to me.  Just let's add a verification case or something to exercise the RAMP_DT feature.

PR #13225, opened 2024-07-26
FDS Source: Don't double count CHEM time in FIRE time

drjfloyd:


mcgratta:
Aren't we losing track of the CPU time devoted to CHEM if we zero out T_USED(16)?

drjfloyd:
Yikes. Yes. Thanks. It won't be quite that simple.

mcgratta:
I would just turn off the FIRE clock just before calling CHEM and turn it back on when your return.

drjfloyd:
CHEM is called in a select case that picks the integrator in a global loop inside of fire.  To save some CPU cycles and not have to muck with the fire clock in every grid cell, I will add a DTCHEM variable that is global in the fire module. Then in COMBUSTION we zero that out, around the call to CVODE we sum DTCHEM, and at the end of COMBUSTION we up T_USED(16) by DTCHEM and remove DTCHEM from T_USED(10).

mcgratta:
If it gets too complicated, could we just retain FIRE only and assume that it is mostly CHEMistry?

drjfloyd:
I'll make the PR after testing and let you look at it instead of just pushing it through. I think it is pretty simple. CHEM would dominate FIRE when using CVODE for more than a handful of reactions. We had added the CHEM timer in part so we could compare integration time with Cantera. Now that things seem to be working that may be less important.

PR #13229, opened 2024-07-26
FDS Source: Fix edits to T_USED(10) and (16) timers.

drjfloyd:
@mcgratta this seemed to work with a CVODES test case.  FIRE was small, CHEM was big, and the sum over routines matched TOTAL.


mcgratta:
Looks OK. But could you change `DTCHEM`  to `T_USED_CHEM`. It makes things clearer, and I sometimes grep on `T_USED` when working with overall timing data.

mcgratta:
Then just push the change.

drjfloyd:
will do thanks.

PR #13285, opened 2024-08-05
Reduce number of timesteps to write in Cantera output;Add Cantera output csv.

cxp484:


mcgratta:
Not to belabor the point, but can the script truncate the numbers at a precision that is meaningful for the comparison with FDS? That might also help one to read the file more easily by eye.

cxp484:
Yes. I reformatted the Cantera output with timeFormat = '{:10.5f}', OHFormat = '{:.3E}', TMPFormat = '{:10.2f}'. Now it should read better. The time format needs higher precision to show the ignition points.

PR #13302, opened 2024-08-09
FDS Build: Adding SUNDIALS support for WINDOWS

cxp484:
/Qdiag-disable:10448 is to suppress upgrade ifort to ifx message.


gforney:
i'm getting errors listed in attached file when I build with the impi_intel_win_dv target.  did I miss something?
[make.txt](https://github.com/user-attachments/files/16566284/make.txt)

gforney:
what version of the oneapi compiler do you have installed, maybe my version is too old (haven't updated since we had blaze)

cxp484:
Yes. I installed the latest one, this Wednesday.

cxp484:
In that case, my binary files will not work. You probably need to build
Sundials using the oneApi you have. :(

cxp484:
My ifort version is 2021.13.0

C:\MyFiles\NIST_Temp\FDS\fds\Build\impi_intel_win_db>ifort --version
Intel(R) Fortran Intel(R) 64 Compiler Classic for applications running on
Intel(R) 64, Version 2021.13.0 Build 20240602_000000
Copyright (C) 1985-2024 Intel Corporation.  All rights reserved.

gforney:
need to add  $(LFLAGS_SUNDIALS_WIN) to the link line of the impi_intel_win  target.  all the other windows targets have this parameter.  once I did this the fds build worked. (I also succeeded in building sundials after ).  I would do this but am not sure how to add a commit to someone elses PR

cxp484:
Thanks Glenn. Good catch. I comitted the correction. Now you should be able
to see it.

PR #13370, opened 2024-08-28
FDS Source : Write .smv file with mpi-io collective call. Drop SHARED_FILE_SYSTEM flag.

marcosvanella:


marcosvanella:
This #PR changes the way the CHID.smv file is written. It accumulates a string containing all lines of the .smv file for each MPI process and uses a collective mpi write call to generate a single CHID.smv file. This allows us to write a single file efficiently in parallel file systems and large numbers of MPI processes without having to break the file in chunks written per process as was done setting SHARED_FILE_SYSTEM=F.
Note: Keep an eye out for any issues that might come up from this when loading cases in SMV.

PR #13390, opened 2024-09-04
FDS Source: New wall renoding

drjfloyd:
This PR revamps how wall renoding is done.  It passed firebot on our cluster. It is now controlled by two parameters:

1. REMESH_RATIO on SURF - sets the fractional cell size change since the last renode below which no check for renoding is done for that layer.

2. REAC_REAC_DELTA on MATL - sets a limit on the fractional change in pyrolysis rate. If two adjacent wall nodes differ enough in temperature that combining them could shift the pyroysis rate by more than R_R_D, renoding is not done for that layer.

In the process of doing this I noticed that how SWELL_RATIO was being handled for 1D and 3D HT wasn't quite right. This assessment is challegning as it requires figuring out all possible child materials for a material with a reaction. This was being done in two somewhat different manners for 1D and 3D HT.  Instread when MATL is processed each MATL figures out all its possible children and stores that. Then for any SURF_ID or MATL_ID applied to an OBST for 1D or 3D, it becomes easy to just look at the initial materials and assemble the complete list of children.

Marcos, while going through the source and updating things I noticed that in ccib you still had the fixed 5000 K for various property arrays. I changed that to use the I_MAX_TEMP. I don't think this should have broken anything in ccib.


mcgratta:
How could this have passed firebot?

drjfloyd:
This had passed firebot before I left for PTO. Looks like I may have bolloxed the merge with the latest source before making the PR.  I'll close this and resubmit. 

This simplifies the overall wall routine for renoding, greatly reduces the number of times FDS checks to renode (which is costly), and improves the physics basis of when we limit renoding to avoid pyrolysis issues. This is at the penalty of an increase in memory only when pyrolysis is happening. The major speed improvemen comes from not bothering to check for renoding unless wall nodes have dropped enough in size since the last renode. That requires storing the wall cell DX dor the last renode. This is only done when there is pyrolysis. This was discussed in the last dev call that it meant a new item in the wall type.

mcgratta:
It's OJK to add the new array to `ONE_D`. If we are not using it, just give it a dummy length of 1. I don't want to have the `IF` logic in the packing routines. I want to keep them relatively free of code-specific logic or else they will become more difficult to manage. A hit of a byte or two increase in the size of `ONE_D` is OK if we can streamline and speed up the wall routines.

marcosvanella:
> Marcos, while going through the source and updating things I noticed that in ccib you still had the fixed 5000 K for various property arrays. I changed that to use the I_MAX_TEMP. I don't think this should have broken anything in ccib.

Thanks Jason, that should be fine.

drjfloyd:
On second thought, I could add integer variables to the surface type to hold the array sizes for these _OLD variables and set to 1 or the needed value for pyrolysis in read.

mcgratta:
Yes, that would work. These packing routines are used both for MPI exchanges and RESTART. In some sense, we want these "packages" to be fairly free of external logic. That's why we unpack the integers first to know how to unpack the reals.

PR #13395, opened 2024-09-05
FDS Verification: add drag force source term test for stretched grids

rmcdermo:


ericvmueller:
For this one I was thinking it might make sense to add more particles in the first two cells above the surface, since I realized I placed the particle directly at the center of the bottom cell (in z) and it wont do the best job of testing distribution of drag onto the stretched grid.

Maybe something like 
```
&INIT PART_ID              = 'Blade of Grass'
      XB                   = 0.475,0.525,0.475,0.525,0.025,0.075
      N_PARTICLES          = 10/
```

rmcdermo:
OK, I'm building the scaffolding.  We can tighten up once I get the code committed.

PR #13439, opened 2024-09-17
Updates to trilinear interpolation for particles on stretched grids

ericvmueller:


rmcdermo:
@ericvmueller What would you think about changing your block of code to the following:

```
REAL(EB) :: WGT(2,2,2,3),VEL_G(2,2,2)

...

WGT=0._EB
DO AXIS=IAXIS,KAXIS
   IL = IIX; JL = JJY; KL = KKZ
   IF (AXIS==IAXIS) THEN
      IL = FLOOR(XI)
      VEL_G = U(IL:IL+1,JL:JL+1,KL:KL+1)
   ELSEIF (AXIS==JAXIS) THEN
      JL = FLOOR(YJ)
      VEL_G = V(IL:IL+1,JL:JL+1,KL:KL+1)
   ELSEIF (AXIS==KAXIS) THEN
      KL = FLOOR(ZK)
      VEL_G = W(IL:IL+1,JL:JL+1,KL:KL+1)
   ENDIF
   IF (ICC>0) THEN
      CALL GET_FACE_IDW(AXIS,IL,JL,KL,BC%X,BC%Y,BC%Z,WGT(:,:,:,AXIS))
   ELSE
      CALL GET_FACE_TLW(AXIS,IL,JL,KL,BC%X,BC%Y,BC%Z,WGT(:,:,:,AXIS))
   ENDIF
   VEL_G_INT(AXIS) = SUM(VEL_G*WGT(:,:,:,AXIS))
ENDDO
UBAR = VEL_G_INT(IAXIS)
VBAR = VEL_G_INT(JAXIS)
WBAR = VEL_G_INT(KAXIS)
```

This would allow us to change `VEL_G` locally without relying on the full 3D array and should give us more control near a wall.

ericvmueller:
Yeah that makes sense to me... I was thinking as I pushed this that it will probably end up looking different as we add near-wall support, so I'm not wedded to the structure. Feel free to push that up.

PR #13450, opened 2024-09-20
Update JH_FRA cases to new spyro syntax

johodges:


Utilities/Matlab/FDS_validation_dataplot_inputs.csv
rmcdermo, line 1: @johodges I cannot see your changes here without checking out the whole branch, but if this PR were to run it would me you used the "Validation" branch for the git files and this needs to be fixed and rerun.

johodges: I have updated runs of these simulations with my master branch. When's a good time to resubmit these PRs?

rmcdermo: Please send me an email with your compiled validation guide.

rmcdermo: I don't see the UL/FRA cases in the git validation stats table in the back on the guide.  That table appears to be listed chronologically.

mcgratta: UL/FRA has not been added `Process_All_Output.sh` in the `Validation` folder.

johodges: I added it to Process_All_Output in my PR so it should be added to the table when firebot runs. The script assumes all the data files are in Validation/Current_results/ rather than the out repo. I could probably point it to the out repo or copy all the files from out to Current_results to check the generated table.

rmcdermo: The git files should be coming from the out repo.

johodges: Sorry, I see now that the table is generated by validation_git_stats.sh. I'll run that one and recompile the guide.

mcgratta: Process these cases in exactly the same way that all other cases are processed. All results are stored in the out repo. No special scripts pointing to Current_Results, etc.  In fact, if you  can avoid writing a process script, do it. It is much better to let dataplot.m deal with the output and output stats.

PR #13475, opened 2024-09-25
FDS Source: Fix Q_S bug when D_T changes

drjfloyd:
This fixes two things:
1) Didn't have dReacRate/dT correct for liquid pools.
2) In heat transfer we compute Q_S from pyrolysis, heat sources, and internal radiation. Then we check DT_BC_SUB with an explicit solve. If the time step changes, we redo pyrolysis as a new DT_BC_SUB could change values that are defined by DT_BC_SUB. We were just calling pyrolysis again which was adding the new values to the old.  The fix was to store the heat source and IR values seperately so that Q_S could be zeroed if a new call to pyrolysis was needed. Then we just add everything together again.

and makes one aesthetic change of writing H_R for a liquid pool both at TMPA and TMP_BOIL. Before we just had at TMPA which didn't echo the input value for H_R.


mcgratta:
Take a look at PR #13471. I made a one-line change to `wall.f90` to prevent an error due to an uninitialized variable that I have seen in the debug verification cases. The line I added disappeared, and I cannot figure out why.

drjfloyd:
There are two copies of that block of code. One is H_MASS in the DNS block and one is H_MASS_DNS in the LES block. I think you only added that line to one of the blocks.

mcgratta:
I on.ly see one block of code. Around 3224.

drjfloyd:
image

drjfloyd:
I have just made a PR that reworks the IF block so we only need the red outlined block once.

PR #13534, opened 2024-10-07
Python: add example of plotting data on 2 y axes

rmcdermo:


rmcdermo:
@mcgratta Give this a try and let me know.  You obviously need to change the `wrk_dir` if you move the script to a new location.  I can help add legends, etc., as needed.

mcgratta:
OK, thanks. The script works when I uncomment the print statement, but it did not if I uncomment the legend statement. It doesn't recognize ax. Anyway, I'll test it in another directory.

PR #13539, opened 2024-10-07
FDS Source: Add BACK for alternate means of giving MASS_FRACTION_0 with detailed chemistry.

drjfloyd:
For detailed chemistry we will have chemical mechanisms in fds format that can be referenced via CATF. These will have SPEC inputs present but we need to be able to run those libraries with different initial conditions. Rather than trying to store many copies of a mechanism with different MASS_FRACTION_0, trying to read multiple sets of SPEC inputs, etc. , this PR adds the namelist BACK which is processed in read_spec before finalizing SPECIES_MIXTUREs. It is a list of SPEC_ID and MASS_FRACTION that just overwrites any previsously defined MASS_FRACTION_0.


drjfloyd:
@rmcdermo @cxp484

rmcdermo:
Thanks, @drjfloyd .  @cxp484 can you please take the lead on testing this?  Thanks!

cxp484:
Thanks much @drjfloyd.
@rmcdermo Yes. I am going to do test it now.

cxp484:
@drjfloyd @rmcdermo  I tested &BACK using the UMD_Line_Burner and Utilities/Input_Libraries/Chemical_Mechanisms/FDS/Methane_TianfengLu.fds. Everything is working fine. For the Load balancing verification case I am going to use this feature. :)

Issue #13564, opened 2024-10-11
Monin-Obukhov PRofiles - z ref


mcgratta:
### Discussed in https://github.com/firemodels/fds/discussions/13554

<div type='discussions-op-text'>

<sup>Originally posted by **fcurow** October 10, 2024</sup>
To whom it may concern,

Is z_ref taken in absolute or relative terms, eg if the domain is -5m to 5 m in the z direction and there are obstructions from -5 m to 1 m representing the ground and I have a wind value at z = 2m, would the Z_ref in this case be 2m in absolute terms or 1m relative to the obstructions that make up the ground or maybe 7m as it would be relative to the z-min of the domain? The user's guide indicates "If Z_REF is not speciﬁed, it is assumed that the wind speed is taken at a height 2 m off the ground." What is "off the ground" relative to?

Best Regards,
Sal</div>

mcgratta:
`Z_REF` now means the height above the ground. That is, a relative measure of height.

PR #13566, opened 2024-10-11
FDS Source: Issue #13564. Clarify meaning of Z_REF

mcgratta:


drjfloyd:
The new guide text says that if Z_REF isn't given it is taken as 2 m above the ground level but the souce code just has Z_REF=2 not GROUND_LEVEL+2.

drjfloyd:
For Z_REF is there a reason for 2 m? If you are grabbing data from a local weather station those are often taking speed at ~10 m.

mcgratta:
`Z_REF` is a relative value, at least now it is. Its default is 2 m above ground level, which is 0 by default. Not sure why 2 m was chosen.

drjfloyd:
According to the guide text if I were to set GROUND_LEVEL=10 then Z_REF would be 12 (10+2).  In the source, Z_REF would just be 2.  I think that Guide or source should be changed to reflect the other.

10 might be better. The World Meteorlgoical Organization standard for wind measurment is 10 m.

5.9.2	anemometers	over	land 
The standard exposure of wind instruments over level, open terrain is 10 m above the ground.

PR #13594, opened 2024-10-18
FDS Makefile : Add ompi_intel_linux target.

marcosvanella:


marcosvanella:
We'll need to build HYPRE and SUNDIALS for this target once the open-mpi intel module is up.

PR #13601, opened 2024-10-18
Makefile: add hypre and sunials flags for opmi_intel_osx targets

rmcdermo:


rmcdermo:
@cxp484 Can you please double check the LFLAGS for sundials?  Thanks

cxp484:
@rmcdermo I think LFLAGS are good.

PR #13604, opened 2024-10-21
Set Default parameters for CVODE

cxp484:


rmcdermo:
I'll let @drjfloyd review.  To me, it looks like there are a lot of opportunities to remove input parameters that are already defaults.  I'm hoping to get the input file to looking something like this:
```
&COMB ODE_SOLVER='CVODE'/
```
which would set the detailed chem defaults.

cxp484:
Right now the minimum input is:
&COMB
   FINITE_RATE_MIN_TEMP=300,
   ODE_SOLVER='CVODE'/
I mentioned it in the UserGuide (snapshot attached).

image

cxp484:
The default value for FINITE_RATE_MIN_TEMP is -273.15 K. Therefore, I think user must override it to a more reasonable value, unless setting it to 300°C for CVODE is okay (?).

rmcdermo:
I would do this in `READ_COMB`.
```
IF (ODE_SOLVER=='CVODE') THEN
    ! set other defaults
    IF (FINITE_RATE_MIN_TEMP<237._EB) FINITE_RATE_MIN_TEMP=300._EB
ENDIF
```

rmcdermo:
OK, then we need to update the PR to change the documentation and minimize inputs in the input files.

PR #13611, opened 2024-10-23
Build: add scripts to automatically build hypre

rmcdermo:


rmcdermo:
With this PR we are starting the scaffolding for automatically building the hypre library.  For now, this is only applied to the `impi_intel_linux_db` target.

You must have defined the environment variable `$FIREMODELS` on your system where you have your `fds` repo cloned.

The idea is that if you clone [hypre](https://github.com/hypre-space/hypre) into the `$FIREMODELS` directory, then the build script will find this repo and copy the `confmake_impi_intel_linux.sh` script into the appropriate hypre directory and execute the hypre build and install the library into `$FIREMODELS/libs`.

The second time you build fds the hypre step will go much faster, but it is not yet instantaneous.  I am working on gracefully jumping over any hypre building and installing if that library already exists.

PR #13637, opened 2024-10-28
Build: choose c compilers for sundials build

rmcdermo:


rmcdermo:
@cxp484 We need to add the version checkout in the `build_sundials.sh` script.  At the moment, firebot will crash trying to build this.

PR #13651, opened 2024-10-30
FDS Build: Automatic third-party lib build scripts.

cxp484:


rmcdermo:
All: please note that if you are pointing to a local shared version of the SUNDIALS library, for example, on spark we have `/shared/sundials/6.7.0`, then you need to rename `lib64` to `lib`.
```
cd /shared/sundials/6.7.0
sudo mv lib64 lib
...
```

PR #13657, opened 2024-10-30
Build: remove script to set INTEL_IFORT by argument, simplifies build…

rmcdermo:
… scripts


rmcdermo:
The environment variable `$INTEL_IFORT` may still be set in your startup script if desired.

Issue #13715, opened 2024-11-07
GitHub Actions fails due to change in Intel compiler


mcgratta:
I pushed a minor FDS source code change to GitHub, and I received an error from GitHub Actions complaining that `ifort` is not recognized. I suspect that the latest oneAPI package has been installed, and the Intel "classic" Fortran compiler (ifort) is no longer recognized. Can we use an earlier version?

rmcdermo:
I'll try to fix

marcfehling:
The github-action `setup-onapi` has been updated, but it can't find `ifort` anymore. This is not a problem of FDS, but of the github-action.

I am in touch with the developer of the github-action to get it resolved. I'll let you know once we find a solution. Feel free to ignore the failing CI for now.

marcfehling:
I just saw that you updated from ifort to ifx. Has this solved the problem for you?

marcfehling:
FYI, `ifort` has been discontinued in October 2024. We _need_ to switch to `ifx`.
https://www.intel.com/content/www/us/en/docs/fortran-compiler/get-started-guide/2024-0/overview.html

mcgratta:
Yes, we fixed the problem. We now compile with mpiifx in GitHub actions, but we can still use the older compiler for our own use.

PR #13721, opened 2024-11-08
github-actions: update oneapi for windows.

marcfehling:


cxp484:
Hi Marc, I think similar to linux.yml you need to add this line in windows action script, as the makefile by default build with ifort. But, we have set the INTEL_IFORT env variable then it build with that.
export INTEL_IFORT=ifx

For windows it would be probably:
set  INTEL_IFORT=ifx

marcfehling:
I tried to `set INTEL_IFORT=ifx`, but the `make_fds.bat` script was still using `ifort`.
Example: https://github.com/marcfehling/fds/actions/runs/11743376784/workflow#L88

I am trying now to specify `I_IFORT=ifx` by default in the Makefile.
Proposal: https://github.com/marcfehling/fds/blob/cd85d8b9ebd8d437a878fa3a2e7c19814ebadcd3/Build/makefile#L16

I'll open up this PR once I found a working solution.

cxp484:
Yes. I found that the Build/Scripts/setup_intel_compilers.bat set the INTEL_IFORT back to ifort. So our explicit setting never get enforce. I think we need to modify the setup_intel_compilers.bat file. Let me talk to @rmcdermo and @mcgratta to see what should be the best way to do it, because we probably still want ifort to be the default for the time being. For github actions we need ifx to be default.

marcfehling:
See also #13715.

`ifort` has been removed from oneapi with the 2025.0 release. `ifx` has already been around for the last few releases.

I would suggest to make fds "future-proof" and switch to `ifx` everywhere, i.e., adjust all occurrences of
```
fds:$ grep -r "ifort"
```

cxp484:
You're correct that we will eventually need to make the transition. However, we've found that the FDS executables built with ifx are approximately 20% slower than those built with ifort, which is a significant difference. For this reason, we still support ifort. That said, over time, we plan to fully transition to ifx. 

I submitted a changed to Windows actions. Now it looks like working.

PR #13735, opened 2024-11-09
github-actions: cleanup and oneapi 2025.0 release

marcfehling:
In your Makefile, I reverted back from `mpiifx` to `mpiifort`. Further following your suggestions in the oneapi workflows, I specified `INTEL_IFORT` as an environmental variable. I also made some smaller miscellaneous changes along the way.

With this patch, you will compile FDS with `ifx` on linux and windows on the latest oneapi release 2025.0. It turns out you really only need to specify `INTEL_IFORT=ifx`. It appears that Intel MPI then automatically calls `mpiifort -fc=ifx`.

Thank you @mcgratta @rmcdermo @cxp484 for the discussion.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -42,17 +42,15 @@ jobs:
     - name: build fds debug
       run: |
         source /opt/intel/oneapi/setvars.sh
-        cd ./Build/impi_intel_linux_db
         export INTEL_IFORT=ifx
-        echo "INTEL_IFORT =" $INTEL_IFORT
+        cd ./Build/impi_intel_linux_db
         sh ./make_fds.sh
         ./fds_impi_intel_linux_db
     - name: build fds release
       run: |
         source /opt/intel/oneapi/setvars.sh
-        cd ./Build/impi_intel_linux
         export INTEL_IFORT=ifx
-        echo "INTEL_IFORT =" $INTEL_IFORT
+        cd ./Build/impi_intel_linux
         sh ./make_fds.sh
         ./fds_impi_intel_linux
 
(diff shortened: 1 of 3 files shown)


mcgratta:
I would prefer to keep `mpiifx` in the makefile because `mpiifort` is not supported anymore. We can still use the "classic" Fortran compiler `ifort` with `mpiifx` using the environment variable. Everything appears to be working fine now. Unless I am not understanding something, I recommend that we close this PR.

marcfehling:
> I would prefer to keep mpiifx in the makefile because mpiifort is not supported anymore.

Okay, understood. I can remove the first of my commits to just bring the CI up-to-date.

marcfehling:
Please note that you are currently using `ifort` and `mpiifx` by default.
https://github.com/firemodels/fds/blob/master/Build/makefile

https://github.com/firemodels/fds/blob/d57658cd3605dbb6cc867846810e250980ecda4b/Build/makefile#L16

https://github.com/firemodels/fds/blob/d57658cd3605dbb6cc867846810e250980ecda4b/Build/makefile#L244

mcgratta:
Yes, we are using `mpiifix` and `ifort`. We would like to switch over to `ifx`, but we have found that it slows FDS by approximately 20%. This is not unusual. We have read various un-line discussions that indicate that floating-point intensive "engineering" codes, like CFD, see a slow down with `ifx` for various reasons. For the moment, we will continue to use `ifort` until we decide either to switch to `ifx` or look for another compiler.

Randy, the changes look fine to me, but could you double-check.

marcfehling:
I understand your reasoning to switch to `mpiifx` while keeping `ifort`. Thanks for clarifying!

PR #13773, opened 2024-11-18
FDS Build: Remove mpiifort from intel thirdparty builds

cxp484:
The latest OneAPI doesn't have ifort, but kept the mpiifort which by default search for ifort compiler. That will cause a Sundials build failure. Hence removing the mpiifort option.

### DIFF
--- a/Build/Scripts/set_thirdparty_compilers.sh
+++ b/Build/Scripts/set_thirdparty_compilers.sh
@@ -59,7 +59,7 @@ if [[ "$FDS_BUILD_TARGET" == *"osx"* ]]; then
 elif [[ "$FDS_BUILD_TARGET" == *"intel"* ]]; then
     select_compiler CC mpiicx icx mpiicc icc
     select_compiler CXX mpiicpx icpx mpiicpc icpc
-    select_compiler FC mpiifort ifort mpiifx ifx
+    select_compiler FC ifort mpiifx ifx
 else  # Default to GNU compilers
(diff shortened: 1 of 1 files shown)


drjfloyd:
Will this change break compilaiton for those still using ifort?

mcgratta:
It should not. The FDS makefile still looks for `ifort`. If you do not have it because you upgraded to the latest oneAPI, you need to ```
export INTEL_IFORT=ifx
```
in your `.bashrc` file. This is what I am doing .  I am testing the latest oneAPI

cxp484:
I realize my previous comment may not have been entirely clear. This change affects only third-party builds, in particular Sundials. Currently, only Sundials requires a Fortran compiler. By default, the script previously searched for compilers in the following order: mpiifort, ifort, mpiifx, and ifx, prioritizing ifort over ifx. However, the latest OneAPI no longer includes ifort but retains mpiifort, which assumes the presence of ifort as a underlying default compiler. As a result, when building Sundials with the latest OneAPI, the script selects mpiifort as fortran compiler, but since ifort is unavailable, the Sundials build fails.

For building Sundials, ifort alone is sufficient, and mpiifort is not necessary. Hence, I’ve removed mpiifort from the list of third-party Fortran compilers while retaining the order ifort, mpiifx, and ifx.

@drjfloyd To answer your question: ifort will still be supported for third-party builds and will maintain priority over ifx.

PR #13800, opened 2024-11-25
FDS Build: Change Build scripts to allow both ifx and ifort, and keep same fortran compiler for FDS and thirdparty.

cxp484:


cxp484:
Closing the pull request. will create one after fixing the actions.

PR #13801, opened 2024-11-26
FDS Build: Change Build scripts to allow both ifx and ifort, and keep same fortran compiler for FDS and thirdparty.

cxp484:


cxp484:
Actions are working fine now. Closing this pull request without merging due to 20 commits. Will create another pull request from my master branch with a single commit.

PR #13900, opened 2024-12-15
Matlab: revert stripping of hyphens in dvcread

rmcdermo:


rmcdermo:
Turns out that the warnings we were seeing were a bug in Matlab 2023b.  The bug is that there is function literally called "mu" that was getting called in the bowels of str2num.  So, my original hypothesis that the hyphen was the problem was not correct, it was the "mu" in "mu-m".

I upgraded to 2024b and the problem is no longer there.

PR #13928, opened 2024-12-20
FDS Validation: Add a check to Process_All script

mcgratta:


mcgratta:
Glenn, could you take a quick look at this when you have a chance. I tested it, but if you see anything potentially troubling, could you let me know.

PR #13940, opened 2024-12-23
FDS Verification: add 3 component pine_wood_TGA cases to dataplot and…

rmcdermo:
… tune parameters to better match Anca-Couce data


rmcdermo:
@ericvmueller Do these plots live in any of the guides yet?  I am not finding them.  Nor do I find "pine_wood_TGA" label.  So, I'm wondering how we are getting away with this.  Usually firebot complains if a case is in dataplot but not referenced in a section of one of the guides.

ericvmueller:
Hmm AFAIK I didn't do anything special - I didn't know that was something which would cause a complaint and wasn't aiming to avoid it. As far as putting it in the guide I think we had gone back and forth on whether we needed/wanted another pyroylsis case, but with the Kashiwagi case progressing it probably makes sense. Happy to put something in but will need to check with you on the new parameters. I'll be back in the office on Thursday.

PR #13968, opened 2025-01-01
build: add cmake build for FDS

JakeOShannessy:
This PR adds a CMake build for FDS.

With the addition of third-party libraries there are now a few more steps involved in building FDS. The scripts help with this but they necessarily make a few assumptions, including the setup of a the users home directory.

CMake is a useful alternative to this, particularly as it is used by both sundials and hypre, and will likely be used by any other libraries in this space.

The current CMakeLists.txt handles all the current platforms contained in Build/ and can either use sundials and hypre installed in the system or handle the download and build of those libraries directly from GitHub. A few different configuration options are provided at the start of CMakeLists.txt.

I build RPMs for rockylinux and fedora and have found this a more reliable solution. The two build systems can coexist, providing for CMake is simply adding this single file.

To build, run the following:

```sh
cmake -B build-dir -S .
cmake --build build-dir
```

This will result in `build-dir/fds` on most platforms.

Options can be appended in the following way (e.g., to turn off hypre):

```sh
cmake -B build-dir -S . -DUSE_SUNDIALS=OFF -DUSE_HYPRE=OFF
cmake --build build-dir
```

P.S.: Currently CMakeLists.txt points to a forked version of hypre due to `HYPRE_FMANGLE`. Hopefully a PR gets merged to resolve this.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,280 @@
+name: cmake
+env:
+  # update urls for oneapi packages according to
+  # https://github.com/oneapi-src/oneapi-ci/blob/master/.github/workflows/build_all.yml
+  WINDOWS_BASEKIT_URL: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/b380d914-366b-4b77-a74a-05e3c38b3514/intel-oneapi-base-toolkit-2025.0.0.882_offline.exe
+  WINDOWS_BASEKIT_COMPONENTS: intel.oneapi.win.mkl.devel
+  WINDOWS_HPCKIT_URL: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/f07e32fa-b505-4b90-8a79-e328ce9ad9d6/intel-oneapi-hpc-toolkit-2025.0.0.822_offline.exe
+  WINDOWS_HPCKIT_COMPONENTS: intel.oneapi.win.ifort-compiler:intel.oneapi.win.mpi.devel
+
+on:
+  push:
+    paths:
+    - .github/**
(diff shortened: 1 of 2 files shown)


mcgratta:
Thanks, we will take a look at this.

rmcdermo:
@JakeOShannessy Thanks!  When I tested the PR I a saw a lot of warnings about a change in Ubuntu version here:

https://github.com/actions/runner-images/issues/10636

I assume these are innocuous at the moment.  But if we can get ahead of this breaking anything, that would be great.  Thanks again!

rmcdermo:
@JakeOShannessy Is it is standard for `CMakeLists.txt` to be at the top level of the project?  If so, that's fine.  But if this is flexible, can we move it to be under `fds/Build/`?

JakeOShannessy:
> @JakeOShannessy Thanks! When I tested the PR I a saw a lot of warnings about a change in Ubuntu version here:
> 
> [actions/runner-images#10636](https://github.com/actions/runner-images/issues/10636)
> 
> I assume these are innocuous at the moment. But if we can get ahead of this breaking anything, that would be great. Thanks again!

I think they are innocuous but it's a good point so I've switched the images ahead of time. The patch I submitted to hypre has been merged (with subtle changes) so I'll make a few updates and send in a PR.



> @JakeOShannessy Is it is standard for `CMakeLists.txt` to be at the top level of the project? If so, that's fine. But if this is flexible, can we move it to be under `fds/Build/`?

It's definitely the most common to have it at the top level but it's not a hard requirement. It will certainly function but may have an impact on how well it integrates with IDEs and the like so I'll test it.

PR #13994, opened 2025-01-07
Updated oneapi.

marcfehling:


marcfehling:
I'm investigating why the Linux oneapi workflow fails in FDS after a recent update to the `setup-oneapi` action.

rmcdermo:
@marcfehling Hi Marc, so you want me to leave this for now or merge it?

marcfehling:
I found the solution. We updated to the most recent versions of mkl and ifx in the `setup-oneapi` action. However, `ifx` is a few versions ahead of mkl, but it seems like a requirement that they have to be the same version. Manually specifying the version number solves it.

Feel free to merge now!

rmcdermo:
@marcfehling Looks like the cmake actions run failed.

marcfehling:
> @marcfehling Looks like the cmake actions run failed.

```
-- The Fortran compiler identification is unknown
/opt/intel/oneapi/mpi/2021.14/bin/mpiifx: line 715: ifx: command not found
```

That is odd. `ifx` should have been initialized with setvars before.

marcfehling:
I now specified versions of all components. A first test on my repository was successful. Let's see if your CI passes now.

PR #14099, opened 2025-01-21
Python: add pyrolysis.py to special cases

rmcdermo:


rmcdermo:
@gforney This will require that we add
```
# add PYTHONPATH for FDS scripts
export PYTHONPATH=$FIREMODELS/fds/Utilities/Python:$PYTHONPATH
```
to firebot.  I'll give this a try.

PR #14140, opened 2025-01-29
FDS Source: Replacing automatic arrays with thread safe scratch arrays

drfailer:
Some compilers allocate automatic arrays on the heap. This results in multiple calls of `malloc` and `free` during the time stepping which decrease the performance of the simulation.

This PR is a draft to propose an implementation of scratch arrays that would be allocated during the initialization and not during the time stepping.


drjfloyd:
We cannot have I_MAX_TEMP be a parameter. There are use cases (including inputs currently in our repo) where a user might need to increase I_MAX_TEMP.

drfailer:
> We cannot have I_MAX_TEMP be a parameter. There are use cases (including inputs currently in our repo) where a user might need to increase I_MAX_TEMP.

Ok I will change this.

drfailer:
> loops that use ZZ_GET

Normally, each thread uses its sub-array from the scratch memory because I'm allocating one array per thread. I also use `OMP_GET_THREAD_NUM` which returns the index of the current thread.

drfailer:
I also have some questions about what would be better in terms of style/practicality:
- Instead of having multiple 2D arrays, we could have a 1D array (1 element per thread/process) of a type that would encapsulate the scratch arrays.
- I wonder if adding the scratch arrays in `cons.f90` is a good idea since this file in included nearly everywhere. It might be better to put theme in a less used file. Doing so would avoid recompiling everything when adding a new scratch array.
- There is also the option of having general purpose scratch arrays if we can make sure that those are not used at the same time.

drjfloyd:
Have you run a wide range of cases to demonstrate this leads to a significant real improvement in runtime? 

We are now running cases with 1000s of cores with also looking towards leveraging GPUs. These scratch arrays could become large if each process must allocate mulitple scratch arrays over all processes and threads

drfailer:
Note that it depends on the compiler. `ifort` has a `-no-heap-arrays` that is on by default, however, most compilers don't have this feature. Automatic arrays for which the size can't be known at compile time are allocated on the heap which has a huge impact on the performance.

There are other memory management techniques that can be used if the program run out of memory.

Using the GPU will requires such memory management system anyway because you have to copy data in the GPU RAM.

mcgratta:
We have switched over to ifx, which I believe handles OpenMP differently than ifort. We should test this all with ifx.

PR #14329, opened 2025-03-04
FDS Source and User Guide: Switch default SMV parallel write to .FALSE.

marcosvanella:
- This change is done to address issues seen when writing the .smv file in parallel with MPI-IO in Windows OS systems.
- Now All strings containing the different MPI ranks pieces of the .smv file are gathered in rank 0 which performs a POSIX write (single .smv file).
- If need to use MPI-IO writing collectives for large number of MPI processes add SMV_PARALLEL_WRITE=T in the &DUMP line (Section 22 UG).


marcosvanella:
Lets keep an eye on this.

PR #14367, opened 2025-03-13
FDS User Guide: Explain change to C_MIN

mcgratta:


mcgratta:
In this PR, I think I have figured out how to stop the `\lstinline` or `\ct` verbatim entries from extending beyond the right margin. Now it breaks on white space and underscores.

Issue #14387, opened 2025-03-18
OpenMP data race in alternative pressure solvers


mcgratta:
If I run `Pressure_Solver/pressure_iteration3d_uglmat.fds` with 2 OpenMP threads 
```
==================
WARNING: ThreadSanitizer: data race (pid=3761659)
  Write of size 4 at 0x7fff4e78885c by main thread:
    #0 globmat_solver_mp_pressure_solver_check_residuals_u_.DIR.OMP.PARALLEL.LOOP.28567.split8590 /home/mcgratta/firemodels/fds/Build/impi_intel_linux_openmp_db/../../Source/pres.f90:5170:13 (fds_impi_intel_linux_openmp_db+0x3838219) (BuildId: e1d19149ff2bd99998eebe70aac46b5cf88ef07c)
    #1 __kmp_invoke_microtask <null> (libiomp5.so+0x145602)
    #2 globmat_solver_mp_pressure_solver_check_residuals_u_ /home/mcgratta/firemodels/fds/Build/impi_intel_linux_openmp_db/../../Source/pres.f90:5145:7 (fds_impi_intel_linux_openmp_db+0x37a5eea) (BuildId: e1d19149ff2bd99998eebe70aac46b5cf88ef07c)
    #3 fds_IP_pressure_iteration_scheme_ /home/mcgratta/firemodels/fds/Build/impi_intel_linux_openmp_db/../../Source/main.f90:1448:18 (fds_impi_intel_linux_openmp_db+0x45c4f7c) (BuildId: e1d19149ff2bd99998eebe70aac46b5cf88ef07c)
    #4 MAIN__ /home/mcgratta/firemodels/fds/Build/impi_intel_linux_openmp_db/../../Source/main.f90:683:30 (fds_impi_intel_linux_openmp_db+0x459caa3) (BuildId: e1d19149ff2bd99998eebe70aac46b5cf88ef07c)
    #5 MAIN__ /home/mcgratta/firemodels/fds/Build/impi_intel_linux_openmp_db/../../Source/main.f90:79:6 (fds_impi_intel_linux_openmp_db+0x45940b9) (BuildId: e1d19149ff2bd99998eebe70aac46b5cf88ef07c)
    #6 main <null> (fds_impi_intel_linux_openmp_db+0x43a32c) (BuildId: e1d19149ff2bd99998eebe70aac46b5cf88ef07c)

  Previous write of size 4 at 0x7fff4e78885c by thread T1:
[...]

PR #14398, opened 2025-03-19
FDS Source: Issue #14387. Fix OpenMP race condition

mcgratta:


rmcdermo:
@mcgratta Can you please explain.  Is this to tell us to not use NULL() pointers anymore?

mcgratta:
Just a comment -- Marcos just discovered that doing this
```
TYPE(LAGRANGIAN_PARTICLE_CLASS_TYPE), POINTER :: LPC=>NULL()
```
caused an OpenMP data race. I did not realize this, but apparently when a pointer is initialized to null, as here, it automatically assumes a `SAVE` attribute. The subroutine containing this line was called by two OpenMP threads such that `LPC` pointed to the same location in memory. I was told by the thread sanitizer that this variable was global, which made no sense to me until we discovered this `SAVE` feature, meaning that this pointer is now saved outside the routine.

Long story short, I'm not sure we need to nullify pointers upon declaration. I can't remember when and why we started this. Does anyone remember?

mcgratta:
BTW, if we accidently use a pointer before it is associated, the `-check all` debug option will flag it at run time. It is often a trick I use to make sure that I've set the pointers appropriately.

marcosvanella:
> Just a comment -- Marcos just discovered that doing this
> 
> ```
> TYPE(LAGRANGIAN_PARTICLE_CLASS_TYPE), POINTER :: LPC=>NULL()
> ```
> 
> caused an OpenMP data race. I did not realize this, but apparently when a pointer is initialized to null, as here, it automatically assumes a `SAVE` attribute. The subroutine containing this line was called by two OpenMP threads such that `LPC` pointed to the same location in memory. I was told by the thread sanitizer that this variable was global, which made no sense to me until we discovered this `SAVE` feature, meaning that this pointer is now saved outside the routine.
> 
> Long story short, I'm not sure we need to nullify pointers upon declaration. I can't remember when and why we started this. Does anyone remember?

This is something we found out with Chandan while debugging chemistry. We don't want to have initialization and declaration of pointers in subroutine that get called inside an OpenMP loop, as the will be shared between threads leading to data races.

mcgratta:
I removed all the pointer initializations to `NULL()` throughout the code.

PR #14461, opened 2025-04-02
improvement to catchpole plotting script

johodges:
Fonts should be working in all the plots now. The y-axis label in the semilogy plot has decimals for all tick levels which is something I think we want to fix still.


ericvmueller (approved):
Looks good - thanks!

PR #14499, opened 2025-04-11
FDS Verification: add ext_heartbeat_std_curve.fds to FDS_Cases.sh

rmcdermo:


shostikk:
Did not try it yet, but "batch" would not work for me. Could the script inherit partition from Run_FDS_cases.sh or from SBATCH_PARTITION environment variable?

PR #14514, opened 2025-04-15
FDS Build: Change HYPRE version to 2.33.0

cxp484:
Glen, when you merge these changes, please make sure that HYPRE is freshly cloned or updated to the latest version on Spark (Linux), Mac, and Windows. Otherwise, the compilation is likely to fail. I have tested them locally in Spark, my mac and a Windows machine.


cxp484:
Glenn, after speaking with Randy, I went ahead and merged it to avoid leaving the PR hanging.

PR #14555, opened 2025-04-23
FDS Build: Simplify MKL library paths

cxp484:
Hi Everyone, We simplified the makefile to deal with the mkl and impi paths. Let me know if you have any issued while compiling with mkl and impi.


cxp484:
Will push again after fixing the github action errors.

PR #14569, opened 2025-04-28
Bring FireX up to master, refactor vtkhdf output for computational speed, windows GPU build

johodges:
@cxp484 here is the PR per the email chain. I have verified compilation and running on my windows and linux machines.


rmcdermo:
@johodges This is not a very clean PR.  Can we talk to discuss this?  There should be different merges for merging master into FireX and then you fixing FireX issues.  This looks like you are combing a ton of this.  And it is error prone.

PR #14734, opened 2025-06-04
qfds.sh: Add option to use all available cores

mcgratta:


mcgratta:
This PR adds a flag to `qfds.sh` called `-c` that removes the node limit for a job. That is, if you do this
```
qfds.sh -p 800 -c -q all my_job.fds
```
then the job will run over the entire cluster looking for any available core. All that the `-c` does is remove 
```
#SBATCH --nodes=...
```
allowing Slurm to use any core within the given queue.

BTW, `c` stands for cram, as in cram the job onto the cluster.

PR #14747, opened 2025-06-07
FDS Source: bug fix for Gaussian impinging jet model

johodges:
The max HTC at the impinging point is stored in the `SF%H_FIXED` in read. The current formulation calculates `NUSSELT_IMPINGE` based on `HEAT_TRANSFER_COEFFICIENT` which is not defined until line 3631 / 3645. This fix changes the calculation to use the user specified values in the calculation of `NUSSELT_IMPINGE` before the max function is applied.


rmcdermo:
Thanks!

PR #14766, opened 2025-06-14
fix: unexpected SUNDIALS compilation termination in linux caused by m…

ramonmi:
…issing -p for mkdir

### DIFF
--- a/Build/Scripts/SUNDIALS/build_sundials.sh
+++ b/Build/Scripts/SUNDIALS/build_sundials.sh
@@ -53,7 +53,7 @@ if [ -d "$FIREMODELS/sundials" ]; then
   cd $FIREMODELS/sundials/BUILDDIR
   echo "Creating library directry..."
   export SUNDIALS_VERSION=$(git describe)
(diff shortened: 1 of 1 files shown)


cxp484:
@ramonmi Could you clarify the exact error you're encountering when omitting the -p option? We’re not seeing any errors on our end when running without it.

ramonmi:
- When I compile the FDS code without existing hypre and sundials libs, some errors occur as below.
    - It seems that `mkdir` cannot directly create a second-level directory (`sundials/$SUNDIALS_VERSION`).

cxp484:
@ramonmi Thanks. After a test, it seems like the "mkdir ..." line is unnecessary. I am removing it.

ramonmi:
> @ramonmi Thanks. After a test, it seems like the "mkdir ..." line is unnecessary. I am removing it.

Yeah, that's right. Following your prompt, I rechecked and retested and found that this command did not cause the previous failure. Now I have successfully compiled, thanks!

PR #14792, opened 2025-06-27
FDS Source: add HYPRE_PARCSRMATRIXPRINT to HYPRE interface

rmcdermo:


rmcdermo:
@marcosvanella @victorapm This is my attempt to add the API to write out the matrix.  The call does not complain, but it is not writing anything.  Do I need to open the file on the FDS side or will the HYPRE routine try to do that?

Here is an input file that should work but hangs at HYPRE_PARCSRPCGSETUP.

[zone_break_fast_uglmat.fds.txt](https://github.com/user-attachments/files/20952246/zone_break_fast_uglmat.fds.txt)

The hypre matrix write is invoked using
```
&PRES SOLVER='UGLMAT HYPRE', PRESSURE_RELAX_TIME=0.5, WRITE_PARCSRPCG_MATRIX=T /
```

PR #14811, opened 2025-07-01
gh-actions: add intel 2025.2.0 release packages.

marcfehling:
There used to be incompatibilities between the `ifx` and `mkl` packages as observed first in #13994. It appears that they have been fixed.

Let's see if the testers agree!

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -2,9 +2,9 @@ name: cmake
 env:
   # update urls for oneapi packages according to
   # https://github.com/oneapi-src/oneapi-ci/blob/master/.github/workflows/build_all.yml
-  WINDOWS_BASEKIT_URL: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/ae29263e-38b9-4d43-86c3-376d6e0668e7/intel-oneapi-base-toolkit-2025.0.1.47_offline.exe
+  WINDOWS_BASEKIT_URL: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/09a8acaf-265f-4460-866c-a3375ed5b4ff/intel-oneapi-base-toolkit-2025.2.0.591_offline.exe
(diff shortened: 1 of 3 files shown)


marcfehling:
Looks like there is some linker issue with gcc and onemkl 2025.2.0.
```
/usr/bin/ld: warning: libmpi.so.40, needed by /usr/lib/gcc/x86_64-linux-gnu/13/../../../x86_64-linux-gnu/libmpi_usempif08.so, may conflict with libmpi.so.12
/usr/bin/ld: /opt/intel/oneapi/mkl/2025.2/lib/libmkl_blacs_openmpi_lp64.a(mklmpi-impl.o): undefined reference to symbol 'ompi_mpi_dblprec'
/usr/bin/ld: /lib/x86_64-linux-gnu/libmpi.so.40: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

I can force the use of an older version to keep it working for now. Maybe you can try an tackle the issue separately (I guess you might need to set a linker flag somewhere). -- @cxp484

marcfehling:
The linker issue for `linux-gnu-openmpi` vanishes when I specify the previous release of mkl, which is 2025.1.0. I provided a downgrade with the second commit. Testers should be happy now!

Maybe this gives you a clue on what to look for. Does Intel provide a changelog?

marcfehling:
Also, the release build for `linux-intel-intelmpi` works again with the new oneapi release with the third commit!

cxp484:
> The linker issue for `linux-gnu-openmpi` vanishes when I specify the previous release of mkl, which is 2025.1.0. I provided a downgrade with the second commit. Testers should be happy now!
> 
> Maybe this gives you a clue on what to look for. Does Intel provide a changelog?

Thanks, Marc. Here for local builds, we're still using MKL 2025.1, which is probably why we haven’t encountered this error. For now, I’m accepting the changes you made—they look good to me. We probably need to fix this once we update our local mkl to 2025.2.

PR #14812, opened 2025-07-01
gh-actions: Move python test to separate job.

marcfehling:
While working on #14811, I noticed the small python sandbox that feels a bit out of place in the compilation job, so I moved it to its own job.

It was introduced in #12703 and #12734. Out of curiosity, may I ask what your progress in the python integration is? Are you generating the plots in the FDS guides with python now? :-)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -55,19 +55,6 @@ jobs:
         ./make_fds.sh
         ./fds_impi_intel_linux
 
-    # Setup python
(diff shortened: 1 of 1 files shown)


marcfehling:
The faling `linux-gnu-openmpi` job is fixed in #14811.

rmcdermo:
@marcfehling Actually, I am working the python scripts over the last couple of days and making some progress.  Almost there with dataplot.  Next will be scatplot.  Then the remaining specialty scripts.

PR #14823, opened 2025-07-03
Surf layer delamination (fall-off), based on density or temperature

shostikk:
A layer and all layers before it can be made to disappear accoring to FALLOFF_TMP or FALLOFF_DENSITY (layer-by-layer parameter), compared at the last cell of a layer. Can be used for delamination of cross-laminated timber.


shostikk:
Ping @drjfloyd

mcgratta:
There is currently a parameter called `MINIMUM_LAYER_THICKNESS`. Would it be better to call this parameter `MINIMUM_LAYER_DENSITY`  ?

shostikk:
Can't remember what it is used for. Not related to this falloff feature, however. ?

mcgratta:
My comment is simply about the name. Should we call the parameter `MINIMUM_LAYER_DENSITY` to be consistent with `MINIMUM_LAYER_THICKNESS`?

shostikk:
Aa - maybe not. This is density or temperature threshold at layer interface, simulating a failure of adhesive bond. Temperature criteria could be seen from glue TGA or CLT DTMA test. Density criterion could relate to charring-induced weakening of the interface.

FALLOFF maybe doesn't tell enough....?

Layer_density sounds like somehow the whole layer is considered.This is LOCAL criterion

mcgratta:
`FALLOFF` is a bit vague. Should we just use the word `DELAMINATION` to avoid confusion?

shostikk:
I think DELAMINATION_TMP is good. Or DELAMINATION_DENSITY.

We stopped using 'delamination' when we saw that delamination can occur also without falloff (just gap formation) but I think it is commonly used term.

PR #14827, opened 2025-07-04
FDS Source: change FALLOFF name to DELAMINATION

shostikk:


shostikk:
As per discussion in https://github.com/firemodels/fds/pull/14823#issuecomment-3033247353

PR #14961, opened 2025-08-14
FDS Source: Rework velocity BC at corners

mcgratta:
This PR essentially applies a no-slip velocity boundary condition to corners. This change improves simulations of flow resistance through orifices.


mcgratta:
Issue #14594

PR #14968, opened 2025-08-15
FDS Source: Fix error for PROF and PART with INIT and multiple MESH

drjfloyd:
Had a case where I put one particle each in two MESHes with INIT and had a bunch of PROF outputs for each particle. The case failed with error 430 when it got to processing the first PROF for the second MESH. The cause was it was trying to process the surface type for the profile for a particle on the second mesh while still on the first mesh. I made a single line edit to the logic for the error message. Now it only checks for the type of surface if it is not a particle or if it is a particle and in the mesh where the particle is. Can you take a look at the edit? I feel like I am probably missing something.

### DIFF
--- a/Source/init.f90
+++ b/Source/init.f90
@@ -2699,7 +2699,7 @@ PROF_LOOP: DO N=1,N_PROF
 
    ! Check for potential errors
 
-   IF (SF%THERMAL_BC_INDEX/=THERMALLY_THICK) THEN
+   IF (SF%THERMAL_BC_INDEX/=THERMALLY_THICK .AND. (PF%IOR/=0 .OR. (PF%IOR==0 .AND. PF%MESH==NM))) THEN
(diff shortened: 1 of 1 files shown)


mcgratta:
I don't understand why the logic failed. If you are on mesh 1 (NM=1) and you are processing a profile that is applied to the particle on mesh 2, wouldn't the surface type be given by
```
SF => SURFACE(LAGRANGIAN_PARTICLE_CLASS(PF%PART_CLASS_INDEX)%SURF_INDEX)
```
which does not seem to be mesh dependent. Can you tell me what the various values were for NM, etc.

drjfloyd:
VOLUME_INIT_PARTICLES is called before INITIALIZE_PROFILE. In V_I_P we loop over all PF and set the part index. So for mesh 1 we enter INIT_PROF and the part index for profiles in mesh 2 results in the INERT surface type being assigned.  

I think I see a cleaner approach. We don't have to wait until we insert particles to set PF%PART_CLASS_INDEX. In read once we have a PROF with an INIT we can check the INIT for a PART_ID and set in read.

PR #14999, opened 2025-08-21
FDS Source: Set MAXIMUM_SCALING_HEAT_FLUX

drjfloyd:
@johodges 

Have a case with Spyro that is giving me problems and it is making me rethink not having a max scaling flux. This PR sets it to 1500 kW/m2 which is ~2200 K to a 300 K surface with h=100 W/m/K.   This should be higher than we would ever expect to see in actual simulations but not so high that we get high speed surface mass fluxes in pyrolysis.


johodges (approved):
I have generally been setting the upper limit to around 500 kW/m2 in my models in the rail car. 1500 kW/m2 as a default is reasonable to me. Changes look good.

johodges:
FYI, I do not think I have permissions to merge the PR.

PR #15064, opened 2025-09-04
FDS Source: BI-DIRECTIONAL PROBE edits and Type-K props

drjfloyd:
I noticed that the properties for a TC are pure nickel. The typical TC used in expirements is Type K which is a nickel alloy whose density is a bit less than pure nickel. Also the specific heat for Type K increases by 1/3 from ambient to flame temperature. This PR also sets the Type K density and makes the TC specific heat temp dependent using Type K properties. It also adds a SPECIFIC_HEAT_RAMP input for PY so someone could set a ramp for another type of TC.

I also changed the explicit TC update to be crank nicholson. For typical TC diameters used in fire tests and typical FDS time steps the explicit is OK but if you get to very fine TCs  or are doing something with large timesteps the implicit can be smoother.


drjfloyd:
Note these changes won't impact the typical steady-state values we use in V&V, just the transient.

PR #15134, opened 2025-09-17
FDS Validation: Fix file structure in JH_FRA cases with CATF

johodges:
I tested running these on our cluster last night and building the guides. Everything seemed to work fine. I forgot to update my tags first though so the plots show FDS-6.9.1. I was going to hold off on committing the updated out files and re-run with the right tags. Do you want to run them on your end or should I?


drjfloyd:
I have changes to push where I have run cases including these with BI-DIRECTIONAL PROBE where they were used for velocity. Was going to wait for a clean firebot since there were some issues with plots recently before pushing up more changes to plotting. So I don't think a need here to push up the out files only to have them quickly replaced.

mcgratta:
Also, I run the validation cases under firebot's account. firebot is one of our users. I like to have all the cases run there, where I can save all the output and refer back if something arises later. So no need to run these cases. I will do it. 

Jason -- you can push your changes any time.

PR #15164, opened 2025-09-22
FDS Source: Do not allow + sign in labels

mcgratta:


drjfloyd:
These restrictions should be added to the Guides.  

I can see why a "," might cause an issue when we go to read a csv file but what is breaking with a "+" sign?

mcgratta:
The comma restriction is in the guides, now. I'll add the +. It is because our Matlab/Python scripts take the + to mean that two columns are to added together.

drjfloyd:
Is there another symbol or something we could use in the script to add columns?  ++? A single plus is a handy item labels at times.

mcgratta:
I did not find any + signs in the FDS V&V cases, and only a few in the experimental data sets. I prefer to wait until we've converted from matlab to python because we don't want to introduce any other sources are error.

drjfloyd:
I have used + in many inputs outside of V&V runs. Things like "X +1 m" where X is some location in the domain or "X+Y" where I have a CTRL that sums X and Y and a DEVC that outputs the result. For making plots where you want to use the DEVC ID as the plot label, not having "+" could mean needing more complex or less intutive labels.

mcgratta:
What has probably happened is that we used to put double quotes around the `DEVC` column heads. That forced us in some Matlab scripts to parse the headers and eliminate the quotes. Not going over the Python, we're finding conflicts with the + syntax. We would rather not put any quotes into the .csv files, so we might have to make some sacrifices on column names to avoid complicated parsings.

PR #15186, opened 2025-09-23
Python: add fm_data_center.py

rmcdermo:
This is not fully tested, but runs.  @drjfloyd you might want to double check this to see if it is doing what you expect.  I am not seeing any data on the scatter plot when I run it, but I suspect there is a reason why it is commented in the main script.


drjfloyd:
I think this is because at one point we had a scater plot in the validation guide built specially for these tests to replication ones that were in the FRPF report.  We now do line plots with the normal csv file and extract comparison data from that.

PR #15267, opened 2025-10-02
Python: improve cross-platform support for FDS_validation_script.py

johodges:
I tested this with:
* linux - bash shell
* windows - git bash shell
* windows - cmd prompt


drjfloyd:
This is working on my PC.

rmcdermo:
If runpy works, I'd rather try that so we don't have to do anything special with PYTHONPATH.  Setting up the environment should be all you need to do.

PR #15275, opened 2025-10-04
Python: add version_fontsize as keyword argument for plot_to_fig

rmcdermo:


rmcdermo:
@mcgratta you should now be able to add, for example, plot_to_fig(fig, ..., version_fontsize=8)

PR #15308, opened 2025-10-10
FDS Source: rework TAU_U to be consistent with Tech Guide

rmcdermo:
In theory, this should be the same implementation as before... fingers crossed.  I spot checked cases in fds/Veification/Fires/


drjfloyd:
Looks like they are equivalent to me.

Issue #15328, opened 2025-10-14
Convert a few matlab scripts


mcgratta:
disp('saad_mms_temporal_error...');       saad_mms_temporal_error
disp('saad_cc_mms_temporal_error...');    saad_cc_mms_temporal_error
disp('shunn_mms_temporal_error...');      shunn_mms_temporal_error
disp('shunn_mms_favreZ...');              shunn_mms_favreZ

See if you can combine the two saads into one script, and the two shunns into another. Avoid having a random function file outside of the script if it only serves one case.

rmcdermo:
These were originally mine.  I can work on them if you want to assign Jason something else.

mcgratta:
Most of them were yours or your students. I'd prefer you continue working the big scripts like scatplot and dataplot. These smaller scripts are good for learning.

drjfloyd:
The four should be converted.  I'll close once firebot chews on the new scripts.

mcgratta:
I just noticed 
```
disp('shunn_mms_error...');               shunn_mms_error
disp('shunn_cc_mms_error...');            shunn_cc_mms_error
```
Add these to the current shunn script

drjfloyd:
In the queue:

disp('ashrae_7...');                       ashrae_7
disp('cat_propane_depo...');               cat_propane_depo
disp('water_evap_1_const_gamma...');       water_evap_1_const_gamma
disp('shunn_mms_error...');               shunn_mms_error
disp('shunn_cc_mms_error...');            shunn_cc_mms_error
disp('hvac_mass_transport...');           hvac_mass_transport

drjfloyd:
Moving on to:
disp('flame_species...');                  flame_species
disp('burke_schumann...');                 burke_schumann

drjfloyd:
disp('rms_cov_corr...');                  rms_cov_corr
disp('hot_layer_collapse...');            hot_layer_collapse
disp('radiating_polygon...');             radiating_polygon

PR #15333, opened 2025-10-14
FDS Verification: Add shunn_mms.py (Issue #15328)

drjfloyd:
replaces shunn_mms_temporal_error and shunn_mms_favreZ with shunn_mms


mcgratta:
If the script works, delete the matlab script and remove its entry from FDS_verification_script.m

PR #15362, opened 2025-10-15
FDS Verification: convert impinging_jet to python.

johodges:


mcgratta:
There is no need to "set_ticks_like_matlab". Go through this script and eliminate the various ax settings that are not absolutely needed. Let `fdsplotlib` do the work. The defaults in `fdsplotlib` and `matplotlib` are usually OK.

PR #15373, opened 2025-10-16
FDS Utilities: Add Spyder IDE for Linux, MacOS, and Windows

cxp484:
I’ve added two setup scripts for the **Spyder IDE** — one for Linux/macOS and another for Windows.  
These scripts create a Python environment using `requirements.txt` and automatically install Spyder.

The setup steps are simple, and I’ve documented them here:  
👉 [Spyder IDE Setup for FDS Python Utilities](https://github.com/firemodels/fds/wiki/Spyder-IDE-Setup-for-FDS-Python-Utilities)

Please give it a try and let me know if you encounter any issues.  
Once it’s working for everyone, you’ll be able to use **Spyder** – much like **MATLAB** – to edit, run, and visualize Python scripts.


mcgratta:
Does Spyder try to launch a GUI from the linux command prompt, or does it just open up a special command-like session?

cxp484:
It will open a Window like this from the linux command prompt:

image

mcgratta:
Yes, assuming you are running Xwindows or you are logged into the console.

johodges:
This should work fine, but it does mean that spyder is installed within the virtual environment. I tend to have a single installation of spyder installed in the base python environment and then connect that spyder to different python interpreters. That way I only have one executable of spyder rather than a separate one for each python environment. It is not a big deal, but it is nice to have a single spyder application on the task bar/shortcut.

image

cxp484:
@johodges Yes, you’re right. If someone already has a separate Spyder installation, they can simply change the Python interpreter as you showed.  
I just wanted to make things easier for users who don’t have Spyder installed and want a ready-to-use FDS Python environment.

PR #15399, opened 2025-10-17
FDS Validation Guide: Convert validation_git_stats.sh to python

mcgratta:
Randy -- I converted this script from bash using CodeConvert. It ran clean and produced a .tex file in `FDS_Validation_Guide/SCRIPT_FIGURES/Scatterplots`. But the table is empty, I suspect because I cannot run the matlab scripts and compile the validation guide. I'm putting the script here as a placeholder. I did not remove the old sh file.

### DIFF
--- a/Utilities/Scripts/validation_git_stats.py
+++ b/Utilities/Scripts/validation_git_stats.py
@@ -1,0 +1,275 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import subprocess
+import re
+import tempfile
+from pathlib import Path
+import argparse
+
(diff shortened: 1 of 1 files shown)


ericvmueller:
pandas has a handy [to_latex](https://pandas.pydata.org/docs/reference/api/pandas.io.formats.style.Styler.to_latex.html#pandas.io.formats.style.Styler.to_latex) method that writes out a table as a latex string. Might help simplify a lot of what is going on here if the scatterplot data is in a pandas dataframe

mcgratta:
Maybe, but the real problem here is that I cannot compile the FDS Validation Guide because I cannot run matlab to make the figures.

PR #15422, opened 2025-10-19
Bphusted patch 1, Error in user guide on STATISTICS_END, default should be T_END instead of T_BEGIN

bphusted:
There is an error in the FDS User Guide.
It stated the the default value to STATISTICS_END is T_BEGIN, it should be T_END.


drjfloyd:
thanks Bjarne

PR #15433, opened 2025-10-20
Python: add Group_Style and Group_Key_Labels to scatter plots

rmcdermo:


rmcdermo:
Left to do:

* check p-test values against Matlab
* check latex tables against Matlab

Issue #15209, opened 2025-09-24
LNG_Dispersion cases failing with numerical instability


mcgratta:
Most of the LNG_Dispersion cases failed with a numerical instability. The one that fails the fastest, in about 5 min, is `/Validation/LNG_Dispersion/FDS_Input_Files/MaplinSands27.fds`. The instability appears at the lateral exterior boundary. The failures began with this commit:
```
(fds_python_env) mcgratta@spark-login fds_central ((no branch, bisect started on master)) $ git bisect good
2bf4ee6753e872aa376fd75eccd161f4744832fe is the first bad commit
commit 2bf4ee6753e872aa376fd75eccd161f4744832fe
Author: rmcdermo <randy.mcdermott@gmail.com>
Date:   Wed Sep 3 17:13:34 2025 -0400

    FDS Source: remove FLUX_LIMITER_SINGLE_COEF parameter

 Source/cons.f90 |   1 -
 Source/divg.f90 | 158 ++++++++++--------------------------------------------
 Source/init.f90 |  29 +++++-----
 Source/main.f90 |  12 +----
 Source/mass.f90 | 187 +++++-----------------------------------------------------------
 Source/read.f90 |   2 +-
 Source/wall.f90 | 317 +++++++++++++++++++++++++-----------------------------------------------------------------------------------
 7 files changed, 133 insertions(+), 573 deletions(-)
```

rmcdermo:
@mcgratta  I plan to dig into this today.  My understanding is that PERIODIC is working, but PERIODIC FLOW ONLY is not.  Let me know if you have any other hints from your work on this.  Thanks

rmcdermo:
FYI, I tried the case with UGLMAT HYPRE and still got an instability at roughly the same time.

mcgratta:
I think the problem is that a small increase in temperature flowing out of the left boundary flows into the right and vis verse. Like a bad feedback loop. I would printout values in the two ghost cells at the point of failure.

rmcdermo:
Is there anything physically wrong with just using a MIRROR bc for the lateral walls?  I'm not sure we can make the periodic velocity consistent with the divergence field.

rmcdermo:
I think we are unintentionally getting caught in this logic, where the external wind is not consistent with the periodic flow.
```
      ! Avoid large fluxes at open downwind boundaries

[...]

PR #15624, opened 2025-11-25
FDS Source: do not reset face values for OPEN_WIND_BOUNDARY if bc is …

rmcdermo:
…PERIODIC FLOW ONLY


rmcdermo:
This addresses Issue #15209

PR #15666, opened 2025-12-04
Python: add usetex to plot_to_fig call

rmcdermo:


mcgratta:
Do we add `use_tex=True` or `usetex=True` to the list of arguments on the `fds_plot_to_fig` line?

rmcdermo:
`usetex=True`.  See compression_wave.py

In general, try to avoid using this.  For now, I am just testing.  I think I can get the super and sub scripts to work without it.  So, you should eventually only need this for more complex latex labels.

PR #15678, opened 2025-12-05
Python: update validation dataplot inputs for exponents

rmcdermo:


rmcdermo:
@mcgratta @drjfloyd @marcosvanella @cxp484 @ericvmueller @johodges @gforney @shostikk

All, I wanted to give everyone a heads up on the latest thinking for fonts when using fdsplotlib.  There are essentially three levels of fidelity we can employ to get the nicest looking plots.  Our goals are to keep plot file sizes to a minimum (ideally about 3 KB for each pdf) but to allow the flexibility where necessary to use full LaTeX math fonts for labels, legends, titles, etc.  Let's consider the three levels, but I will also include a 0-th level, trivial case (no math required).

0. In the case that your plots have no math symbols or exponents, there is really nothing to worry about.  Just make sure that your plots end up at something at or less than 10 KB.  If they don't, something is probably wrong.

1. Wherever you need a superscript 2 or 3, use unicode.  For example, kg/m[unicode 3].  Just literally copy and paste one of the unicode 3 examples from elsewhere in the FDS_validation_dataplot_inputs.csv file.  The reason we do this instead of `$^3$` is that the latter results in a file size about 4 times larger, so 12 KB instead of 3 KB.  Multiply that by 8000 plots and it is significant.

2. For other math text, including simple italic variables, such as `$u$` or `$x$`, the unicode does not exist, so we have to fall back on math text with stix font.  This seems to work well, with the only downside being the increase in file size to O(10s) KB.  You can also do, for example, `$\overline{u'v'}$` and `$X_{\rm H_2}$`.  Python will respect the math text, but you end up with stix font instead of Times (or Nimbus).

3. In cases where you truly need latex, then you write the math text as you would for a latex document and pass `usetex=True` to `plot_to_fig`.   We will soon be making this an option in the config file, replacing the last column that used to specify the Matlab Font_Interpreter as TeX or LaTeX.  It will soon be used to specify the parameter `usetex`.  When you use this on a plot, the file size balloons to 100s of KB.  So, please use sparingly.  Check to see how things look without it first, usually things are ok, in my experience, with the latest settings.

drjfloyd:
in the various .py files do we just also put in unicode. For example in the plate_view_factor.py we just replace the `$^2$` with the unicode alt-0178 typing on keypad?
                             y_label='Heat Flux (kW/m$^2$)')

rmcdermo:
Correct.  I need to grep on this and replace the specialty scripts.  I'll work on that.

PR #15704, opened 2025-12-11
FDS Validation: Zero out bad Catchpole points

mcgratta:


mcgratta:
Randy -- make fdsplotlib ignore data mismatch if the Scatterplot Quantity is 0

PR #15782, opened 2026-01-01
FDS Validation: Update FM Burner cases with latest experimental data.

cxp484:


cxp484:
@rmcdermo: I’ve created two pull requests: one for macfp-db and another for firemodels/fds to update the FM Burner validation case. The firemodels/fds PR hasn’t been merged yet. Would you like me to also update firemodels/out with the latest CSVs? Otherwise, the Validation Manual will fail.

rmcdermo:
@cxp484 we cannot change the guides until the data is in the out repo.  And we cannot run the cases and put them into the out repo until they have a legit commit tag on master.  So, please rework this PR to only contain changes to the FDS input files.  Thanks

PR #15832, opened 2026-01-12
Python: use 0.0 default for slope in Rate of Spread quantity; these p…

rmcdermo:
…oints still excluded from scatplot due to window in scatterplot config file

### DIFF
--- a/Utilities/Python/fdsplotlib.py
+++ b/Utilities/Python/fdsplotlib.py
@@ -106,28 +106,6 @@ def _compute_metrics_block(
 
     N, ncols = Y.shape
 
-    # --- comparison mask (like MATLAB) ---
-    comp_mask = np.isfinite(x)
-    if np.isfinite(comp_start):
-        comp_mask &= (x >= comp_start)
-    if np.isfinite(comp_end):
-        comp_mask &= (x <= comp_end)
-
-    if np.isfinite(dep_comp_start) or np.isfinite(dep_comp_end):
-        y0 = Y[:, 0]
-        dep_mask = np.isfinite(y0)
-        if np.isfinite(dep_comp_start):
-            dep_mask &= (y0 >= dep_comp_start)
-        if np.isfinite(dep_comp_end):
-            dep_mask &= (y0 <= dep_comp_end)
-        comp_mask &= dep_mask
-
-    if not np.any(comp_mask):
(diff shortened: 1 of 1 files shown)


rmcdermo:
@ericvmueller This now returns a 1e-12 for the slope for "Rate of Spread" for the Catchpole cases.  The plot min is still at 0.001 for the scatterplot, so the data are outside that window and do not contribute to the bias.

@mcgratta If you run FDS_validation_script on USFS/Catchpole you will have the raw csv data you can use to analyze the bias.  I'm hoping we can get something like what is below.  The negative point comes from the latest CSIRO grassland data.

To you both: the negative point below comes from the latest CSIRO grassland results.  I can hack in a mask to ignore that, but why do we get a negative ROS?

Screenshot 2026-01-12 at 2 44 57 PM

ericvmueller:
The negative CSIRO spread rate is a case of a fire going out and using nonsense fire position to compute slope. This could be fixed with kill logic like in Catchpole so we stop writing position when the fire goes out.

rmcdermo:
Or I could clip the ROS at 0 in scatplot.  Let me know

ericvmueller:
Sure that would be a good safeguard if you want to add that. But let me also try running a few of these with a kill condition. I think even for 'good' cases we can end up with an artificially low slope by including data after the fire goes out

image

ericvmueller:
Here are two alternatives. The first ends the run when the fire goes out, which gives a better estimation of the slope for the finer grid cases, but it gives a high spread rate for the case that goes out.
image

The second option is to use Jason's logic to clip the position to zero when the fire goes out, but also end the run if the fire reaches the far boundary. This gives good slope values for the finer cases and a negative slope for the case that goes out, which could be clipped to zero in scatplot as you say. 
image

I'd say we go with the second approach and do the same for the Catchpole cases.

rmcdermo:
I'm fine with the second also.  In my mind, there should be data throughout the specified window, else the slope is null and that should be treated as zero.  Or we do something else with null values---I just don't like deciding that the value is null *a priori*, which is what setting 0 in dataplot does.

ericvmueller:
Okay, I'm going to try this similar kind of change for the catchpole cases and run through them again on blackbird. If that looks okay I think we can settle on this method.

PR #15842, opened 2026-01-13
Python: minor edit to Dataname column for UMD Line Burner

rmcdermo:


rmcdermo:
Interesting that these do not seem to be plotted anymore.  I do not recall removing them, but they have "s" as their switch_id.

PR #16021, opened 2026-03-06
FDS User Guide updates. Change from ^circ to \textdegree

drjfloyd:


mcgratta:
Just curious, is there any need for the brackets in `\degC{}`. I see it both ways. In any event, this is better than `\unit{\degreeCelsius}.

drjfloyd:
Goodtip. I'll make that change. 

The {} was needed when there was a space afterwards otherwise there would be no space

10\degC was used 

would wind up

10oCwas used

if you had a . or ) after then the {} would be skipped so you didn't get an unintended space.

mcgratta:
Wouldn't we write it like this 

10~\degC

drjfloyd:
Yes mistyped.  Either way, with that macro command, without a {} after it when using it you do not get a space.

PR #16028, opened 2026-03-08
FDS Source: Add option to dump model info in JSON format

JakeOShannessy:
This PR is based on something we've used for a number of years but didn't contribute the code upstream because it added an optional dependency which FDS wasn't set up for. Now that FDS handles sundials and hypre smoothly this is now
feasible.

This may or may not be of use to others so I've made this PR to see if there is any interest.

We quite often want to do a number of related things (each of which is covered in more detail below):

1. Query key values from an input file.
2. Inspect input files as a form of static analysis.
3. Aggregate values from input files.
4. Verify the actual values that are used in the model data structures.

To help with these tasks, this PR adds a commandline option (`--json`). When present, FDS will only run the initial check stage where the .fds file is parsed and loaded. If no errors are thrown, FDS will dump a subset of the model details for inspection in JSON format (either directly to a file or to stdout).

Once in this format, the dumped model data can be inspected manually or (more likely) passed to other tools or ad-hoc scripts to run the relevant queries.

# Query Key Values

Probably the least critical use case but very useful in a number of workflows is to fetch simple values from an input file. 

There are also other practical application, even something is simple as fetching the CHID of a .fds file. `test_model_small.fds` from the fig repository is used as an example and the `jq` commandline tool is used to filter the JSON.

```sh
# What is the CHID of a given simulation
fds test_model_small.fds --json - | jq .chid
> "test_model_small"
# How long is a simulation?
fds test_model_small.fds --json - | jq .time
> {
  "begin": 0.0,
  "end": 600.0
}
```

This is particularly useful when managing simulations.

# Inspect Inputs for Static Analysis

It's often useful to be able to programmatically analyse a given input file for certain properties.

A trivial example might be to inspect the a given model to inspect the chemistry
of a REAC namelist. 

```sh
fds test_model_small.fds --json - | jq .reacs
```

This gives the following:

```json
[
  {
    "index": 1,
    "id": "null",
    "equation": "null",
    "c": 6.2999999999999998,
    "h": 7.0999999999999996,
    "n": 1,
    "o": 2.1000000000000001,
    "epumo2": 13377858.82877272,
    "heat_of_combustion": 2.27E+7,
    "soot_yield": 0.01,
    "co_yield": 0.0
  }
]
```

Here we can inspect that chemical formula is correct regardless of whether the `FORMULA` parameter is used or `C`, `H`, `N`, `O`, parameters are used. Also, as we are taking the values directly from the data structures in FDS, any logic around default values has already been handled by FDS itself.

# Aggregate Value from Input Files

A common scenario is to have a series of simulations with slightly differing inputs to test a certain input parameter. For example, you have a series of simulations with increasing mesh resolution which you want to plot against the value of a given temperature probe. Instead of carefully generating the input files and using an appropriate naming scheme, the mesh resolution of each value can simply be queried:

```sh
fds test_model_small.fds --json - | jq .meshes[].cell_sizes
> {
  "dx": 0.099999999999999978,
  "dy": 0.099999999999999978,
  "dz": 0.099999999999999978
}
{
  "dx": 0.099999999999999867,
  "dy": 0.099999999999999978,
  "dz": 0.099999999999999978
}
```

Critically, this also accounts for any `&MULT` namelists or other nuances that can complicate just reading simple values from an input file.

# Verify Values in Model Data Structures.

When discussing the static analysis of input files above, the chemical properties of a reaction were used as an example. While it's useful to be able to check these programmatically, the values in the input file are unambiguous.

A far more useful example is where the relationship between the values of an input file and the simulation data is more nuanced. Here are two useful examples.

## Surface Properties

By iterating over each OBST and VENT in the JSON output we can examine every surface and material used. As this pulls information directly from the variables set within FDS it also accounts for any default values set.

```json
"obsts": [
  {
    "index": 1,
    "id": "OBST-1",
    "surfaces": {
      "x_min": "UPHOLSTERY",
      "x_max": "UPHOLSTERY",
      "y_min": "UPHOLSTERY",
      "y_max": "UPHOLSTERY",
      "z_min": "UPHOLSTERY",
      "z_max": "UPHOLSTERY"
    },
    ...
  },
  ...
],

```
## Device Locations

The input values of an FDS input file can be any real number, but within the simulation objects such as devices and obstruction must align to the mesh. This means the location in the simulation will often differ slightly from the input file. Using the JSON output, the precise cell locations use by FDS can be extracted.

```sh
fds test_model_small.fds --json - | jq '.devices[] | {id,points}'
> {
  "id": "SmokeDetector",
  "points": [
    {
      "i": 6,
      "j": 10,
      "k": 10,
      "init_solid": false
    }
  ]
}
```


mcgratta:
Thanks Jake. Let me discuss this with the rest of the development team this week.

JakeOShannessy:
Thanks Kevin.

I've just noted that rockylinux 9 etc. don't have `json-fortran` available by default so I've added a commit so set cmake to build `json-fortran` from source.

A short script for a quick test of this using cmake:
```sh
cmake -B Build/cmakeb -DUSE_HYPRE=OFF -DUSE_SUNDIALS=OFF -DDUMP_JSON=ON
cmake --build Build/cmakeb
ctest --test-dir Build/cmakeb
```

mcgratta:
Jake, we've talked this over and decided that we would rather not incorporate this code directly into FDS. Two reasons. First, it is something that we would have to support and we are short-handed. We worry that every time we change the input file parameters we would have to modify this routine. Second, it seems to us that this routine would be more effective as a stand-alone Python script that one could use to parse the input file. There already exist various types of pre-processors that users have created to do things similar to what yours does, and, of course, there are packages like PyroSim that do error checking of inputs. We'd prefer that these applications not be written directly into FDS because there is no need for that. We want to keep FDS relatively light and build utilities outside of it.

JakeOShannessy:
Perfectly reasonable, I figured that was a possibility and will just keep it separate.

PR #16029, opened 2026-03-09
FDS Source: Remove PUBLIC attribute for subroutine

mcgratta:
FYI, I am debugging a case that seg faults in an HVAC routine when compiled with `-O2 -ipo` and not with `-O1`.  I am eliminating a possible source of confusion for the inter-procedural optimization. Probably a long-shot, but there is no need to assign `PUBLIC` to a subroutine called from within the module.


drjfloyd:
Thanks for catching that. I made some recent changes to HVAC initialization and that routine is no longer externally called.

drjfloyd:
When I was fixing that bug where changed init I did get a segfault in a deallocation which led me to simplify/streamline some of the allocations.  You may be seeing something similar.

mcgratta:
I'm going to run the problem case with and without `-ipo` to see if the problem is with `-ipo` or `-O2`.

PR #16061, opened 2026-03-16
Manuals: slightly rework reaction_time_scale script and plot

rmcdermo:
@mcgratta Have a look at this.  I'd like to keep a description of tau_flame around.


mcgratta:
If you're going to make 100 a variable, I would call it $s_{\rm T}$. Also, $\tau_{\rm flame}$ is currently 1e10. I removed it from the discussion in the Tech Guide, but left it in the code. You might want to add it back to the Tech Guide.

rmcdermo:
I simplified the discussion.  I want to be careful about assigning to much physics to this.  If we act like it is premixed combustion, someone will hammer us on it.  But I agree that the idea of a maximum deflagration velocity (a premixed concept) bounds the time scale.  And that all we can do is provide an order of magnitude estimate of this value.  But if you look at the paper I cited, order 100 m/s seems reasonable.

PR #16064, opened 2026-03-16
 FDS Source: Add TURBULENT_FLAME_SPEED

drjfloyd:


mcgratta:
Aren't flame speeds usually denoted by $s$, as in $s_{\rm L}$, the laminar flame speed?

drjfloyd:
good point.  pushed a change

PR #16196, opened 2026-04-30
Resetting the default value for K_FACTOR to 1 when FLOW_RATE is given.

Malibakh-aalto:
Change default K_FACTOR from 10.0_EB to 1.0_EB (same as user guide) when it is not specified and FLOW_RATE > 0.

This avoids underestimating OPERATING_PRESSURE in the relation
"IF (OPERATING_PRESSURE < 0._EB .AND. K_FACTOR > 0._EB)  OPERATING_PRESSURE = (FLOW_RATE/K_FACTOR)**2", where the previous default led to artificially low values.


drjfloyd:
Typical sprinkler K-factors are 5 or higher. 
OP = (FR/10)^2 = 0.01 FR^2
OP = (FR/5)^2 = 0.04 FR^2
OP = (FR/1)^2 = FR^2

The default of 10 gives pressure within a factor of 4 for K-factor = 5 (and the K-factor could be higher). Changing to 1 is a pressure that is a factor of 25 compared to K-factor = 5.

I don't think this is a good change.

Malibakh-aalto:
> Typical sprinkler K-factors are 5 or higher. OP = (FR/10)^2 = 0.01 FR^2 OP = (FR/5)^2 = 0.04 FR^2 OP = (FR/1)^2 = FR^2
> 
> The default of 10 gives pressure within a factor of 4 for K-factor = 5 (and the K-factor could be higher). Changing to 1 is a pressure that is a factor of 25 compared to K-factor = 5.
> 
> I don't think this is a good change.

Dear Jason,

Thanks for the clarification.
In that case, it might be worth revisiting the user guide. In Table 23.25, the default value for `K_FACTOR` is listed as 1 L/(min·bar^1/2), whereas in the source code it is initialized as

'K_FACTOR = -1.0_EB'

and later assigned a default value of 10 when needed.

Thanks, 
Reza

drjfloyd:
Thanks for noticing the guide issue.  Will fix that.

PR #16212, opened 2026-05-01
FDS User Guide: remove mention of wavelet error from tables

rmcdermo:
This also removes white spaces from blank lines.


drjfloyd:
does this need to be removed from the source as well?

rmcdermo:
Yes, working on it.

PR #16289, opened 2026-05-19
FDS Source: Make arguments to GAS_PHASE_OUTPUT optional

mcgratta:


mcgratta:
FYI -- this change moves many of the global variables in `dump.f90` to the relavant subroutines. There are a few left to deal with, like `DRY`, but I want to make sure things are OK. I did run a local `firebot` on this before committing. Also, variables to `GAS_PHASE_OUTPUTS` are now optional, which cleans things up a bit. 

Marcos, you might want to check calls from the `GEOM` routines.

PR #16388, opened 2026-07-02
FDS Validation: fix inverted y-axis XB in memorial tunnel jet fans

johodges:


mcgratta:
Thanks. FDS inverts `OBST` coordinates automatically, but it's a good idea to set things up so that $x_1<x_2$.

PR #16421, opened 2026-07-16
Correct POINTS value in flame height example

Malibakh-aalto:
Since the grid resolution is 0.2 m, the number of sampling points from z=0.1 m to z=9.9 m is 50, not 100. This is also consistent with DZ=0.1, because each sampling interval, zi-dz <= z <= zi+dz, has a total width of 2dz=0.2 m.


drjfloyd:
Thanks for catching this.

